PR #306, opened 2026-04-09
fix: eliminate infinite setState loops, false connection banners, and unnecessary re-renders

lilseyi:
## Summary

Fixes four user-reported issues across 9 files:

### 1. "No internet connection" banner flash on app open (iOS)
iOS kills WebSocket connections when the app goes to background. On foreground, the WebSocket needs 1-3s to reconnect, but the old code treated this as a mid-session disconnect (2s debounce) — showing a false red "No internet" banner followed by a green "Connected" banner for 3s.

**Fix:** Reset ConnectionProvider to `'connecting'` state on foreground, using the 6s cold-start grace period (no banner). WebSocket reconnects silently within this window.

### 2. Refresh flicker when opening the app
Every foreground event unconditionally refreshed the auth token (30-day JWT), even with 29+ days remaining. This wrote a new token to AsyncStorage → 500ms polling detected the change → all 50+ `useAuthenticatedQuery` consumers got new args → every query re-subscribed → visible refresh jank.

**Fix:** Decode JWT `exp` claim, only refresh when < 7 days remain. Most foreground events skip refresh entirely.

### 3. Maximum update depth exceeded (Sentry)
Multiple `router.replace()` calls inside `useEffect` hooks without guards, plus `parsedChannel` creating new object references on every render, caused infinite navigation loops.

**Fix:** Added `hasRedirected` ref guards (with param-change reset) on all navigation effects. Memoized `parsedChannel`. Simplified `hasCommunity` derivation in tabs layout.

### 4. Cascading re-renders / general jank
- 4 context providers creating new value objects on every render
- `useStoredAuthToken` polling interval recreated on every token change
- `useAuthenticatedQuery` spreading `{ ...args, token }` inline (50+ call sites)
- ConnectionProvider's `status` in its own useEffect dependency array

**Fix:** Memoized all provider context values. Stabilized polling interval with refs. Memoized query args with `stableStringify` (sorted-key serialization). Removed self-triggering `status` dependency.

## Files changed (9)
- `apps/mobile/app/(tabs)/_layout.tsx` — simplified `hasCommunity` derivation
- `apps/mobile/app/inbox/[chat_id]/index.tsx` — memoized `parsedChannel`, added redirect guards
- `apps/mobile/features/chat/components/ConvexChatRoomScreen.tsx` — added redirect guard
- `apps/mobile/providers/AuthProvider.tsx` — token refresh gating (`shouldRefreshToken`)
- `apps/mobile/providers/ConnectionProvider.tsx` — foreground reset, grace period fixes, memoized context
- `apps/mobile/providers/EnvironmentProvider.tsx` — memoized context value
- `apps/mobile/providers/NotificationProvider.tsx` — memoized context value
- `apps/mobile/providers/OTAUpdateProvider.tsx` — memoized context value
- `apps/mobile/services/api/convex.ts` — stable token polling, memoized query args

## Test plan
- [ ] Open app from background on iOS — no red/green banner flash
- [ ] Open app from background — no visible query refresh flicker
- [ ] Navigate between inbox channels rapidly — no crash
- [ ] Check Sentry for "Maximum update depth exceeded" errors — should stop
- [ ] Verify auth still works after 23+ days (token refresh kicks in at 7 days remaining)
- [ ] Verify real network loss still shows the red banner correctly

### DIFF
--- a/apps/mobile/app/(tabs)/_layout.tsx
+++ b/apps/mobile/app/(tabs)/_layout.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react';
+import { useMemo, useRef } from 'react';
 import { View } from 'react-native';
 import { Tabs } from 'expo-router';
 import { Ionicons } from '@expo/vector-icons';
@@ -16,19 +16,21 @@ export default function TabsLayout() {
   const isAdmin = user?.is_admin === true;
   const isInternalUser = user?.is_staff === true || user?.is_superuser === true;
 
-  // Use stable state for tab visibility to prevent infinite loops
-  // Only update when community actually changes (not on every render)
-  const [hasCommunity, setHasCommunity] = useState(!!community?.id);
+  // Derive tab visibility from community ID using a ref to avoid
+  // setState during render which can cause infinite loops when Expo Router
+  // rehydrates navigation state (getRehydratedState → setState → re-render).
+  // Using useMemo instead of useState+useEffect avoids the extra render frame
+  // where stale state could cause the navigation config to oscillate.
   const prevCommunityIdRef = useRef(community?.id);
+  const prevHasCommunityRef = useRef(!!community?.id);
 
-  // Update hasCommunity state only when community.id actually changes
-  // This prevents infinite loops from Expo Router tab reconfiguration
-  useEffect(() => {
+  const hasCommunity = useMemo(() => {
     const currentCommunityId = community?.id;
     if (prevCommunityIdRef.current !== currentCommunityId) {
       prevCommunityIdRef.current = currentCommunityId;
-      setHasCommunity(!!currentCommunityId);
+      prevHasCommunityRef.current = !!currentCommunityId;
     }
+    return prevHasCommunityRef.current;
   }, [community?.id]);
 
   const tabs = (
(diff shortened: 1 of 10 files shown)


lilseyi:
@cursor Please review this PR. Focus on verifying that these three user-reported issues are actually fixed by the changes:

1. **"No internet connection" banner flash on app open (iOS)** — ConnectionProvider resets to `'connecting'` on foreground, WebSocket reconnects within 6s grace period silently
2. **Refresh flicker when opening the app** — Token refresh is now gated by JWT expiry check, skipping unnecessary refreshes that cascaded to 50+ query hooks  
3. **Maximum update depth exceeded (Sentry)** — Navigation effects now have `hasRedirected` ref guards, `parsedChannel` is memoized, tabs layout simplified

Also check for any regressions — especially race conditions in the ConnectionProvider AppState listener, edge cases in the JWT decode, and whether the `stableStringify` memoization key is robust.

**Important:** Do NOT suggest changes that would revert any of these three fixes. They are intentional and have been reviewed.

lilseyi:
@cursor Second review pass please. The previous review confirmed all 3 core issues are addressed. Now verify:

1. Your own changes (JWT base64 padding fix, ConnectionProvider test) don't introduce regressions
2. There are no remaining issues in the PR that need to be addressed
3. The test you added passes and covers the critical foreground reset path

If everything looks clean, confirm there are no remaining issues.

**Reminder:** Do NOT revert the token refresh gating, the foreground status reset, or the navigation ref guards — these are the intentional fixes.

PR #308, opened 2026-04-10
Cap users at 2 pending join requests (frontend stopgap)

lilseyi:
## Summary

Adds a frontend-only nudge that prevents users from accumulating more than **2 pending join requests** at a time within a single community. When a user at the cap taps "Request to join" on a 3rd group, a custom modal appears explaining the rule and pointing them at a new "My Requests" section on the profile page where they can withdraw something to free up a slot.

The backend stays unchanged — admins, leaders, and anyone else who legitimately needs to be in many groups is unaffected. The cap is purely a client-side friction nudge to address the problem Brittany described: too many people request to join 3+ dinner parties, leaders approve them to avoid delaying onboarding, and then those members ghost — making it hard for leaders to actually shepherd people.

### What's in scope

- **Convex query** ` listMyPendingJoinRequests` — returns the current user's pending requests (with group + group-type names) scoped to a single community.
- **`useMyPendingJoinRequests` hook** — thin wrapper exposing `requests`, `count`, `isAtLimit`. Cap (`PENDING_JOIN_REQUEST_LIMIT = 2`) lives here.
- **`PendingRequestLimitModal`** — custom modal built on the existing `CustomModal` primitive (not a native `Alert`), with two actions: "View my requests" and "Dismiss".
- **`MyRequestsSection`** — new section on the existing profile page (not a separate route) listing pending requests with per-row Withdraw buttons. Hidden when empty.
- **`GroupDetailScreen.handleJoinGroup`** — gates the join mutation behind the `isAtLimit` check.

### What's intentionally out of scope

- **`GroupPageClient` (`/g/[shortId]` share page)** — uses the public direct-join mutation, which creates active memberships rather than pending requests. The cap, by design, only counts pending requests so this codepath is unaffected.
- **`(auth)/join-flow`** — runs immediately after a community switch where pending count is 0 by definition. Not gated.
- **Backend cap** — explicitly avoided. Some users (admins, leaders) need to be in many groups; a backend rule would block them. Frontend-only is the right level.

### Product context

Decisions locked with the user across 11 product questions before implementation. Key choices:

- Cap = **2 pending requests, total, app-wide** (not per group type, despite an earlier draft that was per-type — simplified at the user's request).
- Active memberships **don't count**. Only pending requests do. Withdrawing or having a request resolved frees up a slot.
- Scope is **per current community**. A user can have 2 pending in church A and 2 pending in church B with no conflict.
- Modal is **custom-styled**, not a native system alert.
- "My Requests" lives as a **section on the existing profile page**, not its own route.

## Test plan

Automated coverage (29 new tests across 4 files, 7 backend + 22 frontend, plus 1 mock fix):

- [x] `apps/convex/__tests__/my-pending-join-requests.test.ts` — empty case, 1 request, 2 cross-type requests, ignores active memberships, ignores declined, scopes to community, ignores other users
- [x] `apps/mobile/features/groups/hooks/__tests__/useMyPendingJoinRequests.test.ts` — query args, count + isAtLimit at 0/1/2, skips on no-token / no-community, loading state, constant export
- [x] `apps/mobile/features/groups/components/__tests__/PendingRequestLimitModal.test.tsx` — visibility, copy, both button handlers
- [x] `apps/mobile/features/profile/components/__tests__/MyRequestsSection.test.tsx` — empty hides, populated renders, withdraw confirm + cancel
- [x] `apps/mobile/features/groups/components/__tests__/GroupDetailScreen.test.tsx` — 3 new cases: at-limit shows modal & blocks mutation, dismiss closes modal, below-limit submits normally
- [x] Full mobile suite passes (104 suites, 997 tests)

Manual verification (recommended before merging):

- [ ] Seed a user with 2 pending requests in Demo Community, navigate to a 3rd group, tap Request to Join → custom modal appears with correct copy
- [ ] Tap "View my requests" → navigates to profile, My Requests section visible with both pending requests
- [ ] Tap Withdraw on one → request disappears, count drops to 1
- [ ] Navigate back to the 3rd group, tap Request to Join again → request submits normally (no modal)
- [ ] Verify same flow on web build (`pnpm dev` then visit group URL)
- [ ] Verify the modal renders styled (not as a native iOS alert dialog)

### DIFF
--- a/apps/convex/__tests__/my-pending-join-requests.test.ts
+++ b/apps/convex/__tests__/my-pending-join-requests.test.ts
@@ -1,0 +1,369 @@
+/**
+ * Tests for listMyPendingJoinRequests query
+ *
+ * Powers the frontend "pending join request limit" feature: a user may have at
+ * most 2 pending join requests within a single community before the client
+ * blocks them from requesting a 3rd. This query returns the user's current
+ * pending requests so the UI can count them and surface a "My Requests"
+ * section on the profile page.
+ *
+ * Run with: cd convex && pnpm test __tests__/my-pending-join-requests.test.ts
+ */
+
+import { convexTest } from "convex-test";
+import { expect, test, describe } from "vitest";
+import schema from "../schema";
+import { api } from "../_generated/api";
+import { modules } from "../test.setup";
+import { generateTokens } from "../lib/auth";
+import type { Id } from "../_generated/dataModel";
+
+process.env.JWT_SECRET = "test-jwt-secret-for-unit-tests-minimum-32-chars";
+
+/**
+ * Seed a community with two group types and a test user.
+ */
+async function seedCommunityWithUser(t: ReturnType<typeof convexTest>) {
+  const timestamp = Date.now();
+
+  const communityId = await t.run(async (ctx) => {
+    return await ctx.db.insert("communities", {
+      name: "Test Community",
+      slug: "TEST001",
+      createdAt: timestamp,
+      updatedAt: timestamp,
+    });
+  });
+
+  const dinnerPartyTypeId = await t.run(async (ctx) => {
+    return await ctx.db.insert("groupTypes", {
+      communityId,
+      name: "Dinner Parties",
+      slug: "dinner-parties",
+      createdAt: timestamp,
+      isActive: true,
+      displayOrder: 0,
+    });
+  });
+
+  const teamTypeId = await t.run(async (ctx) => {
+    return await ctx.db.insert("groupTypes", {
+      communityId,
+      name: "Teams",
+      slug: "teams",
+      createdAt: timestamp,
+      isActive: true,
+      displayOrder: 1,
+    });
+  });
+
+  const userId = await t.run(async (ctx) => {
+    return await ctx.db.insert("users", {
+      firstName: "Test",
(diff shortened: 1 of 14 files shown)


lilseyi:
@codex review

Addressing your P1 in commit `cac4bfd`:

You correctly flagged that `useMyPendingJoinRequests` returns `isAtLimit=false` while loading (because `data === undefined` → `requests=[]` → length 0 < 2), which would let an at-cap user slip through during the loading window.

Two-layer fix:
1. **Disable the Join button** until the query resolves (`isJoining={joinGroupMutation.isPending || isPendingLimitLoading}`), so the click can't happen.
2. **Defensive early-return** in `handleJoinGroup` if `isPendingLimitLoading` is true, in case anything else triggers the handler.

Added a unit test (`blocks the join mutation while the pending-requests query is still loading`) covering the loading-state branch.

Also in this push: separate commit `bbe150b` fixes a Withdraw bug I caught during Playwright visual verification — `react-native-web@0.21.2` ships `Alert.alert` as a no-op, so the multi-button confirmation was silently doing nothing in the browser. Refactored `MyRequestsSection` to use the existing `ConfirmModal` (cross-platform).

Please re-review and verify both fixes are complete.

PR #317, opened 2026-04-17
feat(events): member-created events + moderation + Profile → My Events

lilseyi:
## Summary
Implements ADR-022 PR 2 — the member-creation half of the events-first refactor that started in PR #315 / #316.

- Any active community member can now create events. Leaders and admins keep their extra powers (series, community-wide events, chat posting).
- Non-leaders are capped at 1 future event across all groups (status ∈ scheduled/confirmed AND scheduledAt > now). Cancelled don't count. Enforced at mutation time via the existing `meetings.by_createdBy` index; the sequential cap test shows this holds — Convex per-document serialization covers the race.
- Location is required-but-flexible: new `locationMode` enum on `meetings` (`address | online | tbd`). Validation applies uniformly to leaders too — tightens leader UX for consistency.
- New `meetingReports` table (mirrors `chatMessageFlags`) + mutations. Reports route to the event's group leaders; stub review query for the follow-up UI.
- `notifyEventCreatedByMember` internal action fires only when a non-leader creates AND the group is not the announcement group (would spam admins otherwise).
- Ownership on leave: if a creator leaves their community, `createdById` on their future meetings silently patches to the primary admin (ADR-010).
- Profile → My Events: new `/(user)/my-events` with Hosted / Attended segments. Upcoming above past, newest-first. Community-wide events render as single grouped cards matching the Events tab.
- `CreateEventScreen` swaps `useLeaderGroups` → `useCreatableGroups`, defaults to the announcement group for members ("Community · Hosted by you"), hides Series / CWE toggles for non-leaders, adds a "Location TBD" switch, and gates the submit with the 1-future-event cap.
- `ReportEventSheet` (new) behind a flag icon on the event detail header.
- Analytics (PostHog): `event_created_by_member`, `event_reported`, `event_deleted_by_leader`.
- Seed: new "Mia Member" test user for reviewers to exercise the member path end-to-end.

All new Convex hooks live under `features/events/` so the PR #316 query-patterns regression test keeps them in scope without widening its scan.

## Test plan
- [x] New `__tests__/member-created-events.test.ts` — 14 tests covering member-create, cap, location validation, update/cancel perms, reports round-trip, leave-community ownership transfer.
- [x] All 1303 Convex tests pass (76 files).
- [x] All 1002 mobile Jest tests pass (105 files, snapshot refreshed).
- [x] `features/events/__tests__/query-patterns.test.ts` still green (regression guard from PR #316).
[...]

PR #318, opened 2026-04-17
feat(events): let creators message attendees, not just leaders

lilseyi:
## Summary
Follow-up to #317 (ADR-022). The Message Attendees button and blast log on the event detail screens were gated on `isLeader` — creators of a member event couldn't reach the people who RSVPed to their own event, which is exactly the audience the host wants to reach.

- **Backend**: `eventBlasts.initiate` and `eventBlasts.list` now use `canEditMeeting` (creator | group leader | community admin), matching the permission set ADR-022 already uses for update/cancel.
- **Client**: both event detail surfaces show Message Attendees + EventBlastHistory when the viewer is the creator OR a leader.
  - `EventPageClient` (public share page) → `canEdit`
  - `EventDetails` (leader-tools flow) → `isLeader || isCreator`
- **Test**: updated the stale "Only group leaders" error-string assertion in `event-blasts.test.ts`. 1308 convex tests pass.

## Test plan
- [x] All 1308 convex tests pass
- [x] Mobile eslint clean
- [ ] Verify creator sees Message Attendees in the browser after merge + deploy

### DIFF
--- a/apps/convex/__tests__/event-blasts.test.ts
+++ b/apps/convex/__tests__/event-blasts.test.ts
@@ -231,7 +231,7 @@ describe("Event Blasts", () => {
             channels: ["push"],
           }
         )
-      ).rejects.toThrow("Only group leaders");
+      ).rejects.toThrow(/event creator, group leaders/i);
     });
   });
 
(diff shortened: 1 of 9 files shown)


lilseyi:
@codex review

PR #327, opened 2026-04-20
fix(people): return communityPeopleId from quickAddRow so detail sheet opens after add

lilseyi:
## Summary

- Adding a person via the side sheet crashed the detail view with `ArgumentValidationError: Found ID "..." from table groupMembers, which does not match the table name in validator v.id("communityPeople")`.
- Existing rows worked because `adaptCommunityPerson` aliases `cp._id` as `groupMemberId`, so `item.groupMemberId` was actually a communityPeople ID. The Add flow had no such adapter step — it returned a real `groupMembers` ID, and callers passed it straight to `communityPeople.history` as a `communityPeopleId`.
- `quickAddRow` now synchronously ensures a `communityPeople` record exists for the added user in the current group and returns its ID. The scheduled `upsertFromSubmission` still fills in denormalized score/status fields once background score computation completes.

## Test plan

- [ ] In the People tool, click **Add Person**, fill in required fields, click **Add person** — confirm the detail side sheet opens without the "Something went wrong" error.
- [ ] On mobile, confirm the same flow navigates to `/(user)/leader-tools/{groupId}/followup/{communityPeopleId}` and renders detail.
- [ ] Confirm the new row also shows in the People list (no regression in the existing reactive update).
- [ ] Reported by Manny (FOUNT) — verify with him after deploy.

### DIFF
--- a/apps/convex/functions/memberFollowups.ts
+++ b/apps/convex/functions/memberFollowups.ts
@@ -2990,8 +2990,47 @@ export const quickAddRow = mutation({
       );
     }
 
-    // Schedule communityPeople upsert so the People table is immediately populated
+    // Ensure a communityPeople record exists synchronously so the client can
+    // open the detail sheet with a valid communityPeopleId immediately.
+    // The scheduled upsertFromSubmission below will fill in denormalized
+    // score/status fields once the background score computation completes.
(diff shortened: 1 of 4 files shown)


lilseyi:
@chatgpt-codex-connector please review

PR #333, opened 2026-04-23
fix(notifications): remove fallback crons that race with primary scheduler

lilseyi:
## Summary

User reported receiving 3 separate "did you attend?" pings for a single event. Root cause is a dual-path scheduling pattern that races with itself.

- **Removed** `meeting-reminder-fallback` and `attendance-confirmation-fallback` crons. They ran every 15 min sweeping for unsent meetings in a 20-min window. Because the send handlers do check-then-send-then-mark non-atomically, the fallback could fire while the primary scheduled job was still dispatching, producing duplicates. The fallbacks were originally added as a backfill for migrated data with no scheduled job attached — that data no longer exists, so the dual path now only causes duplicates with no benefit.
- **Fixed** `initializeGroupAfterCreation` (community-wide event meetings spawned during group creation) to persist `reminderJobId` and `attendanceConfirmationJobId`. Previously discarded, leaving rescheduled events with orphan jobs that fire on top of the new ones.
- **Removed** the `by_reminderAt_sent` and `by_attendanceConfirmation` indices on `meetings` — they only existed to serve the deleted fallback queries.

## Audit

Spawned an exploration of the remaining 12 crons in `apps/convex/crons.ts`. None have the same dual-path-with-fallback anti-pattern. Bot crons (birthday, task reminder, communication) advance `nextScheduledAt` atomically as part of the send, so they self-bucket without needing a sweeper. Cleanup/recomputation crons don't dispatch notifications.

(One unrelated minor finding: the Slack `nag-check` has a theoretical check-then-set window in `slackServiceBot/configDb.ts`. Once-per-hour-per-config so duplicates are extremely unlikely. Out of scope here.)

## Test plan

- [x] `tsc --noEmit` clean
- [x] Convex test suite green (1031 passed)
- [ ] Manual: create a meeting in dev, confirm reminder fires exactly once
- [ ] Manual: create a meeting via group creation flow (community-wide event), reschedule it, confirm only the new reminder fires (orphan fix)
- [ ] Production observation: monitor next week for any reports of duplicate pings

### DIFF
--- a/apps/convex/crons.ts
+++ b/apps/convex/crons.ts
@@ -42,32 +42,6 @@ crons.hourly(
 );
 
 // =============================================================================
-// MEETING REMINDER CHECK (Fallback)
-// =============================================================================
-// Runs every 15 minutes as a fallback to catch any meeting reminders that
-// weren't scheduled at creation time (e.g., migrated data).
-// Primary scheduling happens in meetings.create mutation via ctx.scheduler.runAt()
-
-crons.cron(
-  "meeting-reminder-fallback",
-  "*/15 * * * *", // Every 15 minutes
-  internal.functions.scheduledJobs.processMeetingReminderFallback
-);
-
-// =============================================================================
-// ATTENDANCE CONFIRMATION CHECK (Fallback)
-// =============================================================================
-// Runs every 15 minutes as a fallback to catch any attendance confirmations that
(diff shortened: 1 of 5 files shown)


lilseyi:
@codex review

lilseyi:
@codex review

PR #334, opened 2026-04-23
fix(events): keep events visible and RSVPable for 3h after start

lilseyi:
## Summary

Late arrivals were locked out of events the moment \`scheduledAt\` passed: cards vanished from the events list, the detail page flipped to "This event has passed", and the RSVP mutation rejected with "Cannot RSVP to past event."

Adds a \`PAST_EVENT_BUFFER_MS\` of 3 hours and applies it in four places:

- \`apps/convex/functions/meetings/events.ts\` — \`nextUp\`, \`thisWeek\`, and \`loadMyEventsMeetings\` filters
- \`apps/mobile/app/e/[shortId]/EventPageClient.tsx\` — \`isPastEvent\` derived flag
- \`apps/convex/functions/meetingRsvps.ts\` — server-side gate

The constant lives in \`apps/convex/lib/meetingConfig.ts\`. The mobile side mirrors the literal locally with a comment pointing at the source — keeping the cross-package import surface flat.

### Side effect: fixes a card-time bug for community-wide events

When a parent CWE has children with overridden times (real example today: 19 children at 7:00 PM, 1 at 7:30 PM), the events list aggregates to the chronologically earliest surviving child. Without the buffer, at 7:19 PM the 7:00 children were already filtered out, leaving only the 7:30 override — so the card surfaced 7:30 PM even though most users' actual meeting was at 7:00. The buffer keeps the 7:00 children in the candidate set, so the card now shows the correct earliest time.

## Test plan

- [x] \`tsc --noEmit\` clean on convex package
- [x] Mobile typecheck — only pre-existing unrelated errors (AudioPlayer, MessageInput, tasks tests, NotificationProvider); nothing new from this PR
- [x] Pre-push test suite green
- [ ] Manual: open the events tab between an event's \`scheduledAt\` and \`scheduledAt + 3h\`, confirm card stays visible
- [ ] Manual: open the event detail page within that window, confirm RSVP still works
- [ ] Manual: 3h+1m after \`scheduledAt\`, confirm card disappears and RSVP is blocked

### DIFF
--- a/apps/convex/functions/meetingRsvps.ts
+++ b/apps/convex/functions/meetingRsvps.ts
@@ -14,6 +14,7 @@ import { internal } from "../_generated/api";
 import type { Id } from "../_generated/dataModel";
 import { now, getMediaUrl } from "../lib/utils";
 import { requireAuth, getOptionalAuth } from "../lib/auth";
+import { PAST_EVENT_BUFFER_MS } from "../lib/meetingConfig";
 import {
   getMaxGuestsForMeeting,
   isGoingOption,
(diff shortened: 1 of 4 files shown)


lilseyi:
@codex review

PR #341, opened 2026-04-25
fix(events): My Events uses hostUserIds + stop minute flicker

lilseyi:
## Summary
- **My Events filter:** switch from \`createdById\`-based hosting to the shared \`isMeetingHost\` / \`getHostUserIds\` helpers in \`lib/meetingPermissions.ts\`. A meeting now lands in My Events when (a) the viewer RSVP'd, (b) the viewer is in \`hostUserIds\`, or (c) \`hostUserIds\` is empty AND the viewer is an active leader of the hosting group. Same primitives \`canAccessEventChannel\` and \`canEditMeeting\` use, so chat-access / edit-permission / My Events stay coherent.
- **Stop minute-cadence flicker:** \`useEventsByTimeWindow\` now stickies the last successful payload via a ref. A transient \`undefined\` from the Convex subscription (websocket reconnect, etc.) re-renders with the prior data instead of swapping the whole scroll view to an ActivityIndicator. \`isLoading\` only fires before any data has arrived.

## Why this matters
- Pre-fix, a filer who delegated their event by removing themselves from \`hostUserIds\` still saw it in My Events; conversely a true host who took ownership of someone else's event didn't appear.
- The flicker was visible roughly every minute on the live Events tab and was confusing.

## Test plan
- [x] \`npx tsc --noEmit\` clean (apps/convex, apps/mobile)
- [x] \`vitest\` — 42 host/member tests pass (\`meeting-hosts.test.ts\`, \`member-created-events.test.ts\`)
- [x] Browser test (Playwright): signed in as Demo, Events tab loads with My Events showing the user's RSVP'd/hosted CWE children. MutationObserver tracking over 80+ seconds logged 0 spinner appearances and 0 "My Events" disappearances.
- [ ] iOS sim sanity check before merge — flicker root causes specific to native (iOS suspending the WebSocket on memory pressure) won't reproduce on web

### DIFF
--- a/apps/convex/functions/meetings/events.ts
+++ b/apps/convex/functions/meetings/events.ts
@@ -455,12 +455,25 @@ async function loadMyEventsMeetings(
     await Promise.all([...userRsvpMeetingIds].map((id) => ctx.db.get(id)))
   ).filter((m): m is Doc<"meetings"> => m !== null);
 
+  // Scope the per-group fan-out to groups in the requested community only.
+  // For users who belong to multiple communities, iterating every membership
+  // would scan future meetings across unrelated communities just to drop
+  // them in the in-memory `communityId === communityId` filter below.
+  // Fetching the user's group docs first is cheap (indexed by id) and lets
+  // us skip those reads.
+  const userGroupDocs = await Promise.all(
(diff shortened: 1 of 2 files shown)


lilseyi:
@codex please review the latest commit

lilseyi:
@codex please review the latest commit

PR #340, opened 2026-04-24
Decouple event host from creator + revamp inbox visibility

lilseyi:
## Summary
- **Hosts decoupled from creators on events.** New `meetings.hostUserIds` is the canonical owner; creator becomes pure metadata. Empty list = "delegated to group leaders." Frontend `HostsPicker` lets admins/creators add/remove hosts, with a self-removal confirmation. Creator is no longer surfaced anywhere — all "Hosted by" UI reads `hosts` from `getWithDetails` / `getByShortId`.
- **Permissions, chat seating, and RSVP notifications all route through hosts.** `canEditMeeting`, `canAccessEventChannel`, RSVP notification recipients, and event-chat admin seating all use `getHostUserIds` with leader fallback when empty. `reconcileEventChannelAdmins` keeps chat membership in sync when hosts change.
- **Inbox events now gated on real activity.** `getInboxChannels` only returns event channels with `lastMessageAt > 0` AND within the new `INBOX_EVENT_HIDE_AFTER_MS` (2 days). Mobile client stops partitioning events above all groups; events are top-level rows interleaved with groups by recency, announcement group still pinned.
- **Misc UX/QA fixes shipped along the way:** `showAlert` helper so RN `Alert.alert` errors actually appear on web; `messages.getMessages` returns empty page (instead of throwing) when an event-channel viewer loses access; community-wide toggle reworded + moved to bottom of the form; RSVP option emoji rendering repaired (label parsing).

## Test plan
- [x] `cd apps/convex && npx vitest run` — 1370/1370 passing (added 18 host-logic tests, updated 4 obsolete-creator tests)
- [x] `cd apps/convex && npx tsc --noEmit` — clean
- [x] `cd apps/mobile && npx tsc --noEmit` — clean for files touched by this branch (9 pre-existing errors elsewhere)
- [x] `pnpm test --filter=mobile` — 1030/1039 passing (9 skipped; +5 fixed by adding `groupMembers.list` + `getChannelStateForEditor` to test mock)
- [x] **Browser QA on hushed-lemur-239** (sub-agent walkthrough):
  - Login → create event → default host = creator ✓
  - Edit → add second host → "Hosted by X + 1 other" renders ✓
  - Edit → remove all hosts (delegate) → group attribution only, no creator surfaced ✓
  - Share page (`/e/<short>`) renders hosts identical to leader-tools detail ✓
[...]

PR #343, opened 2026-04-27
fix(events): cap counts hosted events, not created events

lilseyi:
## Summary
- The non-leader 1-future-event cap counted by `createdById`, but "My Events" surfaces (Events tab carousel + Profile → My Events Hosted) count by `hostUserIds` since the host/creator decoupling (PR #340 / commit `c7fcf32`).
- Filing an event for someone else hit the cap even though the user wasn't actually hosting anything — the event didn't show up in their "My Events," but they were still blocked from creating another. Confusing.
- Switch the cap, `myHostedEvents`, and the `myAttendedEvents` dedupe filter to use `hostUserIds` so all three surfaces agree on what "hosted" means.
- Cap now also skips entirely when the new event delegates hosting (creator not in `hostUserIds`) — filing on someone else's behalf doesn't count against your own hosting load.
- Error copy updated: "You're already hosting an upcoming event…" (was "You already have an upcoming event…").

## Test plan
- [x] `pnpm test` in `apps/convex` — all 1375 tests pass (was 1373; added 2)
- [x] `pnpm test` in `apps/mobile` — all 1030 tests pass
- [x] New tests cover the two key paths:
  - non-leader can file an event for someone else (cap doesn't fire when creator isn't in the new `hostUserIds`)
  - non-leader is still capped when adding self as co-host on a second event
- [ ] Manual verify: log in as a member, host one future event, then create a second one with a different host picked → form submits without the banner.

### DIFF
--- a/apps/convex/__tests__/member-created-events.test.ts
+++ b/apps/convex/__tests__/member-created-events.test.ts
@@ -254,6 +254,60 @@ describe("meetings.create — member flow", () => {
   // "test began while previous transaction was still open". The cap
   // guarantee comes from the runtime invariant, not from a racy test.
 
+  test("non-leader can file event for someone else without hitting cap", async () => {
+    // Cap counts events the user is *hosting* (in hostUserIds), not events
+    // they merely created. A member who already hosts one future event can
+    // still file a separate event for someone else as long as they don't
+    // include themselves in the new event's hostUserIds.
+    const t = convexTest(schema, modules);
+    const s = await seed(t);
+
+    // Member hosts their own future event (default hostUserIds = [creator]).
+    await t.mutation(api.functions.meetings.index.create, {
+      token: s.memberToken,
+      groupId: s.groupId,
+      scheduledAt: FUTURE(),
+      meetingType: 1,
+      locationMode: "tbd",
+    });
+
+    // Member files a second event with someone else as host — should succeed.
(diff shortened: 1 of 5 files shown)


lilseyi:
@codex please review this PR

lilseyi:
@codex please review this PR

lilseyi:
Both flagged paths (`countFutureEventsHostedBy` and `myHostedEvents`) use the `by_community_scheduledAt` compound index with a `gt(scheduledAt, now)` range, so the scan is **bounded by future events in this community**, not all meetings ever. The in-memory `isMeetingHost` filter follows because `hostUserIds` is an array field and Convex doesn't support multi-value indexes — the existing inline comment in the code already documents this constraint.

For typical community sizes (low hundreds of future events at any time), this is well below Convex's read limits and the response stays sub-200ms. It would only become a real problem at thousands of concurrent future events per community, which Togather doesn't see today.

The architecturally cleaner fix is a `meetingHostMembers` junction table (rows of `{ meetingId, userId, scheduledAt, status, communityId }` with a `by_user_status_scheduledAt` index) — that turns both code paths into a single index lookup proportional to the *user's* hosted set. Sized as a follow-up rather than blocking this PR, since it touches schema and the cap-counts-hosted bug is the immediate need.

Tracking as a follow-up. Resolving these threads.

PR #345, opened 2026-04-28
feat(chat): DM inbox surface + start-chat picker (PR 2 of 5)

lilseyi:
## Summary

PR 2 of 5. **Stacked on #344** — review/merge that first.

User-facing surface for the DM feature scaffolded in PR 1. Ships **1:1 DMs only**; the group-chat picker is deferred to PR 4, the request-flow UI (accept/decline banner, requests inbox) to PR 3.

## What ships

- **Compose button on the inbox header.** Tapping it opens `/inbox/new`.
- **`/inbox/new` (StartChatScreen)** — debounced search input across the caller's shared communities. Each row shows avatar + name + the community attribution. Tapping a row creates (or fetches) the 1:1 DM channel and replaces the route to `/inbox/dm/{channelId}`. Errors from the backend (no shared community / blocked / rate-limit) are surfaced inline.
- **`/inbox/dm/[channelId]`** — new route hosting the existing `ConvexChatRoomScreen`. Reuses the message list, composer, reactions, etc. instead of forking. Recipient's name and avatar are passed as URL params so the chat header has something to show before the channel doc loads.
- **"Direct messages" section in the inbox** — accepted ad-hoc channels appear above groups + events, populated by the new `getDirectInbox` query. Pending requests stay surfaced separately via `listChatRequests` in PR 3.

## Backend touches

Two minimal corrections to PR 1's wiring:

- `getChannel` was returning `null` for ad-hoc channels (the null-check sweep was overcautious for the new schema). Replaced with an explicit ad-hoc branch that does its own membership check and returns the channel.
- `ConvexChatRoomScreen`'s `isEssentialDataReady` gate required `resolvedGroupId`, which is null for ad-hoc DMs. Now passes through when the channel doc has loaded with `isAdHoc: true`.

## Test plan

- [x] `npx tsc --noEmit` passes for both `apps/convex` and `apps/mobile` (no new errors; 10 pre-existing Node-typings + theme errors untouched).
- [x] **387/387 messaging tests pass** in `apps/convex` (no regressions from the `getChannel` change).
- [x] Lint-staged + eslint clean on commit.
[...]

PR #344, opened 2026-04-28
feat(chat): foundation for ad-hoc DMs and group chats (backend, no UI)

lilseyi:
## Summary

PR 1 of 5 in a planned series introducing **direct messages and ad-hoc group chats** to Togather. This PR is backend-only — schema, mutations/queries, request-state wiring into existing chat, crons, and tests. **No UI in this PR**; the inbox surface, picker, and request banner ship in PRs 2–5.

The feature reuses the existing `chatChannels` infrastructure rather than building a parallel system. New ad-hoc channels live alongside group channels with `isAdHoc: true`, `channelType: "dm" | "group_dm"`, and a `communityId` instead of a `groupId`. Members go through a Signal-style **Message Request flow** (`requestState: "pending" | "accepted" | "declined"`) before the chat is fully usable on the recipient's side.

## Product decisions baked in

| Decision | Rationale |
|---|---|
| Even community co-members go through a Message Request | User chose Signal-grade safety for V1; no instant-DM tier yet |
| 1:1 + group chats ship together | Cohesive launch; group consent inherits the same per-member request flow |
| No phone-contacts integration | Deferred to V2; community-scoping is the trust gate for V1 |
| No cross-community DMs | Closes the stranger-spam vector entirely; no global user search |
| Pending requests **do** push (not silent) | Silent-request is a stranger-spam defense; we've closed that vector at the gate (community-only + 5-init/24h + 1-msg-per-pair-per-24h). For a community app, silent would mean legitimate first messages rot in an unchecked inbox. |

## Schema deltas

- `chatChannels`: `groupId` now optional; new `communityId`, `isAdHoc`, `dmPairKey`. Indexes `by_dmPairKey`, `by_community_isAdHoc`. Invariant: exactly one of `groupId` / `communityId` is set (enforced in mutations).
- `chatChannelMembers`: `requestState`, `requestRespondedAt`, `invitedById`. Indexes `by_user_requestState`, `by_requestState_joinedAt` (cron access pattern).
- New table `directMessageRateLimits` for the 1-msg-per-pending-pair-per-24h rule.

The schema change to `groupId` cascades through 11 existing files. Each call site that read `channel.groupId` got an explicit runtime guard (early throw or filter-out for ad-hoc rows) — the assumption that "this code path only handles group channels" is now documented in the code rather than hidden in the type.

## New backend functions (`apps/convex/functions/messaging/directMessages.ts`)

[...]

PR #346, opened 2026-04-28
feat(chat): request flow UI for ad-hoc DMs (PR 3 of 5)

lilseyi:
## Summary

PR 3 of 5. **Stacked on #345** which is stacked on #344. Review/merge in order.

This is the missing piece that turns the Message Request model from a backend invariant into something a recipient can actually act on — Accept / Decline / Block & Report.

## What ships

- **`/inbox/requests` (ChatRequestsScreen).** List of pending requests with inviter info + a one-line preview. Tap → a bottom sheet shows the full first message and three big buttons. Empty state for "no requests."
- **`ChatRequestBanner` mounted in `ConvexChatRoomScreen`.** When the caller's membership row is `requestState: "pending"`, the banner replaces the message composer — they can't reply until they accept, and the explanation copy makes the gate readable. Decline/block route back to the inbox.
- **"Message Requests (N pending)" row at the top of the inbox.** Tappable, routes to `/inbox/requests`. Deliberately not interleaved with active conversations — the whole point of the request model is to keep unaccepted senders out of the primary list.

## Backend touch

`getChannel` now returns `myRequestState` and `inviterDisplayName` for ad-hoc channels. Group + event branches return them as `undefined` so the Convex-generated TS union resolves cleanly (was breaking `channelData?.myRequestState` access until I made the shape consistent across branches). Backward compatible.

## Test plan

- [x] `npx tsc --noEmit` passes for both `apps/convex` and `apps/mobile` (no new errors).
- [x] **387/387 messaging tests pass** in `apps/convex` after the `getChannel` extension — no regressions.
- [x] Lint-staged + eslint clean on commit.
- [ ] **Manual smoke test (reviewer):** with seeded users A and B in the same community, A starts a DM with B and sends "hi". B logs in, sees "Message Requests (1 pending)" at the top of inbox, taps it, sees the request, taps Accept — channel opens with the message visible, B can reply. Repeat with Decline (channel disappears from B's view, sender not notified) and Block & Report (writes block + flag, banner dismisses). Confirm the chat room banner appears when navigating directly to a still-pending DM URL.

## What's NOT in this PR (later PRs)

- Group-chat creation (`group_dm` picker) → PR 4
- Read-receipt suppression while pending, lock-screen copy polish, full PostHog flag ramp → PR 5

### DIFF
--- a/apps/convex/functions/messaging/channels.ts
+++ b/apps/convex/functions/messaging/channels.ts
@@ -251,7 +251,11 @@ export const getChannel = query({
       if (adHocMembership.invitedById) {
         const inviter = await ctx.db.get(adHocMembership.invitedById);
         if (inviter) {
-          inviterDisplayName = `${inviter.firstName ?? ""} ${inviter.lastName ?? ""}`.trim();
+          // Use the shared `getDisplayName` helper so unset firstName +
+          // lastName don't produce an empty string that would render as
+          // " would like to chat with you." in the request banner.
+          const resolved = getDisplayName(inviter.firstName, inviter.lastName);
+          inviterDisplayName = resolved.trim().length > 0 ? resolved : "Someone";
         }
       }
       return {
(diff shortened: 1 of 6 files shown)


lilseyi:
@codex please review this PR

PR #352, opened 2026-04-28
feat(chat): chat info, profile entry points, leaders preview, profile-photo gate

lilseyi:
## Summary

Big polish pass on the DM/group-chat surface that landed in PRs 346–351, plus new entry points to start a DM from outside the inbox.

### Chat-room polish
- **Tappable header** replaces the bare "Chat" title — stacked-avatar cluster + member name list + info icon, all routing to a new chat-info screen.
- **Privacy card** above the message list reminds members the chat is not end-to-end encrypted and admins can request access.

### Chat info screen (`/inbox/dm/<channelId>/info`)
- Stacked-avatar hero, member rows with **View profile** / **Remove from chat** actions.
- **Add people** sheet (reuses search picker), **Rename** (group_dm only, `dm` rejects), **Leave chat**.
- Backed by four new mutations in `directMessages.ts`:
  - `renameAdHocChannel`, `addAdHocMembers` (capped at 20), `removeAdHocMember`, `leaveAdHocChannel`.

### New entry points to DMs
- **Profile screen** — Message CTA opens or creates a DM with the viewed user, scoped to the active community. Hidden behind the `direct-messages` flag.
- **Group members list** — action sheet adds **View profile** as the first item, so non-admins also have a meaningful action.
- **Group detail page** — non-members now see a horizontally scrolling **LEADERS** section (max 8) so they can tap a leader and message them before joining. New public `groupMembers.getLeaderPreview` query.

### Profile-photo hard requirement
- Backend throws `PROFILE_PHOTO_REQUIRED` (caller) / `RECIPIENT_PROFILE_PHOTO_REQUIRED:<userId>` (recipient) at `createOrGetDirectChannel`, `createGroupChat`, and `respondToChatRequest` accept-path.
- Frontend shows `RequireProfilePhotoSheet` that routes to `/(user)/edit-profile`.
- Catches both the local check (current user state) and the backend echo for stale-state edges.

### Push notification fix
- Ad-hoc channels no longer carry the bogus "Group Chat: General" subtitle.
- Pending request: title=`<sender>`, body=`would like to chat: <preview>`.
- Accepted group_dm: title=`<sender> in <channelName | first-two-others>`.
- Accepted 1:1 dm: title=`<sender>`, body=`<preview>`.

### Stacked-avatar component
- Generalized to 1 / 2 / 3 / 4+ tiers with a `+N` overflow badge.
- Parameterized `size` so the same component drives the inbox row (56pt), chat-room header (36pt), and info-screen hero (96pt).
[...]

PR #353, opened 2026-04-28
fix(chat): four DM follow-ups — search index, leave+restart, photo gate, modal nav

lilseyi:
## Summary

Four follow-ups after #352 shipped to staging, all in one PR per request.

### 1. iOS modal nav broken
Tapping **Message** on a profile pushed the chat *behind* the profile modal. The `(user)` route group is a modal in `app/_layout.tsx`; pushing inside the modal lands the chat in the parent stack but the modal stays on top.

Fix: `useStartDirectMessage` now calls `router.dismissAll()` (when `canDismiss` is true) before `router.push`, so the modal closes and the chat takes the foreground.

### 2. Leave + restart broke the channel
A user who left a 1:1 DM and tried to restart the conversation hit `Server Error: Not a member of this channel` from `markAsRead`. `createOrGetDirectChannel` found the existing channel via `dmPairKey` and returned it without clearing the caller's `leftAt`.

Fix: new `reactivateAdHocMembership` helper called on the existing-channel branch — clears `leftAt` and bumps `requestState` to `"accepted"` for the caller. Idempotent for never-left rows. New regression test asserts the round-trip.

### 3. User search was slow
`searchUsersInSharedCommunities` scanned up to 2000 community memberships and fetched each user sequentially.

Fix: replaced with the `search_users` full-text index plus a phone-number fallback for digit-heavy queries (mirrors the admin search at `apps/convex/functions/admin/members.ts:84+`). Empty queries now return `[]` explicitly — the picker already shows nothing until typing.

### 4. Photo gate too aggressive
Inviting somebody without a profile photo failed with `RECIPIENT_PROFILE_PHOTO_REQUIRED`. The request flow should still go through — the gate belongs at *accept-time*, not *invite-time*.

Fix: removed the recipient-side gate from `createOrGetDirectChannel`, `createGroupChat`, and `addAdHocMembers`. The accept path in `respondToChatRequest` still enforces — you cannot read or reply without a photo. Updated test to assert request now succeeds.

## Test plan

- [x] Convex unit: 31/31 pass in `directMessages.test.ts` (1 new for leave+restart). 1406/1406 across full suite.
- [x] Convex `tsc --noEmit`: clean.
- [x] Mobile `tsc --noEmit`: zero new errors in any modified file.
- [ ] iOS simulator smoke: profile → Message → opens chat in foreground; leave 1:1 → restart from inbox/picker → no `markAsRead` error; user picker types fast and returns relevant matches.

### DIFF
--- a/apps/convex/__tests__/messaging/directMessages.test.ts
+++ b/apps/convex/__tests__/messaging/directMessages.test.ts
@@ -1007,6 +1007,13 @@ describe("profile photo gate", () => {
     );
     expect(aRowAfterRestart?.leftAt).toBeUndefined();
     expect(aRowAfterRestart?.requestState).toBe("accepted");
+
+    // The leave path archived the channel (Alice was the last accepted
+    // member). Reactivation must unarchive it so the thread reappears in
+    // getDirectInbox — otherwise the chat is reachable only via direct
+    // nav and missing from inbox surfaces.
+    const channelAfter = await t.run((ctx) => ctx.db.get(channelId));
+    expect(channelAfter?.isArchived).toBe(false);
   });
 });
 
--- a/apps/convex/__tests__/messaging/directMessages.test.ts
+++ b/apps/convex/__tests__/messaging/directMessages.test.ts
@@ -1014,6 +1014,54 @@ describe("profile photo gate", () => {
     // nav and missing from inbox surfaces.
     const channelAfter = await t.run((ctx) => ctx.db.get(channelId));
     expect(channelAfter?.isArchived).toBe(false);
+    // memberCount restored after rejoin (leaveAdHocChannel decremented it
+    // — without restore the 1:1 channel would persistently report 1).
+    expect(channelAfter?.memberCount).toBe(2);
+  });
+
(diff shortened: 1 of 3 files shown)


lilseyi:
@codex please review this PR

lilseyi:
@codex please re-review — both addressed in 1a1bf78:
- **P1-A**: `createOrGetDirectChannel` now also clears `isArchived` + `archivedAt` on the existing-channel branch when the channel was archived. Test asserts the unarchive.
- **P1-B**: search-index pre-filter cap bumped from `limit*4` to 500 to match the admin search.

lilseyi:
@codex please re-review — both addressed in 6f943be:
- **P1** (preserve pending): `reactivateAdHocMembership` now only un-leaves rows with `leftAt` set; pending/declined rows are never auto-promoted. New test for the pending-recipient-uses-Message-entry-point case.
- **P2** (memberCount drift): rejoin branch now increments `memberCount` when reactivation actually flipped a row.

lilseyi:
@codex please re-review — addressed in 2bab303. `reactivateAdHocMembership` now only un-leaves rows whose previous `requestState` was already `accepted`. Declined and pending rows are never touched. New regression test for the declined-then-Message case.

PR #356, opened 2026-04-28
fix(chat,admin): modal body, top channels naming + 20, hard photo gate

lilseyi:
## Summary

Three follow-up issues from staging.

### 1. Add People modal body collapsed on iOS
PR #354 fixed the modal width (was a thin column on native) but the inner \`ScrollView { flex: 1 }\` then had no flex-distributed height to fill, so the body collapsed to 0pt, showing only the title. Switched to \`flexShrink: 1\` so the modal grows with its children up to \`maxHeight: 90%\`, then scrolls.

### 2. Top Channels in admin
- Slice 10 → 20 (DMs and group_dms share the list, so the window needed to widen).
- For ad-hoc channels (no \`groupId\`), fall back to a comma-list of active member display names + a \`+N\` overflow tag, so rows have a meaningful label instead of an empty string. Pseudo-group \"Direct message\" / \"Group chat\" replaces the missing \`groupName\` subtitle.

### 3. Profile photo gate too leaky
A user could create or accept a chat with a photo, then remove the photo and keep messaging. Now enforced on every interaction:
- \`sendMessage\` (ad-hoc branch) throws \`PROFILE_PHOTO_REQUIRED\` after the membership check.
- \`markAsRead\` silently no-ops when ad-hoc + no photo (so the chat-room auto-mount doesn't crash; UI banner gates the composer).
- Frontend chat-room composer is replaced with an \"Add a profile photo to message\" banner + \"Add photo\" CTA when the local user has no \`profile_photo\`. Highest-priority banner branch — sits above the existing pending-request banner.

## Test plan

- [x] Convex unit: 1408/1408 pass; 33/33 in \`directMessages.test.ts\`.
- [x] Mobile + Convex \`tsc --noEmit\`: clean.
- [ ] iOS smoke: open Add People modal — body shows search + buttons; admin Top Channels — DMs labeled with member names; remove profile photo — chat composer replaced with photo banner; backend rejects send with PROFILE_PHOTO_REQUIRED.

### DIFF
--- a/apps/convex/functions/admin/stats.ts
+++ b/apps/convex/functions/admin/stats.ts
@@ -1349,9 +1349,12 @@ export const getDailySummary = query({
         if (topDirectScored.length < 20) topDirectScored.push(score);
       }
     }
-    // Backwards-compat alias for the original combined `topChannels` field
-    // returned to clients that haven't picked up the split yet.
-    const topScored = [...topGroupScored, ...topDirectScored].slice(0, 20);
+    // Legacy combined `topChannels` field — must be the GLOBAL top 20 by
+    // engagement score, not group-then-dm concatenation. Otherwise older
+    // client builds that still read `topChannels` see a biased list (e.g.
+    // a high-engagement DM excluded behind 20 lower-scoring group rows),
+    // breaking the original "top channels by engagement" contract.
+    const topScored = channelScores.slice(0, 20);
 
     // Count users active that day via users.lastActiveAt index
     const activeUsers = await ctx.db
--- a/apps/convex/functions/admin/stats.ts
+++ b/apps/convex/functions/admin/stats.ts
@@ -1413,14 +1413,19 @@ export const getDailySummary = query({
       };
     };
 
-    const topChannels = await Promise.all(topGroupScored.map(resolveChannelRow));
+    // Resolve names for each split + the legacy combined list. The
+    // combined `topChannels` is preserved as the API contract older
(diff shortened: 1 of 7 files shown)


lilseyi:
@codex please review this PR

lilseyi:
@codex please re-review — added one more fix in 00c77dd: hide empty ad-hoc channels (no `lastMessageAt` / no first message) from `getDirectInbox` and `listChatRequests` to reduce clutter from chats that were never sent.

lilseyi:
@codex please re-review — added one more change in 0a33187: split admin Top Channels into a separate Top Channels (group channels) + Top Direct Messages (DMs + group_dms) card so DM activity stops drowning out group activity in the dashboard.

lilseyi:
@codex please re-review — addressed all three in 6024ddc:
- **P1** (`getDirectInbox`): never-sent vs sent-then-deleted now distinguished by peeking at `chatMessages` (any row, including soft-deleted, keeps the channel visible).
- **P2** (`listChatRequests`): same distinction — pending requests with deleted-original-message stay visible.
- **P2** (legacy `topChannels`): restored as the combined list; added `topGroupChannels` for the new split. Frontend prefers `topGroupChannels` with `topChannels` fallback.

lilseyi:
@codex please re-review — addressed in 46ad822: legacy `topChannels` now reads the global top 20 by engagement score (`channelScores.slice(0, 20)`), not group-then-dm concatenation. Older clients see the correct combined ranking.

PR #359, opened 2026-04-28
fix(chat): DMs get the same thread reply UX as group channels

lilseyi:
## Summary
Replies in DMs were broken three ways. Group-channel parity:

- **Empty reply preview**: `ConvexChatRoomScreen.tsx` passed a hardcoded `{ _id, content: "", senderName: "" }` to `MessageInput`, so the "Replying to …" banner rendered with no name and no quoted text. Now loads the parent via `useParentMessage(replyToMessageId)` and passes real values.
- **Replies vanished**: `getMessages` filters out any message with a `parentMessageId` — they only render inside a thread page. DMs had no thread page, so replies were unreachable.
- **Thread nav was groupId-gated**: `MessageItem.handleThreadPress` only ran `if (groupId)` and `<ThreadReplies onPress={...} />` was wired to `undefined` for DMs. The "1 reply" indicator was non-tappable.

Fix:
- Add `/inbox/dm/[channelId]/thread/[messageId]` (mirrors the group route).
- Generalize `ThreadPage` to accept either `groupId` or `dmChannelId` — DMs skip the leader/admin check (no role concept) and back-nav returns to `/inbox/dm/<channelId>`.
- `MessageItem.handleThreadPress` picks the route by context (group vs DM) instead of refusing to navigate.

## Test plan
- [x] Mobile `tsc --noEmit` — no new errors in changed files (only pre-existing errors elsewhere remain)
- [ ] Manual: in a DM, long-press a message → Reply → composer shows "Replying to <name>: <preview>" (was blank)
- [ ] Manual: send the reply → it disappears from the main DM list (expected — thread reply), parent shows "1 reply"
- [ ] Manual: tap "1 reply" on the parent → navigates to `/inbox/dm/<channelId>/thread/<messageId>` and shows the parent + replies
- [ ] Manual: send another reply from the thread page → appears inline; back arrow returns to the DM
- [ ] Manual: same flow in a group channel → unchanged behavior, still routes to `/inbox/<groupId>/thread/<messageId>`

### DIFF
--- a/apps/mobile/app/inbox/dm/[channelId]/_layout.tsx
+++ b/apps/mobile/app/inbox/dm/[channelId]/_layout.tsx
@@ -18,6 +18,7 @@ export default function DmChannelLayout() {
     >
       <Stack.Screen name="index" options={{ animation: "none" }} />
       <Stack.Screen name="info" options={{ animation: "slide_from_right" }} />
+      <Stack.Screen name="thread" options={{ animation: "slide_from_right" }} />
     </Stack>
   );
 }
--- a/apps/mobile/app/inbox/dm/[channelId]/thread/[messageId].tsx
+++ b/apps/mobile/app/inbox/dm/[channelId]/thread/[messageId].tsx
@@ -1,0 +1,34 @@
+/**
+ * DM Thread Page Route
+ *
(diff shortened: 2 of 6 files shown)


lilseyi:
@codex please review this PR

PR #360, opened 2026-04-28
fix(chat): drop 1-message-per-24h cap on pending DMs

lilseyi:
## Summary
Ad-hoc DM/group_dm `sendMessage` enforced a 1-message-per-pending-pair-per-24h cap as stranger-spam defense. Users hit it on their second message before the recipient had time to accept, and the throw was `new Error(...)` so Convex sanitized it to "Server Error" / "failed" with no signal that the cap was the problem.

Drop the cap entirely. Ad-hoc channels are already scoped to shared community membership — sender and recipient aren't strangers — so the original concern doesn't really apply. Also stop writing to \`directMessageRateLimits\` (table left in schema as orphan; follow-up cleanup PR can drop it once we confirm no other readers).

Defense-in-depth kept while a recipient is still pending:
- Attachments still rejected.
- Per-message text length still capped at 1000 chars.
- All these throws (plus "Accept the request before replying" and the block-recipient throw) now use \`ConvexError\` so the actual message reaches the client.

## Test plan
- [x] \`__tests__/messaging/directMessages.test.ts\`: replaced the rate-limit test with a positive "inviter can send multiple messages while pending" test — 33/33 pass.
- [x] Full messaging suite: 405/405 pass.
- [x] Convex \`tsc --noEmit\` clean.
- [ ] Manual: in a fresh DM, send first message → succeeds; send second message immediately → also succeeds (was "failed" before).
- [ ] Manual: try to attach an image while recipient still pending → clear "Cannot send attachments…" error reaches the UI (was sanitized).
- [ ] Manual: type >1000 chars while pending → clear length error reaches the UI.

## Follow-up
The \`directMessageRateLimits\` table is now unused. Leaving in schema for one release to be safe; can be dropped in a follow-up PR.

### DIFF
--- a/apps/convex/__tests__/messaging/directMessages.test.ts
+++ b/apps/convex/__tests__/messaging/directMessages.test.ts
@@ -408,7 +408,13 @@ describe("respondToChatRequest", () => {
 // ============================================================================
 
 describe("sendMessage gating on pending channels", () => {
-  test("rate limit: sender can only send 1 message to a pending recipient per 24h", async () => {
+  test("inviter can send multiple messages while a recipient is still pending", async () => {
+    // Regression: a 1-message-per-24h cap used to fire here as a stranger-spam
+    // defense, but ad-hoc channels are scoped to shared community membership
+    // (recipient and sender aren't strangers) and the cap surfaced as an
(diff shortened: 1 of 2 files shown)


lilseyi:
@codex please review this PR

PR #362, opened 2026-04-29
fix(chat): clearer error on Accept (Sentry — Server Error masking)

lilseyi:
> Replaces #361 — that PR accidentally committed 27 unrelated files. Please close #361. This is the clean one-file version.

## Summary

A recipient tapping **Accept** on a chat request was getting an opaque \"[CONVEX M(...directMessages:respondToChatRequest)] [Request ID: ec9b7b6dcf7bc863] Server Error\" alert (screenshot in the bug report).

## Root cause

\`requireAuth\` throws \`new Error(\"Not authenticated\")\` for an expired/revoked token. Convex hides plain-Error messages as \"Server Error\" in prod (only \`ConvexError\` messages surface). The recipient's auth token most likely expired between page load (\`listChatRequests\` succeeded) and the Accept tap.

## Changes

- Wrap \`requireAuth\` in \`respondToChatRequest\` to re-throw as \`ConvexError\`:
  > Your sign-in has expired. Pull down to refresh, then try again.
- Make Accept idempotent — if the caller's membership is already \`accepted\` and not left, return \`{ ok: true }\` instead of throwing \"not pending response\". Covers double-taps / stale UI.

## Test plan

- [x] Convex \`tsc --noEmit\`: clean.
- [x] Convex unit: 33/33 pass in \`directMessages.test.ts\`.
- [ ] Prod smoke: recipient with expired token taps Accept → sees \"Your sign-in has expired\" instead of \"Server Error\".

### DIFF
--- a/apps/convex/functions/messaging/directMessages.ts
+++ b/apps/convex/functions/messaging/directMessages.ts
@@ -490,13 +490,26 @@ export const respondToChatRequest = mutation({
     // "Server Error" (which `throw new Error` produces in production mode).
     // See Sentry — recipients tapping Accept were getting an opaque "Server
     // Error" alert with no signal that they needed to re-authenticate.
+    //
+    // Catch is NARROWED to actual auth failures only. `requireAuth` performs
+    // DB lookups internally; a transient DB/runtime error from those would
+    // otherwise be misdiagnosed as a session-expiry issue, masking real
+    // server faults during incidents. Only the documented auth-failure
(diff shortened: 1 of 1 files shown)


lilseyi:
@codex please review this PR

lilseyi:
@codex please re-review — addressed in 918f178: catch narrowed to only the literal "Not authenticated" error; all other failures inside `requireAuth` (DB / runtime) re-throw unchanged.

PR #374, opened 2026-04-30
fix(events): inbox row alignment + address truncation + comment reactions

lilseyi:
## Summary

Three independent bugs you reported on event surfaces, bundled in one PR.

### 1. Inbox event row — avatar misalignment

The 3px accent bar + 10px marginRight pushed the event avatar 13px right of where DM/group avatars sit, so event rows read as inset relative to the neighboring rows. **Fix**: pin the accent bar absolutely to the row's left edge and restore the row's paddingLeft to 16 (matches \`dmRow\` / \`groupItem\`). Avatars now line up vertically across the mixed inbox.

### 2. Inbox event row — address truncation

The DirectionsButton's location text was overflowing past the row's right edge instead of truncating with ellipsis. **Root cause**: \`flexShrink: 1\` was set on the inner View, but the actual flex item in the row is the outer \`Pressable\`, which had no style at all. **Fix**: \`flexShrink: 1, minWidth: 0\` on the Pressable wrapper so the inner \`numberOfLines={1}\` text truncates correctly.

### 3. Event comments — no way to add a reaction

\`EventComment\` had no gesture or button to open a reaction picker. Users could tap an existing reaction pill to toggle their vote, but adding a brand-new emoji had no entry point. **Fix**: inline emoji picker with three discoverable openers:
- "+ smiley" pill at the end of the existing reactions row
- "React" link in the action row next to "Reply" (only when no reactions exist)
- Long-press anywhere on the comment body (matches the chat \`MessageItem\` gesture)

The picker renders inline (a horizontal ScrollView below the comment) instead of as a Modal because \`EventActivity\` is embedded in the event page's outer ScrollView and modal overlays render behind it on native (same root cause as the prior thread-page-behind-event-page bug).

The picker re-uses the \`REACTIONS\` constant exported from \`MessageActionsOverlay\` so the chat + event surfaces share the same emoji set.

## Verification

- 115 / 115 chat + event tests pass
- Mobile typecheck clean for changed files

## Test plan

- [ ] Inbox: event row avatar lines up vertically with DM and group rows above/below it
- [ ] Inbox: a long event address truncates with ellipsis instead of overflowing past the right edge
- [ ] Event activity: tap an existing reaction pill — your vote toggles
- [ ] Event activity: tap the "+ smiley" pill at the end of an existing reactions row — picker opens, tap an emoji, reaction added
- [ ] Event activity: on a comment with no reactions, tap "React" next to "Reply" — picker opens
- [ ] Event activity: long-press anywhere on a comment body — picker opens
- [ ] Tapping an emoji closes the picker and adds the reaction

### DIFF
--- a/apps/mobile/app/e/[shortId]/EventComment.tsx
+++ b/apps/mobile/app/e/[shortId]/EventComment.tsx
@@ -22,12 +22,14 @@ import {
   StyleSheet,
   Pressable,
   Linking,
+  ScrollView,
 } from 'react-native';
 import { Ionicons } from '@expo/vector-icons';
 import { useRouter } from 'expo-router';
 import type { Id } from '@services/api/convex';
 import { AppImage, ImageViewer } from '@components/ui';
 import { useReactions, type Reaction } from '@features/chat/hooks/useReactions';
+import { REACTIONS } from '@features/chat/components/MessageActionsOverlay';
 import { parseMessageContent } from '@features/shared/utils/linkify';
 import { getMediaUrl } from '@/utils/media';
 import { useTheme } from '@hooks/useTheme';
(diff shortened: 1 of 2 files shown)


lilseyi:
@codex please review this PR

PR #375, opened 2026-05-03
fix(chat): refresh sender avatar/name from live user record on read

lilseyi:
## Summary

Profile pictures in chat go stale after a user updates their photo because messages denormalize `senderName` + `senderProfilePhoto` at send time and that snapshot is never refreshed on historical rows.

This PR overrides the snapshot at read time by batch-fetching the sender's `users` record in `getMessage`, `getMessages`, and `getThreadReplies`. The denormalized fields are still written on send and used as a fallback for system/bot messages and deleted users, so historical context isn't lost.

- `applyLiveSenderProfile(message, user)` — single-message override
- `attachLiveSenderProfile(ctx, messages)` — one `users` lookup per unique `senderId` per page

DM inbox + channel list previews already pull `lastMessageSenderName` and `chatChannelMembers.profilePhoto`, both of which `syncUserProfileToChannels` already keeps fresh — those are unchanged.

## Test plan

- [x] New unit tests under "Live Sender Profile Override" assert that `getMessage`, `getMessages`, and `getThreadReplies` return the user's current avatar after a profile update, not the snapshot
- [x] Full convex suite: 1412/1412 passing across 80 files
- [ ] Manual: change profile photo in dev, scroll back through an old DM thread, confirm previously-sent messages now show the new avatar

### DIFF
--- a/apps/convex/__tests__/messaging/messages.test.ts
+++ b/apps/convex/__tests__/messaging/messages.test.ts
@@ -1403,6 +1403,91 @@ describe("Get Single Message", () => {
 });
 
 // ============================================================================
+// Live sender profile (always-fresh avatar/name)
+// ============================================================================
+
+describe("Live Sender Profile Override", () => {
+  test("getMessage returns the sender's current profile photo, not the snapshot", async () => {
+    vi.useFakeTimers();
+    const t = convexTest(schema, modules);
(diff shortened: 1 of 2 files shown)


lilseyi:
@codex please review this PR

lilseyi:
@codex please review this PR

PR #376, opened 2026-05-04
feat(landing-page): auto-reply SMS composed from automation rule snippets

lilseyi:
## Summary

Lets community admins configure a single SMS auto-reply that fires when a visitor submits the landing-page form. The body is built from a fixed intro + per-rule snippets (one bullet per matching automation rule with the new "Append to auto-reply SMS" action) + a fixed outro, with `{firstName}` substitution.

Use case from the field: Brittany at Fount wants visitors who check "Dinner Party + Foundations + Baptisms" to get **one** text covering all three selections, not three separate texts. Today there's no way to send any auto-reply at all.

## What changed

**Schema** (`apps/convex/schema.ts`)
- `automationRules.action` gets optional `snippet` (for the new `append_sms` action type)
- New optional `autoReplySms` config on `communityLandingPages` — `{ enabled, intro, outro, sendIfNoSnippetsMatch }`

**Backend** (`apps/convex/functions/communityLandingPage.ts`, `communityLandingPageActions.ts`)
- Rule evaluator collects every matching `append_sms` snippet while preserving first-match-wins for `set_assignee`
- `saveConfig` validates snippet presence and an 800-char intro+outro cap
- New `tryConsumeSmsRateLimit` — 2 sends per phone per 24h, returns boolean instead of throwing
- New `recordSmsAuditOnNote` — appends "Auto-Reply SMS sent: <body>" or "suppressed: ..." to the existing landing-page submission note
- `submitForm` action composes the SMS, enforces the per-phone cap, dispatches via the existing `internal.functions.auth.phoneOtp.sendSMS` action through the scheduler so a Twilio outage cannot fail the form submission

**Admin UI** (`apps/mobile/features/admin/components/LandingPageContent.tsx`)
- New "Auto-Reply SMS" card above Automation Rules with intro/outro/enabled/sendIfNoSnippetsMatch
- Rule editor modal gets an action-type picker; snippet textarea swaps in when "Append to auto-reply SMS" is chosen
- Rule list summary shows snippet preview

## Safety nets

- Form-submission rate limit (5/h per phone, existing) gates the source
- New SMS-specific cap (2/24h per phone) stops repeat submissions from spamming the same recipient with auto-replies
- Twilio's 1600-char body limit enforced at compose time
- SMS dispatched via scheduler — Twilio errors never fail the form
- Test phones short-circuit in `sendSMS` (existing infra)
- Existing schema field stays optional → no migration needed

## Test plan

- [x] Convex + mobile typecheck pass
- [x] Browser E2E (Playwright):
  - [x] Admin UI: enable Auto-Reply SMS, edit intro/outro, save
  - [x] Add `append_sms` rule via modal (action-type picker, snippet textarea)
  - [x] Submit form with one matching rule → SMS dispatched with composed body
  - [x] Submit form with two matching rules (Dinner Party + Foundations) → both bullets in one composed SMS
  - [x] Submission audit appended to follow-up note (verified `memberFollowups.content` in DB)
  - [x] 3rd submission within 24h → rate-limit cap fires, "suppressed" audit recorded, form submission still succeeds
- [ ] Reviewer to sanity-check Twilio path on staging with a non-test phone

## Follow-ups (out of scope)

UX-review subagent flagged these — worth a separate PR:
- Live SMS preview in the auto-reply card
- Cross-section warnings (SMS enabled with no append rules; rules exist with SMS disabled)
- "Insert {firstName}" affordance + plain-language hint
- Friendlier action-type label ("Add a line to the auto-reply text")

### DIFF
--- a/apps/convex/_generated/api.d.ts
+++ b/apps/convex/_generated/api.d.ts
@@ -156,6 +156,7 @@ import type * as lib_membership from "../lib/membership.js";
 import type * as lib_notifications_definitions from "../lib/notifications/definitions.js";
 import type * as lib_notifications_emailTemplates from "../lib/notifications/emailTemplates.js";
 import type * as lib_notifications_enabledCounter from "../lib/notifications/enabledCounter.js";
+import type * as lib_notifications_enabledStatus from "../lib/notifications/enabledStatus.js";
 import type * as lib_notifications_index from "../lib/notifications/index.js";
 import type * as lib_notifications_registry from "../lib/notifications/registry.js";
 import type * as lib_notifications_send from "../lib/notifications/send.js";
@@ -330,6 +331,7 @@ declare const fullApi: ApiFromModules<{
   "lib/notifications/definitions": typeof lib_notifications_definitions;
   "lib/notifications/emailTemplates": typeof lib_notifications_emailTemplates;
   "lib/notifications/enabledCounter": typeof lib_notifications_enabledCounter;
+  "lib/notifications/enabledStatus": typeof lib_notifications_enabledStatus;
   "lib/notifications/index": typeof lib_notifications_index;
   "lib/notifications/registry": typeof lib_notifications_registry;
   "lib/notifications/send": typeof lib_notifications_send;
--- a/apps/convex/functions/communityLandingPage.ts
+++ b/apps/convex/functions/communityLandingPage.ts
@@ -18,7 +18,7 @@ import { VALID_CUSTOM_SLOTS } from "../lib/followupConstants";
 import { normalizePhone, buildSearchText, now } from "../lib/utils";
 import { syncUserChannelMembershipsLogic } from "./sync/memberships";
 import { ensureChannelsForGroupLogic } from "./messaging/channels";
-import { checkRateLimit } from "../lib/rateLimit";
+import { checkRateLimit, RateLimitExceededError } from "../lib/rateLimit";
 import { parseDateOptional } from "../lib/validation";
 
 // ============================================================================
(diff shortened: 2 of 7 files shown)


lilseyi:
@codex please review this PR

lilseyi:
Both codex findings addressed in 6ecb863:

- **P1** (line 142, action aborts on SMS failure): wrapped `maybeSendAutoReplySms` in try/catch in `submitForm` so any internal-mutation or scheduler error logs and is swallowed — the form submission can no longer fail after user/membership records are written.
- **P2** (quota burned before enqueue succeeds): combined the rate-limit check and schedule into a single internal mutation `dispatchAutoReplySmsIfAllowed`, so quota consumption and the schedule write commit atomically. The quota can no longer be burned without an SMS being scheduled.

Bonus: the submission note now also reflects when Twilio itself fails. A new `sendAutoReplySmsAndAudit` internal action calls Twilio inside the scheduled work and writes either `Auto-Reply SMS sent: ...` or `Auto-Reply SMS FAILED to send (<error>). The recipient did not receive this message: ...` to the audit block, preserving the body so staff know what to follow up about.

lilseyi:
@codex please re-review after 6ecb863

lilseyi:
Round 2 codex findings addressed in 55b12a8:

- **Surrogate pair truncation** (`communityLandingPageActions.ts:205`): added `truncateSmsBody()` that detects a high surrogate at the slice boundary and trims one char so we never ship a lone surrogate.
- **Bare catch in dispatch** (`communityLandingPage.ts:163`): introduced `RateLimitExceededError` in `lib/rateLimit.ts` and narrowed the catch in `dispatchAutoReplySmsIfAllowed` so only true cap-reached signals translate to `suppressed_cap`. Unexpected errors propagate to the action's outer try/catch and get logged as operational failures.

lilseyi:
@codex please re-review after 55b12a8

PR #377, opened 2026-05-04
fix(landing-page): keyboard avoidance + visible validation in admin modals

lilseyi:
## Summary

Fixes two reported issues in the landing-page admin field/rule editor modals:

1. **iOS keyboard blocked the Add/Update button** — when an input was focused the save row was hidden behind the keyboard with no way to scroll past it.
2. **Web "Add Rule" appeared to do nothing on click** — the click was actually firing, but \`Alert.alert(\"Error\", \"Rule name is required\")\` resolves to \`window.alert()\` on web, which some browsers/extensions silently suppress. The user saw no validation feedback and assumed the button was broken.

Same modal pattern is shared by FieldEditorModal and RuleEditorModal so both got the same treatment.

## What changed

- **\`KeyboardAvoidingView\`** wraps each modal's overlay with \`behavior=\"padding\"\` on iOS so the save row lifts above the keyboard.
- **Inline error banner** above the save button replaces \`Alert.alert\` validation calls — always visible, no browser-dependent dialog.
- **Modal container** gets explicit \`display: flex; flexDirection: column\` and \`ScrollView\` gets \`flexShrink: 1\` so the save button stays anchored at the bottom of the 85% maxHeight on every viewport.
- **\`keyboardShouldPersistTaps=\"handled\"\`** on the ScrollView so chip taps register without dismissing the keyboard first.

## Test plan

- [x] Convex + mobile typecheck pass
- [x] Web E2E (Playwright):
  - [x] Open admin landing → Add automation rule → click "Add Rule" with empty form → inline "Rule name is required" appears (was silent before)
  - [x] Fill name + select condition field + select \"Append to auto-reply SMS\" + add snippet → click "Add Rule" → modal closes, rule appears in list, Save Changes button surfaces
- [ ] Reviewer to spot-check on iOS Simulator: focus an input in either modal, verify the save button stays reachable above the keyboard

### DIFF
--- a/apps/mobile/features/admin/components/LandingPageContent.tsx
+++ b/apps/mobile/features/admin/components/LandingPageContent.tsx
@@ -19,6 +19,8 @@ import {
   Alert,
   ActivityIndicator,
   Modal,
+  KeyboardAvoidingView,
+  Platform,
 } from "react-native";
 import { Ionicons } from "@expo/vector-icons";
 import { useTheme } from "@hooks/useTheme";
@@ -810,6 +812,7 @@ function FieldEditorModal({
   const [options, setOptions] = useState("");
   const [placeholder, setPlaceholder] = useState("");
   const [buttonUrl, setButtonUrl] = useState("");
+  const [validationError, setValidationError] = useState<string | null>(null);
 
   useEffect(() => {
     if (visible) {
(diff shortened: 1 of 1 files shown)


lilseyi:
@codex please review this PR

PR #378, opened 2026-05-08
fix(chat): stop chat-path queries/mutations from throwing on lost membership

lilseyi:
## Summary

When a user loses access to a channel mid-session — confirms a Leave Group alert, gets demoted from leader, has a channel disabled by an admin from another session — the still-mounted chat screen's reactive queries and auto-fired mutations re-execute on the now-stale viewer. Several of them threw `new Error(...)`, which crashed the screen via ErrorBoundary (queries) or surfaced as unhandled rejections (mutations from `useEffect` / keystroke handlers without `.catch`).

This PR sweeps the chat-screen backend surface and converts the membership/access throws to the same return-empty pattern the rest of the file already uses.

| Function | File | Before | After |
| -------- | ---- | ------ | ----- |
| `getMessages` | `messages.ts` | throw on lost membership / disabled leader channel; throw `Error("Channel not found")` | return `{ messages: [], hasMore: false, cursor: undefined }` for membership/disabled cases; `ConvexError` for channel-not-found |
| `getMessageReadBy` | `readState.ts` | throw on lost membership / deleted message; throw `Error` on channel mismatch | return `{ readByCount: 0, totalMembers: 0 }` for the first two; `ConvexError` for channel mismatch |
| `markAsRead` | `readState.ts` | throw on lost membership | silent no-op (matches the existing ad-hoc-channel no-op a few lines below) |
| `startTyping` | `typing.ts` | throw on lost membership | silent no-op |

## Why these specifically

Identified from a Sentry investigation of user `m97607fsnnzp5cwcknqdmhw7mx860f9t` (eventID `bdc338e1b5ec48dda1953d676af2354d`, 2026-05-07T21:00:14Z). Breadcrumbs:

```
21:00:13.559Z  RCTAlertController presented — title: "Leave Group"
21:00:14.590Z  ErrorBoundary caught getMessages Server Error
```

`getMessages` was the only function that had actually crashed the UI tree in production — aggregating 30d of Sentry events by groupID:

| Sentry issue | Events | Latest      |
| ------------ | ------ | ----------- |
| 7347937366   | 15     | 2026-04-14  |
| 7393363858   | 21     | 2026-04-09  |
| 7430608070   | 21     | 2026-04-30  |
| 7466431624   | 21     | 2026-05-07  |

Each cluster is a single user; the ~20 events per incident are Convex sync-worker retries of the same failing query, not 20 different users. All four currently share the generic `"Server Error"` title because plain `Error` (vs `ConvexError`) is sanitized before the message reaches the client.

`getMessageReadBy`, `markAsRead`, and `startTyping` had the same shape and would fire from the same chat screen during the same Leave-Group transition; they aren't currently a top issue in Sentry but they're firing as unhandled rejections during the ~1s gap between membership loss and screen unmount. `ConvexChatRoomScreen.tsx:434` is the smoking gun — `useEffect` calls `markAsRead()` without a `.catch`.

## Why empty-return is the right shape

The same files already use it elsewhere:

- `messages.ts:298-301` (event channels) — with a comment explaining the exact reasoning: *"callers are React useQuery hooks, and a hard throw crashes the UI tree via the error boundary."*
- `messages.ts:222-266` (`getMessage` singular) — `try`/`catch` returns `null` for any access failure
- `readState.ts:36-38` (`getUnreadCount`) — returns `0`
- `channels.ts:521-540` (`getChannelsByGroup`), `656-716` (`getChannelMembers`), `910-955` (`listGroupChannels`), `1594-1613` (`hasAutoChannels`)
- `typing.ts:42-44` (`getTypingUsers`)
- `blocking.ts` (all) — no membership checks needed

## Audit notes

These chat-path queries/mutations were checked and were already safe:

- `messaging/readState:getUnreadCount`, `getUnreadCounts`
- `messaging/channels:getChannelsByGroup`, `getUserChannels`, `getChannelMembers`, `listGroupChannels`, `getInboxChannels`, `hasAutoChannels`
- `messaging/typing:getTypingUsers`, `stopTyping`
- `messaging/blocking:getBlockedUsers`, `isBlocked`, `isBlockedBy`

DM-inbox functions (`directMessages:listChatRequests`, `getDirectInbox`) were not audited in this PR — different surface, different lost-state semantics.

## Follow-ups (not in this PR)

- Enable a Convex Log Stream → Sentry sink in the dashboard so future incidents surface their actual server-side error message in Sentry without log archeology. Convex MCP retention is too short to retrieve 4-6h-old request IDs (this is what made today's investigation harder than it needed to be).

## Test plan

- [x] `pnpm -w exec tsc -p apps/convex/tsconfig.json --noEmit` clean
- [x] `pnpm test` in `apps/convex` — 80 files / 1412 tests pass
- [x] Updated three existing tests that asserted the old throw-on-lost-membership behavior:
  - `readState.test.ts`: "should throw error if user is not a member of the channel" → "returns zero counts if user is not a member of the channel"
  - `readState.test.ts`: "should throw error if message does not exist" → "returns zero counts if message does not exist"
  - `typing.test.ts`: "should reject from non-channel-member" → "silently no-ops for a non-channel-member" (also asserts no indicator row was written)
- [ ] Manual: open a group chat as a non-leader member → confirm Leave Group → expect smooth navigation away with no ErrorBoundary screen and no Sentry events from `getMessages`, `getMessageReadBy`, `markAsRead`, or `startTyping` during the transition
- [ ] Manual: open a group chat → admin disables the channel from another session → expect: chat shows empty (no crash), screen remains responsive
- [ ] Watch Sentry for ~1w after merge — the four recurring `getMessages: Server Error` clusters should taper off, and any remaining occurrences should be true `Channel not found` cases (now distinguishable by message via `ConvexError`)

### DIFF
--- a/apps/convex/__tests__/messaging/readState.test.ts
+++ b/apps/convex/__tests__/messaging/readState.test.ts
@@ -745,7 +745,9 @@ describe("Get Message Read By", () => {
     expect(result.totalMembers).toBe(2);
   });
 
-  test("should throw error if user is not a member of the channel", async () => {
+  test("returns zero counts if user is not a member of the channel", async () => {
+    // Reactive query: must not throw on lost membership or it crashes the
+    // chat screen via ErrorBoundary.
     const t = convexTest(schema, modules);
     const { communityId, channelId } = await seedTestData(t);
 
@@ -775,32 +777,30 @@ describe("Get Message Read By", () => {
       });
     });
 
-    await expect(
-      t.query(api.functions.messaging.readState.getMessageReadBy, {
-        token: unauthorizedToken,
-        messageId,
-        channelId,
-      })
-    ).rejects.toThrow("Not a member of this channel");
+    const result = await t.query(api.functions.messaging.readState.getMessageReadBy, {
+      token: unauthorizedToken,
+      messageId,
+      channelId,
+    });
+    expect(result).toEqual({ readByCount: 0, totalMembers: 0 });
   });
 
-  test("should throw error if message does not exist", async () => {
+  test("returns zero counts if message does not exist", async () => {
+    // Reactive query: stale message id from a deleted message must not crash.
     const t = convexTest(schema, modules);
     const { userId, channelId, accessToken } = await seedTestData(t);
 
-    // Create a message and then delete it to get a valid but non-existent ID
     const messageId = await sendMessage(t, channelId, userId, "Test");
     await t.run(async (ctx) => {
       await ctx.db.delete(messageId);
     });
 
-    await expect(
-      t.query(api.functions.messaging.readState.getMessageReadBy, {
-        token: accessToken,
-        messageId,
-        channelId,
-      })
-    ).rejects.toThrow("Message not found");
+    const result = await t.query(api.functions.messaging.readState.getMessageReadBy, {
+      token: accessToken,
+      messageId,
+      channelId,
+    });
+    expect(result).toEqual({ readByCount: 0, totalMembers: 0 });
   });
 
   test("should throw error if message does not belong to channel", async () => {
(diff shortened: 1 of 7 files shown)


lilseyi:
@codex addressed in e8ff9d8 — also clearing the persisted cache. Please confirm.

PR #384, opened 2026-05-08
fix(groupMembers): clear request state when reactivating existing rows

lilseyi:
## Summary

Follow-up to #383 addressing Codex feedback.

When `groupMembers.add` (and the new `addByPcoPersonId`) reactivates an existing inactive row by clearing `leftAt`, the row could have leftover `requestStatus` ("pending" / "declined") + `requestedAt` from a prior join request. The reactivation branch was clearing `leftAt` but leaving those fields, so the row would be active per the member-list filter (no `requestStatus` or `accepted`) yet still surface in `listMyPendingJoinRequests`. A leader using "Add to group" (or "Add member") bypasses the request flow entirely, so the row must look like a clean accepted membership.

The bug applies to both `add` and `addByPcoPersonId` (same reactivation pattern in both); fixing both here.

## Test plan

- [ ] Manual: previously-declined member is added via leader-tools "Add member" → no longer shows up in their pending requests list.
- [ ] Manual: PCO unmatched person with a prior pending request is added via "Add to group" quick action → membership is active, request no longer appears as pending.
- [x] Convex `tsc --noEmit` clean.


lilseyi:
@codex please review this PR

PR #386, opened 2026-05-09
feat(chat,polls): show voters + RSVP-style avatar stack

lilseyi:
Stacked on #385.

## Summary

- Tapping a poll's footer (or the new RSVP-style avatar stack on each option) opens a sectioned sheet listing voters by option, with avatar + full display name per row.
- Each option row gets an inline overlapping avatar stack of the first 4 voters, plus a "+N" overflow chip when there are more.
- Backed by a new \`getPollVoters\` query and an extended \`getPoll\` that returns \`voterPreview\` per option.
- Future-proofed for anonymous polls: queries gate identities on \`canSeeIdentities\` (always true in v1; false only for non-leaders viewing an \`isAnonymous\` poll).

## Architecture

- \`apps/convex/functions/messaging/polls.ts\`: \`getPoll\` now batches user lookups for the top-N voters per option and returns \`voterPreview: [{ userId, displayName, profilePhoto }]\`. New \`getPollVoters\` query returns the full per-option voter list (display name + photo + createdAt) for the modal.
- \`apps/mobile/features/chat/components/PollVotersSheet.tsx\`: SectionList modal — section header is option text + count, rows are avatar + name. Voter rows are sorted earliest-first per option.
- \`apps/mobile/features/chat/components/PollCard.tsx\`: new inline avatar stack at the right of each option row (size 20, -8px overlap, surface-color border ring) and a tappable footer with chevron. Both surfaces open the same voters sheet.

## Test plan

- [x] Convex typecheck clean (no new errors)
- [x] Mobile typecheck clean (no new errors)
- [x] Manual sim: cast a vote — avatar appears inline on the option row, count updates
- [x] Manual sim: cast a second vote on the same option — two avatars stacked
- [x] Manual sim: footer reads "N voter · M vote" with chevron when there are votes
- [ ] Manual sim: tap footer → voters sheet (verified visible UI; tap couldn't be exercised through the simulator MCP because the outer MessageItem Pressable layer captures the tap region — works on real device)

### DIFF
--- a/apps/convex/functions/messaging/polls.ts
+++ b/apps/convex/functions/messaging/polls.ts
@@ -817,6 +817,19 @@ export const getPoll = query({
     const canModerate = isAuthor || isLeader;
     const canSeeIdentities = !poll.isAnonymous || isAuthor || isLeader;
 
+    // Counts always include every vote (so a blocker doesn't see a
+    // skewed total). Voter identity surfaces — preview avatars and the
+    // voters sheet — strip blocked users so the blocker doesn't see the
+    // person they explicitly hid from their feed reappear here. This
+    // mirrors how `getMessages` filters blocked senders' messages.
+    const blockedRows = await ctx.db
+      .query("chatUserBlocks")
+      .withIndex("by_blocker", (q) => q.eq("blockerId", userId))
+      .collect();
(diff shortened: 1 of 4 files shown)


lilseyi:
@codex please re-review the latest commit

PR #410, opened 2026-05-23
AssignSheet: placeholders re-assignable + show phone for disambiguation

lilseyi:
## Summary

Two small UX fixes on the AssignSheet after testing on prod:

1. **Placeholders can be assigned to multiple roles.** Previously the picker greyed them out after the first assignment and the backend's `isActive === false` guard rejected re-assignment via `assignFromCommunity`. Now they behave exactly like real volunteers — assignable to as many roles as the leader needs.
2. **Phone numbers display under every name.** Disambiguates name collisions (two "Test User"s, etc.).

### Frontend (`ea9a840`)
`AssignSheet` was force-routing every `isPlaceholder` row into the COMMUNITY section AND disabling the row. Two problems:
- Placeholders *are* group members (`inviteAndAssign` inserts the row). They should appear in **GROUP** based on their actual `inGroup` flag.
- Disabling the row blocks the leader from putting the same person on a second role on the same event. The existing `assignedUserIds` per-role guard already covers "already in this role" — the placeholder-wide disable was over-eager.

Fix: partition by real `inGroup`, drop the placeholder disable. Keep the small "Invited" badge inline next to the name (still useful context, no greyed-out look).

### Backend (`4d9b8e7`)
`assignFromCommunity`'s `target.isActive === false` guard was originally intended to block deactivated accounts. Placeholder users are intentionally `isActive: false` until they sign up and claim — but they're valid scheduling targets. Tightened the predicate:

```ts
if (target.isActive === false && target.isPlaceholder !== true) {
  throw new ConvexError("This person's account is deactivated and cannot be scheduled");
}
```

Defense-in-depth — the typical path is `assignRole` (placeholder appears in GROUP), but the COMMUNITY-section path also works now for the rare case where a placeholder ended up there.

## Bug 2 — show phone numbers everywhere

`searchCommunityPeople` already returns `phone` in E.164. The AssignSheet now renders it under each name in tertiary color at `fontSize: 12`. New inline helper `formatPhoneForDisplay`: `+1XXXXXXXXXX` → `(XXX) XXX-XXXX`; falls back to the raw string for non-US numbers. Missing phone → no second line.

## Tests

- `assignFromCommunity` permits a placeholder user (`addedToGroup: false`, assignment created).
- `assignFromCommunity` still rejects a deactivated non-placeholder.
- 26 assignments tests pass; full scheduling suite green.

## Verification

- `apps/convex` typechecks clean.
- `apps/mobile` typechecks clean (13 pre-existing unrelated errors).
- 1030 mobile tests pass.

### DIFF
--- a/apps/convex/__tests__/scheduling/assignments.test.ts
+++ b/apps/convex/__tests__/scheduling/assignments.test.ts
@@ -755,6 +755,53 @@ describe("assignFromCommunity", () => {
       }),
     ).rejects.toThrow(/already assigned/);
   });
+
+  it("permits a placeholder user (isActive: false + isPlaceholder: true)", async () => {
+    const { t, world } = await setupSchedulingWorld();
+    const leaderToken = (await generateTokens(world.groupLeaderId)).accessToken;
+    const planId = await makeEvent(t, world, leaderToken, 7);
+
+    // The fixture's `placeholderUserId` is `isActive: false, isPlaceholder: true`
+    // with active community + group memberships — exactly what a placeholder
+    // looks like after `inviteAndAssign` creates it. They must remain
+    // schedulable so a leader can put them on multiple roles, just like a
+    // real volunteer.
+    const result = await t.mutation(
+      api.functions.scheduling.assignments.assignFromCommunity,
+      {
+        token: leaderToken,
+        planId,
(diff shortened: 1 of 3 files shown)


lilseyi:
@codex review

PR #539, opened 2026-07-02
feat(channels): surface cross-team channels in group list + reachable sync settings

lilseyi:
## Problem

Cross-team chat channels were invisible in two places:

1. A group's **CHANNELS** settings list — and their role-sync editor was unreachable from that surface.
2. The **event-mode (serving) Inbox** tab — only serving-team channels showed up.

**Root cause (1 — group settings):** `ChannelsSection.tsx` builds its rows from a `channelType` **allowlist** — it branches on `main` / `leaders` / `reach_out` / `announcements`, then appends `pco_services` and `custom` rows — but has **no branch for `cross_team`**. The backend `messaging.channels.listGroupChannels` already returns cross-team channels; they were simply dropped on the frontend, so a leader could never tap through to Channel Info → **Edit synced roles**.

**Root cause (2 — serving inbox):** `getInboxChannels` (with a `servingPlanId`) intersects the inbox against the Set built by `resolveServingChannelIds` in `scheduling/serving.ts`. That resolver only added (a) team channels and (b) linked meeting channels — it had **no `cross_team` branch**, so cross-team channels were filtered out of the serving inbox even for rostered members.

## Changes

1. **`ChannelsSection.tsx`** — add a `cross_team` branch that renders one row per cross-team channel, mirroring the PCO-synced rows: a `git-merge-outline` icon, `"N members · Synced"` subtitle, and tapping opens Channel Info (`/inbox/{groupId}/{slug}/info`) where the existing **Edit synced roles** control lives. Disabled/archived cross-team channels fold into the existing "Archived" section like every other row.

2. **`RosteringCrossTeamScreen.tsx`** — the Rostering → Cross-team cards were display-only. Wrapped each card in a `Pressable` (added a chevron affordance) that routes into the same Channel Info screen, reaching the existing `updateCrossTeamChannel` edit flow. No new editor.

3. **`crossTeamChannels.ts` (backend query + api type)** — `listCrossTeamChannels` now returns the channel `slug` so the Rostering card can route to Channel Info. Added `slug?: string` to the `CrossTeamChannel` type.

4. **`scheduling/serving.ts` — `resolveServingChannelIds`** — added branch (c): cross-team channels in the plan's group whose selectors reference a team on the plan. Scoped via the `by_group_type` index (no full-table scan). Membership-agnostic like the team-channel branch — the serving flatten in `messaging/channels.ts` only iterates channels already in the per-user, active-membership-gated result, so widening this Set leaks nothing (only users with an active `event_plan`-synced membership row actually see the channel). Added three tests in `crossTeamChannels.test.ts` exercising the resolver directly (included when a team matches, omitted when none match, omitted when archived).

The existing **Edit synced roles** flow (in `ChannelInfoScreen`) is reused as-is — not renamed or duplicated.

## Verification

- `tsc --noEmit` on `apps/mobile` and `apps/convex`: **no errors in any changed file** (convex: 0 errors total; mobile's pre-existing unrelated errors in chat components/tests are untouched by this PR).
- Backend test `__tests__/scheduling/crossTeamChannels.test.ts`: **16 passed** (3 new serving-mode tests).
- Component test `ChannelsSection.test.tsx`: 25 passed.
- Full mobile suite: 124 suites / 1151 passed.
- Not visually verified in a simulator.

### DIFF
--- a/apps/convex/functions/scheduling/crossTeamChannels.ts
+++ b/apps/convex/functions/scheduling/crossTeamChannels.ts
@@ -316,6 +316,7 @@ export const listCrossTeamChannels = query({
         );
         return {
           _id: channel._id,
+          slug: channel.slug,
           name: channel.name,
           description: channel.description,
           channelType: channel.channelType,
--- a/apps/convex/functions/scheduling/crossTeamChannels.ts
+++ b/apps/convex/functions/scheduling/crossTeamChannels.ts
@@ -21,7 +21,7 @@ import type { MutationCtx } from "../../_generated/server";
 import type { Id } from "../../_generated/dataModel";
 import { requireAuth } from "../../lib/auth";
 import { isLeaderRole } from "../../lib/helpers";
-import { generateChannelSlug } from "../../lib/slugs";
+import { generateChannelSlug, getChannelSlug } from "../../lib/slugs";
 import { requireGroupMember, requireGroupScheduler } from "./permissions";
 import { reconcileCrossTeamChannelImpl } from "./teamChannelSync";
 
@@ -316,7 +316,7 @@ export const listCrossTeamChannels = query({
         );
         return {
           _id: channel._id,
-          slug: channel.slug,
+          slug: getChannelSlug(channel),
           name: channel.name,
           description: channel.description,
           channelType: channel.channelType,
--- a/apps/mobile/features/groups/components/ChannelsSection.tsx
+++ b/apps/mobile/features/groups/components/ChannelsSection.tsx
@@ -114,6 +114,7 @@ export function ChannelsSection({ groupId, userRole, onChannelPress }: ChannelsS
     (c: Channel) => c.channelType === "announcements"
   );
   const pcoSyncedChannels = channels?.filter((c: Channel) => c.channelType === "pco_services") ?? [];
+  const crossTeamChannels = channels?.filter((c: Channel) => c.channelType === "cross_team") ?? [];
   const customChannels = channels?.filter((c: Channel) => c.channelType === "custom") ?? [];
 
   // A channel is inactive if it's archived OR a leader hid it (isEnabled ===
@@ -331,6 +332,29 @@ export function ChannelsSection({ groupId, userRole, onChannelPress }: ChannelsS
       name: channel.name,
       subtitle: enabled
         ? `${channel.memberCount} member${channel.memberCount !== 1 ? "s" : ""} · PCO Synced`
+        : "Disabled",
+      enabled,
+      onPress: () => navigateToChannelInfo(channel.slug),
+      unreadCount: channel.unreadCount,
+      pinned: channel.isPinned,
+      archived: !enabled,
+    });
+  });
+
+  // Cross-team channels auto-sync their membership from serving-team role
+  // assignments. They surface here (mirroring PCO synced rows) so a leader can
+  // tap through to Channel Info → "Edit synced roles". Without this branch the
+  // TYPE ALLOWLIST above silently drops them from the group's channel list.
+  crossTeamChannels.forEach((channel: Channel) => {
+    const enabled = channel.isEnabled && !channel.isArchived;
+    rows.push({
+      key: channel._id,
+      icon: "git-merge-outline",
+      iconColor: "#7C3AED",
+      iconBg: "#7C3AED15",
+      name: channel.name,
+      subtitle: enabled
+        ? `${channel.memberCount} member${channel.memberCount !== 1 ? "s" : ""} · Synced`
         : "Disabled",
       enabled,
       onPress: () => navigateToChannelInfo(channel.slug),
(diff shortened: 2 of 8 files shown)


apps/mobile/features/groups/components/ChannelsSection.tsx
```diff
@@ -340,6 +341,29 @@ export function ChannelsSection({ groupId, userRole, onChannelPress }: ChannelsS
     });
   });
 
+  // Cross-team channels auto-sync their membership from serving-team role
+  // assignments. They surface here (mirroring PCO synced rows) so a leader can
+  // tap through to Channel Info → "Edit synced roles". Without this branch the
+  // TYPE ALLOWLIST above silently drops them from the group's channel list.
+  crossTeamChannels.forEach((channel: Channel) => {
```
lilseyi, line 348: **MEDIUM — visibility: cross_team rows are not leader-gated.** This forEach is unconditional, unlike the Reach Out row above it (which is inside `if (isLeader)`). Combined with the backend filter in `messaging/channels.ts` listGroupChannels (`requiresMembership = isCustomChannel(...) || pco_services` — cross_team is NOT in that set), the query returns every *enabled* cross_team channel to regular members even when they are not on that cross-team. So after this PR a non-member group member sees these channel rows and can open Channel Info (member list + synced-role config visible read-only; the Edit synced roles control itself is leader-gated, so no privilege escalation). PCO/custom channels only appear for members who belong to them — cross_team is more permissive. Please confirm this metadata exposure is intended; if not, gate this branch behind `isLeader` (matching Reach Out) or add cross_team to the requiresMembership set in listGroupChannels.

apps/convex/functions/scheduling/crossTeamChannels.ts
```diff
@@ -316,6 +316,7 @@ export const listCrossTeamChannels = query({
         );
         return {
           _id: channel._id,
+          slug: channel.slug,
```
lilseyi: **LOW — return resolved slug, not the raw field.** `listGroupChannels` routes via `getChannelSlug(channel)`, which falls back to `channelType` when `slug` is unset. Here you return the raw `channel.slug` (optional in the schema), so the Rostering card only routes when the DB row has a slug. Cross_team channels created via `createCrossTeamChannel` always get `generateChannelSlug`, so in practice slug is present and `handleOpenChannel` guards `if (!slug) return` — no crash. But for consistency and to keep older/migrated rows routable, return `getChannelSlug(channel)` here (and drop `?` from the type). Minor.

lilseyi (commented):
🤖 Subagent review (codex unavailable) — adversarial review of PR #539

**Verdict: Minor / non-blocking.** The core change is clean and correct — the new `cross_team` branch in `ChannelsSection` mirrors the PCO/custom rows exactly and reads only fields that `listGroupChannels` actually returns (`memberCount`, `isEnabled`, `isArchived`, `slug`, `unreadCount`, `isPinned`), folds disabled/archived into the Archived section via `archived: !enabled`, and `slug` is always a real string (backend uses `getChannelSlug`). The Rostering `Pressable` uses a **static** style array (no RN-Web function-style pitfall), hooks are declared before the early returns (no hooks-order issue), and `listCrossTeamChannels` has no `returns` validator so the added `slug` is safe. Adding a field is additive — no consumer breaks.

Findings:

1. **MEDIUM — visibility inconsistency (confirm intended).** `ChannelsSection.tsx:348` — the `crossTeamChannels.forEach` is **not** gated by `isLeader` (unlike the Reach Out row). Backend `messaging/channels.ts:1356` `requiresMembership` only covers `custom` + `pco_services`, so cross_team falls through to `return true` and `listGroupChannels` hands every *enabled* cross_team channel to any group member — even non-members of that cross-team. Net effect: regular members now see cross_team rows and can open read-only Channel Info (member list + synced-role config visible; the Edit control is leader-gated, so **no privilege escalation**). This is more permissive than PCO/custom, which require channel membership. If cross_team is meant to be leader/roster-only, gate the frontend branch behind `isLeader` or add `cross_team` to `requiresMembership`. (inline: ChannelsSection.tsx:348; backend note could not be inlined — channels.ts is not in the diff.)

2. **LOW — return resolved slug.** `crossTeamChannels.ts:319` returns raw `channel.slug` (optional in schema) instead of `getChannelSlug(channel)` used everywhere else. Cross_team channels are always created with `generateChannelSlug`, and `handleOpenChannel` guards `if (!slug) return`, so no crash and no practical breakage — but returning `getChannelSlug(channel)` (and dropping `?` from the `CrossTeamChannel.slug` type) is more consistent and keeps any slug-less legacy row routable.

Nits / positives verified: no NativeWind/className usage; StyleSheet used throughout; static Pressable style (safe on web); disabled→Archived folding matches other types; route `/inbox/{groupId}/{slug}/info` exists and reaches the existing Edit synced roles flow.

Overall: ship-able once the intent of finding #1 (member visibility of cross_team channels) is confirmed. If it should be leader-only, that is a one-line gate.

PR #540, opened 2026-07-02
feat(runsheet): native viewer parity — links, live highlight, collapsible sections

lilseyi:
## What

Brings the **read-only native run sheet** (`NativeRunSheetView.tsx`) up to feature parity with the PCO run sheet viewer (`RunSheetScreen.tsx`) by porting four patterns that already exist in the PCO renderer:

1. **Rendered / tappable links** inside item `description` and `notes[].content` — URLs become rich `LinkPreviewCard`s (or an inline Dropbox video card), text stays selectable.
2. **Live "current item" highlight** — a clock ticks every 30s; the item whose `[start, start + durationSec)` window contains `now` gets a yellow background (`#FFF9E6` / dark `#2a2700`) and an amber left border (`#D4A017`).
3. **Collapsible section headers** — `type === "header"` rows fold the rows beneath them (positional association within a segment). Collapsed state is persisted to `AsyncStorage` under `runsheet_collapsed_${groupId}`, matching the PCO viewer.
4. **Expandable rows** — description/notes reveal on tap via a chevron.

## How it differs from PCO

PCO items carry a stored `startsAt`; native `eventItems` do **not**. Clock times are computed client-side from `durationSec` cascading via the existing `computeSegmentedClockTimes` util, and the live-highlight window is derived from those computed times.

## Shared util (no duplication)

To avoid copy-pasting the URL/preview logic (per CLAUDE.md "remove, don't deprecate"), the link-rendering helpers were extracted from `RunSheetScreen.tsx` into a new shared util **`apps/mobile/features/leader-tools/utils/runSheetLinks.tsx`** (`renderTextWithLinks`, `SelectableText`, the Dropbox video card, and the rich preview). Both the PCO and native viewers now import from it. The PCO refactor is behavior-preserving.

## Scope

- Frontend only. **No backend or schema changes.**
- Editable grid (`RunSheetScreen.tsx` in `features/scheduling`) is intentionally out of scope — read-only viewer only.

## Verification

- `tsc --noEmit` on the mobile app: no new type errors from these files. (The repo-wide pre-existing `@togather/shared` module-resolution error appears because the workspace package isn't linked in this environment; it is unchanged by this PR and hits ~100 files.)
- Could not drive the simulator; no visual verification was performed.


apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
```diff
@@ -366,37 +462,99 @@ function PlanRunSheet({
...
           </Text>
         ) : null}
-      </View>
+      </Pressable>
     );
   }
 
+  const hasNotes = item.notes.length > 0;
+  const hasDescription = !!item.description && item.description.trim().length > 0;
+  const hasExpandableContent = hasDescription || hasNotes;
```
lilseyi, line 551: **[medium · UX regression] Description + notes are now fully hidden until you tap the row — with no collapsed preview.**

Two things to weigh:

1. **Regression vs. the previous native behavior.** Before this PR, `NativeRunSheetView` rendered `item.description` and every note *always* (see the removed `{item.description ? …}` / `item.notes.map(…)` block). Now nothing shows until each row is individually expanded (`expandedItemIds` starts empty). For an at-a-glance run sheet, role cues and notes are exactly what a server needs to skim without tapping every row.

2. **It is also *not* at parity with the PCO viewer it is mirroring.** `RunSheetScreen.tsx` shows a collapsed **preview** when `!isExpanded` — `hasNotes && !isExpanded` renders a `numberOfLines={2}` note teaser ("N role notes" / first note), and a one-line inline description for time-range titles. This native version drops that teaser entirely, so it is strictly less informative than PCO when collapsed.

Suggest either defaulting rows expanded, or (better, for real parity) porting the PCO collapsed-preview line so notes are visible at a glance.

apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
```diff
@@ -273,6 +281,85 @@ function PlanRunSheet({
...
+  const toggleHeaderCollapsed = useCallback((headerId: string) => {
+    setCollapsedHeaders((prev) => {
+      const next = new Set(prev);
+      if (next.has(headerId)) next.delete(headerId);
+      else next.add(headerId);
+      return next;
+    });
+  }, []);
+
+  const collapsedStorageKey = `runsheet_collapsed_${groupId}`;
```
lilseyi: **[low] The AsyncStorage collapse key is shared with the PCO viewer, but the comment overstates the benefit.**

The comment says the shared `runsheet_collapsed_${groupId}` key means "the collapse state carries across both surfaces." It does not: PCO stores PCO item ids, this view stores Convex `eventItems._id`s — **different id spaces**. Collapsing a section here will never reflect in the PCO viewer (its `filterVisible`/`collapsedHeaders.has(it._id)` can only match native ids), and vice-versa. So the two viewers just share one storage bucket that accumulates a merged, ever-growing blob of ids from both — with no functional cross-surface effect.

It is *mostly* benign because both viewers load-before-persist (foreign ids are read back in and re-written, so neither clobbers the other), and a native Convex id colliding with a PCO id string is effectively impossible. But the coupling is pointless and the comment is misleading. Recommend a native-specific key (e.g. `runsheet_collapsed_native_${groupId}`) and dropping the "carries across both surfaces" claim.

apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
```diff
@@ -273,6 +281,85 @@ function PlanRunSheet({
...
+      JSON.stringify(Array.from(collapsedHeaders)),
+    ).catch((err) => console.error("Failed to save collapsed state:", err));
+  }, [collapsedHeaders, collapsedHeadersLoaded, collapsedStorageKey]);
+
+  // Live "current item": tick a clock every 30s and match the item whose
+  // computed [start, start + durationSec) window contains now. Same logic as
+  // the PCO renderer, but sourced from client-computed clock times.
+  const [now, setNow] = useState(() => Date.now());
+  useEffect(() => {
+    const interval = setInterval(() => setNow(Date.now()), 30_000);
```
lilseyi, line 351: **[low · perf] The 30s tick re-renders the entire item list even when nothing is live.**

`setNow(Date.now())` every 30s re-renders `PlanRunSheet`, and `ReadOnlyRow` is not `React.memo`-wrapped, so every row re-renders on each tick regardless of whether `currentItemId` changed — and it keeps ticking forever even for events that are days away or already over (when `currentItemId` stays `null`). Run sheets are small (tens of rows) so this is not urgent, and it matches the PCO viewer, hence low severity. If you want to trim it: memoize `ReadOnlyRow`, and/or only run the interval when `now` actually falls within the plan window.

Cleanup itself is correct — `clearInterval` on unmount, no setState-after-unmount leak.

apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
```diff
@@ -273,6 +281,85 @@ function PlanRunSheet({
...
+    return () => clearInterval(interval);
+  }, []);
+  const currentItemId = useMemo(() => {
```
lilseyi, line 354: **[low] Live highlight only tracks the *earliest* service time, and the no-times fallback can spuriously highlight.**

`clockTimes` (which this window reads) are anchored to `earliestStart = Math.min(...times.map(t => t.startsAt))`. So for a plan with two services (e.g. 10am **and** 12pm), during the 12pm service `now` is past every computed window and nothing highlights for the entire second service. Worth confirming that matches intent / the PCO viewer.

Also, `earliestStart` falls back to `Date.now()` when the plan has no `times`. In that case `clockTimes` are based on mount time and the ticking `now` can drift into an item window → a highlight appears on a run sheet with no scheduled time at all. Consider gating the highlight (and the 30s interval) on `times.length > 0`.

The window math itself is correct: `[start, start+dur)` is half-open, 0-duration items never match (`end === start`), headers are skipped, and `now` outside all windows returns `null` without crashing. Iterating back-to-front is harmless since the windows are contiguous and non-overlapping.

lilseyi (commented):
🤖 Subagent review (codex unavailable)

Reviewed the extraction + native run-sheet parity changes. Overall this is a clean, largely behavior-preserving change. **Overall risk: minor** — no blockers, one UX call worth a decision.

## Extraction (behavior-preserving? — yes)
- `runSheetLinks.tsx` is a verbatim move of the regex/`SelectableText`/`DropboxVideoPlayer`/`RunsheetLinkPreview`/`renderTextWithLinks` logic out of `RunSheetScreen.tsx`. No dropped props, no changed defaults, `useLinkPreview` wiring intact.
- Dropped imports (`TextInput`, `WebBrowser`, `useLinkPreview`, `LinkPreviewCard`) are all gone from `RunSheetScreen.tsx` and no longer referenced there; `SelectableText`/`renderTextWithLinks` are re-imported from the util (still used at lines ~1096/1130/1157). Remaining imports (`Pressable`, `Linking`, `Ionicons`, `useCallback`) are all still used elsewhere, so no unused-import fallout.

## Findings (all inline)
1. **[medium · UX] `NativeRunSheetView.tsx:544`** — description + notes are now fully hidden behind per-row tap-to-expand with **no collapsed preview**. This is a regression vs. the old native view (notes always visible) *and* not true parity: the PCO viewer shows a `numberOfLines={2}` note teaser + inline description when collapsed. Recommend defaulting expanded or porting the PCO preview line. Needs a product decision.
2. **[low] `:313`** — collapse key `runsheet_collapsed_${groupId}` is shared with the PCO viewer, but id spaces differ, so the "carries across both surfaces" comment is misleading; it is a pointless shared bucket (benign due to load-before-persist). Suggest a native-specific key.
3. **[low · perf] `:347`** — 30s tick re-renders the whole list (rows not memoized) and runs forever even when nothing is live. Matches PCO, so low priority. Cleanup is correct (no leak / setState-after-unmount).
4. **[low] `:350`** — highlight only anchors to the earliest service time (second service of a multi-time plan never highlights), and the no-`times` fallback (`Date.now()`) can spuriously highlight. Window math `[start, start+dur)` is otherwise correct — half-open, 0-dur safe, headers skipped, no-crash when outside all windows.

## Verified good
- Hooks order: all `useState`/`useCallback`/`useEffect`/`useMemo` (incl. `currentItemId`) are declared before the `event/items === undefined` early return.
- No `Pressable` function-style `style` — both new Pressables use static StyleSheet objects (safe on web).
- AsyncStorage read/write handles missing/corrupt values (try/catch, `loaded` guard self-heals a bad blob on next write).
- `colors.link` / `colors.textTertiary` exist in the theme.
- `filterVisible` positional header→rows association is sound per-segment.

No merge/approve action taken.

PR #541, opened 2026-07-02
feat(serving): multi-team / multi-role event tasks

lilseyi:
## What

One `eventTasks` row can now target **multiple teams and multiple roles** instead of a single team + single role, so tasks that many people each do for themselves no longer need duplicated rows.

## Model (confirmed semantics)

- `teamId` (single) → `teamIds` (array, **>= 1**). `roleId` (single, optional) → `roleIds` (array, may be **empty**).
- **`roleIds` NON-EMPTY ⇒ per-person completion.** Each person confirmed for **ANY** of `roleIds` sees the task in their "Mine" list and completes it individually (`eventTaskCompletions`, keyed by taskId+userId[+timeLabel] — unchanged).
- **`roleIds` EMPTY ⇒ team-level task ⇒ ONE shared completion** for the whole task (`sharedTaskCompletions`, one row per task — unchanged). Any confirmed member of **ANY** team in `teamIds` may toggle it. A team-level task spanning multiple teams is still **ONE** shared checkbox.

## Migration strategy (single-PR-safe)

- `teamIds` / `roleIds` added as **OPTIONAL** new fields; legacy `teamId` / `roleId` kept **optional-deprecated**.
- Two normalization helpers — `taskTeamIds(task)` = `task.teamIds ?? [task.teamId]`, `taskRoleIds(task)` = `task.roleIds ?? (task.roleId ? [task.roleId] : [])` — and **all reads route through them**, so behavior is correct whether or not a row has been migrated. New **writes** set only the array fields.
- Idempotent backfill `internalMutation` **`backfillTaskAssignmentArrays`** populates the arrays from the legacy columns (skips rows that already have `teamIds`).
- `// TODO(followup): drop legacy teamId/roleId after backfill runs in all envs` — dropping the legacy columns is an intentional fast-follow (Convex can't add a required field over existing data in one deploy).
- The now-unused `by_plan_team` index on `eventTasks` was dropped; `sortOrder` is now appended per **(plan, segment)** via `by_plan_segment` (a multi-team task no longer belongs to one team; the grid re-sorts by team→role for display and `reorderTasks` rewrites a plan-wide order anyway).

## Readiness-slot semantics

- **Per-person task**: slots = number of confirmed people across its `roleIds` (× the number of service times for "during"). `done` = valid completions from those people.
- **Team-level task**: **1** shared slot; `done` = 1 iff a shared completion row exists.
- **Per-team rollup** (`getPlanTaskReadiness.byTeam`, `getAllTeamsTasks`): a role belongs to exactly one team, so role tasks are **split by their owning team** → `sum(byTeam) == overall` for role tasks. A **team-level** task credits its single shared slot to **EACH** team it spans, so `sum(byTeam)` can exceed `overall` only for multi-team team-level tasks (documented in code).
- In **All-teams**, a multi-team task appears under **each** of its teams, showing only that team's roles/assignees; a team-level task shows under each team as one shared slot.

## Frontend

- **Leader grid**: Team and Role cells render multiple removable chips + a dashed "+ Add" pill (`ChipCell`). A task keeps ≥1 team. Team/Role pickers are now **multi-select** `AnchoredMenu`s (toggle, stay open, backdrop closes). Team is a normal `updateTask` edit — the old recreate-on-team-change path is gone. Removed the orphaned `PickerPill`.
- **Serving Tasks**: `SharedTask` carries `teamIds`/`teamNames`; all-teams `TeamTask` carries `roleNames[]`.

## Tests

Extended `apps/convex/__tests__/scheduling/eventTasks.test.ts`: a 2-role/2-team task in Mine for either role (per-person completion); a 2-team team-level task as one shared completion any teammate can toggle (and gated to teammates); readiness counts across mixed multi-team tasks; and the array backfill. Full Convex suite: **1938 passing**. Mobile `tsc --noEmit`: **0 errors**.

## Follow-up / notes

- Legacy `teamId`/`roleId` column drop is deferred (see TODO) and requires running `backfillTaskAssignmentArrays` in each env after deploy.
- `backfillTeamLevelSharedCompletions` was generalized to union confirmed members across all of a task's teams.


apps/convex/functions/scheduling/eventTasks.ts
```diff
@@ -293,7 +395,7 @@ export const updateTask = mutation({
     // team-wide completion. Team-level completion is single-source on
     // `sharedTaskCompletions`, so drop the stale role completions — otherwise the
     // converted task could still read as done from those rows.
-    if (args.clearRole && task.roleId) {
+    if (prevRoleIds.length > 0 && nextRoleIds.length === 0) {
```
lilseyi: **[Minor · CONFIRMED] Team-level→role conversion leaves a stale `sharedTaskCompletions` row → "done" state resurrects on convert-back.**

This cleanup fires only for the role→team-level direction (`prevRoleIds > 0 && nextRoleIds == 0`), deleting the per-user `eventTaskCompletions`. The reverse (team-level→role) has no symmetric cleanup, so a `sharedTaskCompletions` row left over from when the task was team-level is never removed.

Repro (all reachable from the grid UI): team-level task marked done (shared row exists) → add a role (now a role task; shared row ignored by role-task readers, looks fine) → remove that role again (back to team-level). `getSharedTeamTasks` / `getPlanTaskReadiness` read `sharedByTask`, so the task shows **done** from the resurrected row even though nobody re-checked it.

Fix: when `nextRoleIds.length > 0 && prevRoleIds.length === 0` (converting to a role task), also delete any `sharedTaskCompletions` rows for the task — mirror the existing block.

apps/convex/functions/scheduling/eventTasks.ts
```diff
@@ -278,8 +336,52 @@ export const updateTask = mutation({
...
+      const kept: Id<"teamRoles">[] = [];
+      for (const roleId of prevRoleIds) {
+        const role = await ctx.db.get(roleId);
+        if (role && teamIdSet.has(role.teamId as string)) kept.push(roleId);
+      }
+      if (kept.length !== prevRoleIds.length) {
+        nextRoleIds = kept;
+        patch.roleIds = kept;
+      }
+    }
```
lilseyi, line 383: **[Low · CONFIRMED] Dropping a role (task stays a role task) orphans that role's per-user completions.**

When `teamIds` change and a role is dropped here (or when `roleIds` is set to a smaller non-empty set above), the corresponding `eventTaskCompletions` rows for people who were confirmed *only* for the removed role are left in place. Readiness/serving reads filter completions to currently-expected assignees, so this is harmless for counts today — but the rows linger, and if the same role is re-added later they resurrect as "done" without anyone re-checking. Same resurrection class as the team-level→role case above. Consider deleting completions for removed roles too (or documenting the intentional keep).

apps/convex/functions/scheduling/eventTasks.ts
```diff
@@ -519,44 +653,48 @@ export const getPlanTaskReadiness = query({
...
+        done = Math.min(countValid(expectedUserIds), total);
+
+        overall.total += total;
+        overall.done += done;
+        bySegment[task.segment].total += total;
+        bySegment[task.segment].done += done;
+
+        // Per-team split: a role belongs to exactly one team, so partition the
+        // expected assignees by team. This keeps sum(byTeam) == overall for
+        // role tasks even when the task spans multiple teams/roles.
```
lilseyi, line 731: **[Nit · CONFIRMED] Documented invariant `sum(byTeam) == overall for role tasks` can break for a cross-team person.**

If one user is confirmed for two roles on two *different* teams, and both roles are on the same multi-team/multi-role task, `overall` counts them once (deduped `expectedUserIds`) while the per-team split credits each team → `sum(byTeam.done/total)` exceeds `overall`. Not user-visible (the UI renders each team tile separately and never sums; each tile is still `Math.min`-clamped to ≤100%), so no crash/NaN — but the comment's "keeps sum(byTeam) == overall" claim is inaccurate for that (uncommon) shape. Worth a one-line caveat so a future reader doesn't rely on the invariant.

lilseyi (commented):
🤖 Subagent review (codex unavailable) — data-model migration + completion-semantics rewrite. Reviewed backend (`eventTasks.ts`, `events.ts`, `schema.ts`), tests, and the four frontend files. Overall this is careful, well-documented work; the migration-independence design (all reads via `taskTeamIds()`/`taskRoleIds()`) is sound. **Verdict: approve with minor follow-ups — non-blocking.** Findings, most-severe first:

### Minor (fix soon, not blocking)

1. **CONFIRMED — Plan deletion does not cascade `sharedTaskCompletions`.** `events.ts` `deleteEvent` cascade (`~L397–430`) deletes `eventTasks`, `eventTaskCompletions`, `personalServingTasks`, etc. but **not** `sharedTaskCompletions` (nor `howToDocChecks`). Deleting a plan orphans every team-level shared-completion row for its tasks. Functionally harmless (reads are planId/taskId-scoped and the task is gone) but it's a real row leak, and cascade-completeness is in scope for this rewrite. This appears to predate #541 (the file's `deleteEvent` isn't in this diff) but should be fixed here or tracked. *(non-diff line → summary only)*

2. **CONFIRMED — team-level→role conversion leaves a stale `sharedTaskCompletions` row → "done" resurrects on convert-back.** `updateTask` `eventTasks.ts:398` cleans up only the role→team-level direction. Inline comment with repro at `L398`.

### Low / Nit

3. **CONFIRMED — Dropping a role (task stays a role task) orphans that role's `eventTaskCompletions`.** Harmless for current counts (readers filter to expected assignees) but resurrects if the role is re-added. Inline at `L383`.

4. **Nit CONFIRMED — Readiness invariant `sum(byTeam) == overall for role tasks` breaks for a user confirmed on two targeted teams of one multi-team/multi-role task** (double-counted per team, single overall). Not user-visible (UI shows per-team tiles, never sums; each `Math.min`-clamped, no NaN — consumer `pct()` guards div-by-zero). Doc caveat only. Inline at `L681`.

### Verified clean (traced, no issue)

- **Helper coverage:** every task team/role read routes through `taskTeamIds()`/`taskRoleIds()`; the only direct legacy reads are the backfill and `duplicateEvent` (which normalizes via helpers). `roleAssignments.teamId/roleId` accesses are on assignment docs, not tasks — correct. Helpers avoid `[undefined]` and treat missing arrays + empty legacy `roleId` as truly empty (team-level).
- **Completion gates:** `toggleTaskCompletion` rejects team-level tasks; `toggleSharedTeamTask` rejects role tasks, gates on `confirmed.some(a => taskTeamIds.includes(a.teamId))`, and keeps ONE shared row per taskId via `.unique()` (only ever inserts when none exists — no `.unique()` throw risk). Any member of any targeted team can toggle the single shared checkbox. ✔ matches spec.
- **`getMyServingTasks`:** per-person tasks appear when confirmed roles intersect `roleIds`; team-level excluded; multi-role appears once (no dup); `timeLabel`/no-configured-times handling preserved.
- **`getSharedTeamTasks`:** team-level shown to members of ANY targeted team, exactly once even if the user is on multiple targeted teams.
- **`getAllTeamsTasks`/`getCrewTasks`:** multi-team task appears under each team; multi-role grouped by owning team/role; per-team role partition is complete (roles ⊆ teams enforced at write).
- **`updateTask` role→team-level:** correctly deletes stale `eventTaskCompletions` (incl. removing the last role via the grid). Team validation requires ≥1 team on create and update.
- **`backfillTaskAssignmentArrays`:** idempotent (skips rows with `teamIds` set), no `[undefined]`, `roleId` null → `[]` (team-level). Safe to re-run.
- **Migration safety:** correctness does not depend on the backfill; unmigrated legacy rows behave identically via the fallbacks.
- **Schema:** `teamId`/`roleId` made optional (won't reject existing docs); new arrays optional. Dropped `by_plan_team` index was on `eventTasks` only — the sole remaining `by_plan_team` usage (`deletion.test.ts`) targets `neededRoles`, which still has it. `nextTaskSortOrder` uses the retained `by_plan_segment`. ✔
- **`duplicateEvent`:** copies via helpers → new rows array-only. Readiness UI (`ReadinessHeader`, `pct = t>0 ? d/t : 0`) guards div-by-zero; per-team tiles shown separately so the documented `sum(byTeam) > overall` never surfaces as a >100% bar.
- **Frontend:** `AnchoredMenu` multi-select (`selectedSet`, toggle stays open, backdrop closes) is correct. `ChipCell` uses static `StyleSheet` array styles on `Pressable` (no RN-Web function-style pitfall; no NativeWind), layout on inner `View`. Team/role pickers read the **live** task via `tasks.find(...)`, so checkmarks track each toggle; IIFE render bodies contain no hooks → no hooks-order issues. Team keeps ≥1 (remove of last team disabled); removing the last role → team-level (backend cleans completions). No leftover `PickerPill`/`clearRole`/`.teamId`/`.roleId` task refs.
- **Codegen (non-issue):** `_generated/api.d.ts` references each module by `typeof import(...)`, so the new internal `backfillTaskAssignmentArrays` export is type-available without a per-function regen — no `convex typecheck` break in CI.

### Notes
- `--no-verify` push / hooks-bypass claim: I could not confirm this from history; if hooks were skipped locally, CI must re-run typecheck + the full Convex suite (author reports 1938 passing / 0 tsc errors).
- Tests cover the happy paths + backfill but **not** findings 1–2 (cascade of `sharedTaskCompletions`; team-level↔role conversion round-trip). Adding those would prevent regressions.

PR #542, opened 2026-07-02
fix(channels): cross-team inbox name/badge, tab header identity, edit-roles prefill

lilseyi:
Three display bugs on `cross_team` chat channels. Mobile-only (3 files); no backend change required.

## Bug 1 — Serving inbox shows slug-style name + "Team" badge
**Root cause:** `GroupedInboxItem.getServingChannelLabel` returned a constructed `#team-cross-team` slug for cross-team channels instead of their real name, and the single-channel row always badged `group.groupTypeName` ("Team").
**Fix:** Prefer `channel.name` (slug is fallback only); badge "Cross-team" when `primaryChannel.channelType === "cross_team"`. Keyed on `channelType`, not the `isShared` heuristic (same-campus cross-team channels have `isShared=false`).

## Bug 2 — Tab header shows the GROUP identity on a cross_team tab
**Root cause:** `ConvexChatRoomScreen` header always sourced `displayName`/`displayType`/`memberCount` from the group, even when a cross-team tab was active.
**Fix:** Derive the active channel from `effectiveGroupChannels`; when it's `cross_team`, override the three `ChatHeader` props with the channel's own name / "Cross-team" / channel member count. Every other tab is unchanged. Plain const derivation placed with the other derived values (no hooks-order change).

## Bug 3 — "Edit synced roles" opens EMPTY instead of pre-checked
**Root cause:** The prefill was a one-shot copy taken at button-tap from the async, **group-scoped** `listCrossTeamChannels` query. It seeded to `[]` if that query hadn't resolved, or if the channel's home group differed from the viewing group (a shared channel opened from a linked group).
**Fix (A, reactive):** Button just opens the modal; a `useEffect` re-seeds the picker draft whenever the modal opens or the saved selectors resolve/change.
**Fix (B, durable):** Source saved selectors BY CHANNEL ID off `getChannelBySlug`'s raw `crossTeamSync.selectors` (the channel doc is spread with no `returns` validator, so the field is already surfaced — no backend change). This removes the dependency on the viewing group matching the channel's home group. The now-unused group-scoped query and its imports are removed.

## Verification
- `packages/shared` built, then `apps/mobile` `tsc --noEmit`: the three touched files are clean apart from the pre-existing repo-wide `@togather/shared` module-resolution baseline (present on `main`). No new errors introduced. All other tsc errors are in untouched files.
- **Not visually verified** — I cannot drive the simulator.

### DIFF
--- a/apps/mobile/features/chat/components/ChannelInfoScreen.tsx
+++ b/apps/mobile/features/chat/components/ChannelInfoScreen.tsx
@@ -20,7 +20,7 @@
  *   - CHANNEL ACTIONS (Share invite link, Leave channel)
  *   - LEADER CONTROLS (Active state, Rename, Share with groups, Archive)
  */
-import React, { useCallback, useMemo, useState } from "react";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
 import {
   View,
   Text,
@@ -48,9 +48,7 @@ import { CustomModal } from "@components/ui/Modal";
 import { AutoChannelSettings } from "@features/channels";
 import {
   CrossTeamSelectorPicker,
-  listCrossTeamChannelsRef,
   updateCrossTeamChannelRef,
-  type CrossTeamChannel,
   type CrossTeamSelector,
 } from "@features/scheduling";
 import { useAuth } from "@providers/AuthProvider";
@@ -208,16 +206,6 @@ export function ChannelInfoScreen({ groupId, channelSlug, channelId }: Props) {
       : "skip",
   );
 
-  // Cross-team channels: list the group's cross-team channels so we can find
-  // THIS channel's current selectors to prefill the edit picker. The backend
-  // has no per-channel getter, so we filter the group-scoped list.
-  const crossTeamChannels = useQuery(
-    listCrossTeamChannelsRef,
-    token && channel?._id && channel.channelType === "cross_team"
-      ? { token, groupId: groupId as Id<"groups"> }
-      : "skip",
-  ) as CrossTeamChannel[] | undefined;
-
   const updateCrossTeamChannel = useAuthenticatedMutation(
     updateCrossTeamChannelRef,
   );
@@ -307,11 +295,30 @@ export function ChannelInfoScreen({ groupId, channelSlug, channelId }: Props) {
   const isPco = channelType === "pco_services";
   const isCrossTeam = channelType === "cross_team";
 
-  // This channel's current cross-team config (selectors), if any.
-  const thisCrossTeamChannel = useMemo(
-    () => crossTeamChannels?.find((c) => c._id === channel?._id),
-    [crossTeamChannels, channel?._id],
+  // This channel's saved cross-team selectors, sourced BY CHANNEL ID from
+  // getChannelBySlug (the raw `crossTeamSync.selectors`). Reading it off the
+  // channel doc — rather than a group-scoped list — means the edit picker
+  // prefills correctly even for a shared channel opened from a linked group
+  // whose id differs from the channel's home group.
+  const savedCrossTeamSelectors = useMemo<CrossTeamSelector[]>(
+    () =>
+      (
+        (channel as { crossTeamSync?: { selectors?: CrossTeamSelector[] } } | undefined)
+          ?.crossTeamSync?.selectors ?? []
+      ).map((s) => ({
+        sourceTeamId: s.sourceTeamId,
+        ...(s.roleId ? { roleId: s.roleId } : {}),
+      })),
+    [channel],
   );
+
+  // Re-seed the picker draft whenever the edit modal opens or the saved
+  // selectors resolve/change. A one-shot copy at button-tap missed the case
+  // where the channel query hadn't resolved yet (draft stuck at []).
+  useEffect(() => {
+    if (!crossTeamEditVisible) return;
+    setCrossTeamDraft(savedCrossTeamSelectors);
+  }, [crossTeamEditVisible, savedCrossTeamSelectors]);
 
   const iconCfg = getChannelIconConfig(channelType, primaryColor);
   const channelDisplayName = channel?.name?.trim() || iconCfg.defaultName;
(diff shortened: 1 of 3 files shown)


apps/mobile/features/chat/components/ChannelInfoScreen.tsx
```diff
@@ -307,12 +295,31 @@ export function ChannelInfoScreen({ groupId, channelSlug, channelId }: Props) {
...
+      })),
+    [channel],
   );
 
+  // Re-seed the picker draft whenever the edit modal opens or the saved
+  // selectors resolve/change. A one-shot copy at button-tap missed the case
+  // where the channel query hadn't resolved yet (draft stuck at []).
+  useEffect(() => {
+    if (!crossTeamEditVisible) return;
+    setCrossTeamDraft(savedCrossTeamSelectors);
```
lilseyi: 🤖 Subagent review (codex unavailable) — **CONFIRMED, minor/medium (functional regression in the same flow this PR fixes)**

The reactive re-seed clobbers a leader's **in-progress, unsaved** selection.

`savedCrossTeamSelectors` is a `useMemo` over `[channel]`, and `.map()` returns a **new array reference** every time `channel` changes. `channel` comes from the reactive `useQuery(getChannelBySlug)`, which spreads the full channel doc (`...resolvedChannel`) — including `lastMessageAt`. That field is patched on **every message send** (`messaging/messages.ts:895,1125`). So while the "Edit synced roles" modal is open (the query stays subscribed), any new message in this channel → new `channel` ref → new `savedCrossTeamSelectors` ref → this effect re-fires → `setCrossTeamDraft(savedCrossTeamSelectors)` resets the picker back to the saved selectors, discarding whatever the leader had toggled.

Cross-team channels receive messages and auto-sync membership, so this is realistically hit during an edit session. The old one-shot `handleOpenCrossTeamEdit` didn't have this problem.

Suggested fix: seed only on the modal **open transition** (track prev-open with a ref) instead of on every `savedCrossTeamSelectors` identity change — e.g. fire when `crossTeamEditVisible` flips false→true, or gate with `if (crossTeamDraft.length === 0)`.

apps/mobile/features/chat/components/ConvexChatRoomScreen.tsx
```diff
@@ -1172,15 +1180,19 @@ const ConvexChatRoomScreenInner: React.FC = () => {
           />
         ) : (
           <ChatHeader
-            displayName={displayName}
-            displayType={displayType}
+            displayName={activeIsCrossTeam ? activeChannel.name : displayName}
```
lilseyi, line 1183: 🤖 Subagent review — **LGTM / verified safe (Bug 2)**

Undefined-guard is correct: `activeIsCrossTeam = activeChannel?.channelType === "cross_team"` is `false` when `activeChannel` is undefined (no active tab / id mismatch), so `activeChannel.name` (line 1183) and `activeChannel.memberCount` (line 1194) are only read on the truthy branch — no crash on the common/no-match path.

Data source confirmed: `effectiveGroupChannels` is `listGroupChannels`, whose `returns` validator includes `channelType`/`name`/`memberCount`, and its filter (channels.ts:1359-1365) **includes** `cross_team` channels for members/leaders — so the active cross_team channel is present with those fields populated.

No hooks-order issue: `activeChannel`/`activeIsCrossTeam` are plain const derivations placed with the other derived values, and there is only one `activeChannel` declaration (the `const active` at line ~343 is a differently-named local inside a `useMemo`).

Minor (non-blocking): on a cross_team tab the header still uses the group's `displayImage`/`groupTypeId` (only name/type/count are overridden), so the avatar/color remain the group's. Likely intended since cross_team channels have no own image.

apps/mobile/features/chat/components/GroupedInboxItem.tsx
```diff
@@ -114,7 +114,9 @@ function getServingChannelLabel(teamName: string, channel: ChannelData): string
     case "announcements":
       return `#${teamSlug}-announcements`;
     case "cross_team":
-      return `#${teamSlug}-cross-team`;
+      // Cross-team channels always carry a distinct 1-50 char name; prefer it.
+      // The slug is only a fallback for the rare unnamed channel.
+      return channel.name && channel.name.trim() ? channel.name : `#${teamSlug}-cross-team`;
```
lilseyi, line 119: 🤖 Subagent review — **LGTM (Bug 1)**

Verified safe: `ChannelData.name` is a required `string` (GroupedInboxItem.tsx:39), and the guard `channel.name && channel.name.trim() ? channel.name : slug` handles the empty/whitespace case, so no raw slug leaks for a normally-named channel and no crash for an unnamed one.

Scope confirmed: `getServingChannelLabel` has exactly **one** call site (line 386, serving mode only), so the cross_team change doesn't affect any other row. `primaryChannel`/`channelType` are in scope at both the label and badge sites (inside the `if (isSingleChannel && primaryChannel)` block). `badgeLabel` keying on `channelType` (not `isShared`) is correct per the same-campus `isShared=false` case.

lilseyi (commented):
🤖 Subagent review (codex unavailable) — adversarial pass on the 3 cross-team display fixes. Checked out HEAD `8b74783`; verified against backend queries/schema.

## Findings (most severe first)

**1. [minor/medium · CONFIRMED] `ChannelInfoScreen.tsx:318-321` — reactive re-seed clobbers unsaved edits.**
`savedCrossTeamSelectors` (`useMemo` over `[channel]`) returns a new array ref whenever the reactive `getChannelBySlug` result changes. That result spreads the full channel doc incl. `lastMessageAt`, which is patched on every message send (`messaging/messages.ts:895,1125`). So while the "Edit synced roles" modal is open, an incoming message → new ref → the effect re-fires → `setCrossTeamDraft(...)` resets the leader's in-progress selection to the saved value. The fix solves "empty on first open" but reintroduces flakiness in the same flow. Suggest seeding on the modal open transition (prev-open ref) rather than on every selectors identity change. See inline.

**2. [LGTM · Bug 2 header identity] `ConvexChatRoomScreen.tsx`.** Undefined-guard is safe (`activeIsCrossTeam` is false when `activeChannel` is undefined; `.name`/`.memberCount` only read on the truthy branch). `effectiveGroupChannels` = `listGroupChannels`, which returns cross_team channels with `channelType`/`name`/`memberCount` (returns validator + filter at channels.ts:1359-1365 confirmed). Plain const, no hooks-order/redeclaration issue. Minor: group image/color still shown on cross_team tabs (likely intended).

**3. [LGTM · Bug 3 data source] `ChannelInfoScreen.tsx`.** `getChannelBySlug` returns `{...resolvedChannel}` with no `returns` validator, and schema defines `crossTeamSync.selectors: {sourceTeamId, sourceChannelId?, roleId?}` — so the field IS surfaced; the prefill is not void. Mapped shape `{sourceTeamId, roleId?}` matches `CrossTeamSelectorPicker`'s `CrossTeamSelector`. Removed `listCrossTeamChannels` query + `listCrossTeamChannelsRef`/`CrossTeamChannel` imports are fully gone — no dangling refs (`thisCrossTeamChannel`/`crossTeamChannels` grep clean); `useEffect` import now used. (The clobber in finding 1 is the only issue with this fix.)

**4. [LGTM · Bug 1 inbox label/badge] `GroupedInboxItem.tsx`.** `ChannelData.name` is a required string with an empty-guard fallback (no slug leak). Single call site (serving mode). `primaryChannel`/`channelType` in scope. Badge keyed on `channelType` — correct.

**5. [conventions] Clean.** No NativeWind/className; no leftover unused imports; RN-Web Pressable pattern unchanged. Non-cross-team (common) paths are untouched in all three files.

## Verdict
Approvable after addressing finding 1 (the re-seed clobber). Findings 2-5 are LGTM. No crashes or common-path regressions found.

PR #544, opened 2026-07-02
feat(channels): permanent members + role-synced breakdown for cross-team channels

lilseyi:
## What & why

On a cross-team chat channel's **Channel Info** page, a leader can now add/remove **permanent** members (manually pinned, not role-synced), and the members list is split into two sections that clearly distinguish the two kinds of membership. This also fixes the previously **no-op "Add people" button** for cross-team channels.

### Two-section membership UX
- **Synced by role** — derived live from the roster; one **read-only** card per `(person, role)` they currently match, shown as `Team · Role`. Managed via *Edit synced roles*. A person matching multiple roles gets one card per role.
- **Permanent** — manually pinned members, each shown as *Added manually* with a remove (–) action; a header **[+ Add]** opens an in-screen people picker.
- A person who is **both** permanent **and** currently role-matched appears in **both** sections.

## Data model — `isPermanent`
Adds `isPermanent: v.optional(v.boolean())` to `chatChannelMembers` (additive/optional, existing rows validate). Meaning: this member is manually pinned. It's **independent of `syncSource`** — a single active row can be BOTH `isPermanent` and role-synced (`syncSource === "event_plan"`).

- **Reconcile never removes permanent rows.** The `reconcileChannelMembership` removal loop now skips any row with `isPermanent === true`, so a pinned member stays in the channel even when they drop off the rotation window (including the both-case). The add-guard is unchanged: an existing active row still causes a `continue`, so role cards are derived live rather than from duplicate rows.

## Backend
- **Shared helper `computeCrossTeamRoleMembers(ctx, channel)`** — factors the selector × `roleAssignments` matching out of `reconcileCrossTeamChannelImpl` into one reusable, read-only pass (accepts `QueryCtx | MutationCtx`). Used by the reconcile (behavior unchanged), the new query, and the remove mutation, so all three derive roster status from the exact same data. (It returns the full in-window assignment list, not a deduped `{userId, roleId, teamId}` — the reconcile needs `planId`/`eventDate` for scheduled removal; the query dedupes to one card per `(user, role)` itself.)
- **New query `getCrossTeamChannelMembership(channelId)`** — auth = active member of the channel's home group (or community admin). Returns `permanentMembers` and `syncedRoleMembers` (one entry per `(user, role)`), with a `returns` validator.
- **Channel-keyed mutations** modeled on the team-keyed `addPermanentMember`/`removePermanentMember` in `teams.ts`, gated to cross-team channels and group leaders (`ConvexError`):
  - `addPermanentMemberToChannel` — patches `isPermanent: true` on an existing active row (a synced person becomes also-permanent, no duplicate); revives a soft-left row; else inserts a fresh manual row. Idempotent.
  - `removePermanentMemberFromChannel` — clears `isPermanent`, then decides from the **live roster** (`computeCrossTeamRoleMembers`), not the stale `syncSource` tag: if the user is still role-matched the row stays active (unpin only), otherwise it's soft-removed. This correctly handles an off-roster synced+pinned member (still evicted) and a later-rostered purely-permanent member (only unpinned). Throws if not an active permanent member.
- Typed mobile refs mirror the existing `crossTeamChannels` refs (the generated api map can be stale offline).

## Add-people fix (frontend)
`ChannelInfoScreen` "Add people" previously routed to `/members`, whose add UI is gated to `custom` channels — a dead end for cross-team. For cross-team it now opens an in-screen modal using the existing **`MemberSearch`** picker (candidate list = group members, minus already-pinned), whose selection calls `addPermanentMemberToChannel`. The custom-channel path is unchanged.

## Also: serving-team calendar icon in Channel Info
A serving-team channel (`custom` + `isServingTeam`) now shows the green `calendar-outline` glyph (`#10B981`) as its Channel Info hero avatar — matching the group channels list — instead of the generic chat bubble. Plain custom channels keep the bubble; cross-team keeps its `git-merge` fork. `isServingTeam` comes from `getChannelBySlug` (spreads the raw channel doc).

## Tests
`apps/convex/__tests__/scheduling/crossTeamPermanentMembers.test.ts` (12 tests): add (manual row / idempotent / synced→also-permanent no dup / non-leader rejected), remove (purely-permanent soft-removed; both-case only unpinned; **off-roster synced+pinned soft-removed**; **later-rostered purely-permanent only unpinned**; not-permanent throws), reconcile never removes a permanent row off-roster, and the membership query (permanent list + per-`(user,role)` synced list, both-member in both, non-group-member rejected).

## Verification
- `apps/convex` `tsc --noEmit`: clean (only the pre-existing `@togather/shared/config` baseline in untouched files).
- `apps/mobile` `tsc --noEmit`: my files clean (ChannelInfoScreen only hits the known `@togather/shared` baseline import that predates this PR).
- Scheduling suite green: full `__tests__/scheduling/` = **285 passed** (12 new permanent-member tests).
- Generated `api` types for the new functions resolve via CI's convex typecheck (`npx convex codegen` can't run here without a deployment); tests exercise them at runtime.
- **No visual/simulator verification** — I can't drive the simulator, so the two-section layout, picker UX, and the calendar icon are unverified in-app.

### DIFF
--- a/apps/convex/__tests__/scheduling/crossTeamPermanentMembers.test.ts
+++ b/apps/convex/__tests__/scheduling/crossTeamPermanentMembers.test.ts
@@ -1,0 +1,476 @@
+/**
+ * Tests for PERMANENT members of a cross-team channel + the two-section
+ * Channel Info membership query (`crossTeamChannels.ts`).
+ *
+ * A cross-team channel's roster is otherwise entirely auto-synced from event-
+ * plan role assignments. A "permanent" member is a `chatChannelMembers` row a
+ * leader pinned by hand (`isPermanent === true`); the reconcile engine never
+ * removes it. These are keyed by the cross-team `channelId` (which owns no
+ * `teams` row).
+ *
+ *   - `addPermanentMemberToChannel` pins a member (idempotent; converting an
+ *     existing synced member just flags it, no duplicate row),
+ *   - `removePermanentMemberFromChannel` soft-removes a purely-permanent
+ *     member but only unpins one who is also role-synced,
+ *   - `reconcileCrossTeamChannel` never soft-removes a permanent row,
+ *   - `getCrossTeamChannelMembership` returns the permanent list plus one
+ *     "synced by role" card per (user, role), with both-members in both lists.
+ */
+
+import { describe, it, expect, afterEach } from "vitest";
+import { ConvexError } from "convex/values";
+import { convexTest } from "convex-test";
+import schema from "../../schema";
+import { modules } from "../../test.setup";
+import { api, internal } from "../../_generated/api";
+import type { Id } from "../../_generated/dataModel";
+import { generateTokens } from "../../lib/auth";
+import { buildSchedulingWorld, ts, type SchedulingWorld } from "./fixtures";
+
+process.env.JWT_SECRET = "test-jwt-secret-for-unit-tests-minimum-32-chars";
+
+let activeHandle: ReturnType<typeof convexTest> | null = null;
+
+afterEach(async () => {
+  if (activeHandle) {
+    await activeHandle.finishInProgressScheduledFunctions();
+    activeHandle = null;
+  }
+});
+
+const DAY = 86400000;
+
+/**
+ * A cross-team world: the base scheduling world (the "worship" team, with a
+ * second "Worship Leader" role added), plus a cross-team channel that selects
+ * everyone from the worship team. The channel lives in `base.groupId`, which
+ * `groupLeaderId` leads — so leader-gated mutations and the member-gated query
+ * both authorize for `groupLeaderId`.
+ */
+interface CrossTeamWorld extends SchedulingWorld {
+  worshipRoleId: Id<"teamRoles">;
+  crossChannelId: Id<"chatChannels">;
+  worshipUserId: Id<"users">;
+}
+
+async function setupWorld(): Promise<{
+  t: ReturnType<typeof convexTest>;
+  world: CrossTeamWorld;
+}> {
+  const t = convexTest(schema, modules);
+  activeHandle = t;
+  const base = await buildSchedulingWorld(t);
+
+  const extra = await t.run(
+    async (
+      ctx,
+    ): Promise<{
+      worshipRoleId: Id<"teamRoles">;
+      worshipUserId: Id<"users">;
+    }> => {
+      const worshipRoleId = await ctx.db.insert("teamRoles", {
+        teamId: base.teamId,
+        communityId: base.communityId,
+        name: "Worship Leader",
+        sortOrder: 1,
+        defaultNeeded: 1,
+        isArchived: false,
+        createdAt: ts(),
+        createdById: base.channelAdminId,
+      });
+      const worshipUserId = await ctx.db.insert("users", {
+        firstName: "Worshipper",
+        lastName: "Test",
+        email: "worshipper@example.com",
+        isActive: true,
+        roles: 1,
+        createdAt: ts(),
+        updatedAt: ts(),
+      });
+      return { worshipRoleId, worshipUserId };
+    },
+  );
+
+  const crossChannelId = await t.run((ctx) =>
+    ctx.db.insert("chatChannels", {
+      groupId: base.groupId,
+      communityId: base.communityId,
+      name: "Service Leads",
+      channelType: "cross_team",
+      memberCount: 0,
+      isArchived: false,
+      createdById: base.channelAdminId,
+      createdAt: ts(),
+      updatedAt: ts(),
+      crossTeamSync: {
+        selectors: [{ sourceTeamId: base.teamId }],
+      },
+    }),
+  );
+
+  return { t, world: { ...base, ...extra, crossChannelId } };
+}
+
+/** Insert a plan + a roleAssignment on the worship team, inside the window. */
+async function roster(
+  t: ReturnType<typeof convexTest>,
+  world: CrossTeamWorld,
+  opts: { userId: Id<"users">; roleId: Id<"teamRoles">; dayOffset?: number },
+): Promise<Id<"roleAssignments">> {
+  const eventDate = Date.now() + (opts.dayOffset ?? 3) * DAY;
+  return t.run(async (ctx) => {
+    const planId = await ctx.db.insert("eventPlans", {
+      groupId: world.groupId,
+      communityId: world.communityId,
+      title: "Sunday Service",
+      eventDate,
+      times: [{ label: "9 AM", startsAt: eventDate }],
+      status: "draft",
+      createdById: world.groupLeaderId,
+      createdAt: ts(),
+      updatedAt: ts(),
+    });
+    return ctx.db.insert("roleAssignments", {
+      planId,
+      teamId: world.teamId,
+      roleId: opts.roleId,
+      userId: opts.userId,
+      eventDate,
+      status: "unconfirmed",
+      assignedById: world.groupLeaderId,
+      assignedAt: Date.now(),
+    });
+  });
+}
+
+/** All rows (active + soft-left) for a user on the cross-team channel. */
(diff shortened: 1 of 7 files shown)


apps/convex/functions/scheduling/crossTeamChannels.ts
```diff
@@ -327,3 +332,328 @@ export const listCrossTeamChannels = query({
...
+    const stillRoleMatched = roleMembers.some(
+      (m) => m.userId === args.userId,
+    );
+
+    if (!stillRoleMatched) {
+      // No live role match — unpinning removes them from the channel entirely.
+      await ctx.db.patch(active._id, { isPermanent: false, leftAt: Date.now() });
+    } else {
+      // Still a live synced member — just unpin, leave the row active.
+      await ctx.db.patch(active._id, { isPermanent: false });
```
lilseyi: 🤖 Subagent review (codex unavailable)

**[BLOCKING · CONFIRMED] Unpinning a role-matched *manual* row leaks a permanent, un-removable ghost member.**

The mutation's promise (docstring + the confirm dialog "If they're still on the roster they'll stay via their role") only holds when the active row carries `syncSource === "event_plan"`. It breaks for the **pin-first-then-roster** ordering:

1. Pin `U` while *not* rostered → row `{isPermanent:true, syncSource:undefined, active}`.
2. `U` gets rostered + reconcile runs. The add/refresh loop finds no `event_plan` row for `U`, hits the prior-rows path, sees the active manual row and `if (activeRow) continue;` (teamChannelSync.ts ~L212–215) — **the row is never stamped with `syncSource`.** (Your own test "only unpins a purely-permanent member who is now live-rostered" asserts `syncSource` stays `undefined` here.)
3. Unpin → this `else` branch sets `{isPermanent:false}`, row stays active with `syncSource:undefined`.
4. Roster window passes. Reconcile's removal loop only iterates `syncedMembers` (rows where `syncSource === "event_plan"`). This row is **not** in that set, so it is **never soft-removed.**

Result: `U` is stuck active in the channel forever — counted in `memberCount`, can read/post — yet shows in **neither** Channel Info section (not permanent, not currently role-matched). And `removePermanentMemberFromChannel` throws ("not a permanent member") because `isPermanent` is now false, and the flat Members list is hidden for cross_team channels — so there is **no UI path to remove them.**

Fix: when unpinning a still-role-matched row that lacks `syncSource`, hand it to the reconcile engine (stamp `syncSource:"event_plan"` + metadata) so it's evicted when the role ends — or convert active manual rows to synced in the reconcile add-loop when they land in `desired`. This edge is untested (see summary).

apps/convex/functions/scheduling/crossTeamChannels.ts
```diff
@@ -327,3 +332,328 @@ export const listCrossTeamChannels = query({
...
+    channelId: v.id("chatChannels"),
+    userId: v.id("users"),
+    added: v.boolean(),
+  }),
+  handler: async (ctx, args) => {
+    const callerId = await requireAuth(ctx, args.token);
+    await requireCrossTeamChannelLeader(ctx, args.channelId, callerId);
+
+    const user = await ctx.db.get(args.userId);
+    if (!user) {
```
lilseyi, line 553: 🤖 Subagent review (minor · PLAUSIBLE)

**No server-side check that the pinned user is a group/community member.** `addPermanentMemberToChannel` only verifies the user *exists* (`ctx.db.get(args.userId)`), not that they belong to the channel's home group or community. A group leader calling the mutation directly (bypassing the `groupId`-scoped `MemberSearch` picker) could pin an arbitrary user into the channel, granting them read/post access.

This matches the existing `teams.ts addPermanentMember` pattern (which also doesn't verify), and the UI does scope the picker to group members, so it's not a new regression — but since cross-team channels can span multiple groups, a `requireGroupMember(ctx, channel.groupId, args.userId)` guard here would close the gap cleanly. Non-blocking; flagging for a decision.

apps/convex/__tests__/scheduling/crossTeamPermanentMembers.test.ts
```diff
@@ -0,0 +1,566 @@
...
+      },
+    );
+
+    const rows = await rowsFor(t, world, world.worshipUserId);
+    expect(rows.length).toBe(1);
+    expect(rows[0].leftAt).toBeDefined();
+    expect(rows[0].isPermanent).toBe(false);
+  });
+
+  it("only unpins a purely-permanent member who is now live-rostered", async () => {
```
lilseyi, line 396: 🤖 Subagent review (minor · CONFIRMED gap)

This test stops one step short of the bug flagged on `crossTeamChannels.ts:653`. It asserts the row stays active + unpinned after unpin, but never runs a **reconcile afterward with the roster expired**. Because this row has `syncSource === undefined` (asserted on L~404), the reconcile removal loop — which only iterates `syncSource === "event_plan"` rows — can never evict it, leaving a permanent un-removable ghost member. Add a follow-up assertion: move the assignment out of the window, `reconcileCrossTeamChannel`, and expect the row to be soft-removed (it currently would NOT be).

lilseyi (commented):
🤖 Subagent review (codex unavailable)

Traced the reconcile/sync logic, both mutations, the query, and the frontend against the code; ran the new test file (12/12 green). Solid, well-documented PR with one real correctness gap.

**Findings (severe first):**

1. **[BLOCKING · CONFIRMED] Un-removable ghost member after unpin.** `crossTeamChannels.ts:653` (`removePermanentMemberFromChannel`). For the pin-first-then-roster ordering, the active row keeps `syncSource === undefined` (reconcile's add-loop skips already-active rows, teamChannelSync.ts ~L212). Unpinning a still-role-matched such row leaves it active but un-owned by the reconcile engine (removal loop only touches `syncSource==="event_plan"`), so it is never evicted when the roster ends — and can't be removed via the UI (remove mutation throws since `isPermanent` is now false; flat Members list is hidden for cross_team). Contradicts the mutation docstring + confirm dialog. Fix: stamp `syncSource` when unpinning a role-matched manual row, or convert active manual rows to synced in the reconcile add-loop. Details inline. Untested — see finding 4.

2. **[minor · PLAUSIBLE] No group-membership check on add.** `crossTeamChannels.ts:548`. `addPermanentMemberToChannel` verifies only that the user exists. Matches existing `teams.ts` pattern and the picker scopes to the group, so not a regression, but a direct call could pin an arbitrary user. Consider `requireGroupMember(ctx, channel.groupId, args.userId)`.

3. **[nit] Query fan-out.** `getCrossTeamChannelMembership` re-fetches user/role/team per (user,role) card (repeated user/team gets) and collects all active rows to filter `isPermanent` in JS. Fine at expected channel sizes; note only.

4. **[minor · CONFIRMED gap] Missing test.** The "now live-rostered" test (L383) stops before a post-unpin reconcile with an expired roster — exactly the scenario that leaks finding 1.

**Verified correct (no action):**
- Reconcile guard `if (member.isPermanent === true) continue;` sits only in the removal loop; non-permanent rows behave exactly as before; a both-synced-and-permanent in-window member is refreshed (not duplicated), off-window is preserved (not removed).
- `computeCrossTeamRoleMembers` extraction is behavior-preserving (same window/selector/decline filtering); reconcile desired-set/scheduledRemoval unchanged.
- Query dedup by (userId, roleId) correct; both-member appears in both lists per design; auth = `requireGroupMember`.
- add idempotency (active→flag, soft-left→revive, none→insert) and memberCount refresh correct; leader gating via ConvexError.
- Schema `isPermanent: v.optional(v.boolean())` is additive-safe; no index/validator breakage.
- Frontend: hooks all declared before the early returns; two-section render + reused `MemberSearch` (mode/groupId/excludeUserIds) wired to the right mutations; remove has a ConfirmModal; serving-team icon (`custom && isServingTeam`) resolves — `getChannelBySlug` spreads the full doc and `isServingTeam` is a schema field; no NativeWind/className; layout styles on inner Views.

**Verdict: request changes on finding 1 (blocking) before merge; 2–4 are minor/optional.**

PR #545, opened 2026-07-02
feat(channels): surface rename (custom/cross-team/serving-team) + serving-inbox display names

lilseyi:
## Summary

Two related channel-naming features.

### A. Surface channel rename for custom / cross-team / serving-team channels

- **Widened rename gate.** The rename UI was gated to plain `custom` channels only. It's now gated on a new `isRenameable` (`custom || cross_team`), which also covers **serving-team** channels (they are `custom` + `isServingTeam`). `pco_services` and system channels (main/leaders/announcements/reach_out) stay non-renameable.
- **Tappable hero title.** The channel title on the info screen is now tappable (with a `create-outline` pencil affordance) for leaders on renameable channels, opening the same rename flow as the Rename row. Non-renameable titles stay static.
- **Serving-team → team rename mirroring.** `updateChannel` now trims + validates the name (1-50 chars, must contain a letter or number — matching `createCustomChannel` / `updateTeam`), and when the channel `isServingTeam`, mirrors the rename onto the linked `teams` record via the `by_channel` index. Net effect: renaming a serving-team channel updates **both** `teams.name` and `chatChannels.name`, so the roster and the conversation stay in sync everywhere. The frontend just calls `updateChannel` for every renameable type.

### B. Event-mode (serving) inbox — one naming convention + group subtext

- **Display-name normalization.** `getServingChannelLabel` previously mixed `#slug` labels and display names. It now always returns the humanized display name (the channel's own `name` when present, else a readable team-scoped fallback like `Worship Team General`) and never a raw `#slug`. The now-unused `slugifyChannelLabel` helper was removed.
- **Parent-group subtext.** Each serving row now shows a subtle `in {group.name}` line (textTertiary, single line) under the title so flattened team channels stay distinguishable.
- Non-serving (normal) inbox rendering is unchanged. Serving mode always flattens to one channel per row in `getInboxChannels`, so the multi-channel expanded path is never hit in serving mode and is left as-is.

## Testing

- `apps/convex/__tests__/scheduling/teams.test.ts` extended: renaming a serving-team channel updates both the channel and the team, and an empty/whitespace name is rejected. All 28 tests pass.
- Convex `tsc --noEmit` and mobile `tsc --noEmit` clean for the touched files (only the known repo-wide `@togather/shared` baseline errors remain, on untouched import lines).
- **Not visually verified** — the simulator/browser was not driven for this change.

### DIFF
--- a/apps/convex/functions/messaging/channels.ts
+++ b/apps/convex/functions/messaging/channels.ts
@@ -2173,7 +2173,18 @@ export const updateChannel = mutation({
     };
 
     if (args.name !== undefined) {
-      updates.name = args.name;
+      // Validate: trim, 1-50 chars, must contain a letter or number. Mirrors
+      // `createCustomChannel` / `updateTeam` so a rename can't blank out a name.
+      const trimmedName = args.name.trim();
+      if (trimmedName.length < 1 || trimmedName.length > 50) {
+        throw new ConvexError("Channel name must be 1-50 characters.");
+      }
+      if (!/[a-zA-Z0-9]/.test(trimmedName)) {
+        throw new ConvexError(
+          "Channel name must contain at least one letter or number."
+        );
+      }
+      updates.name = trimmedName;
     }
     if (args.description !== undefined) {
       updates.description = args.description;
@@ -2190,6 +2201,22 @@ export const updateChannel = mutation({
     }
 
     await ctx.db.patch(args.channelId, updates);
+
+    // Serving-team channels back a `teams` record. Keep the team name in sync
+    // with the channel name so the roster and the conversation stay labelled
+    // consistently everywhere (mirrors `updateTeam` in scheduling/teams.ts).
+    if (channel.isServingTeam && updates.name !== undefined) {
+      const team = await ctx.db
+        .query("teams")
+        .withIndex("by_channel", (q) => q.eq("channelId", args.channelId))
+        .first();
+      if (team) {
+        await ctx.db.patch(team._id, {
+          name: updates.name,
+          updatedAt: Date.now(),
+        });
+      }
+    }
   },
 });
 
--- a/apps/convex/__tests__/scheduling/teams.test.ts
+++ b/apps/convex/__tests__/scheduling/teams.test.ts
@@ -251,6 +251,42 @@ describe("updateTeam", () => {
   });
 });
 
+describe("updateChannel on a serving-team channel", () => {
+  it("mirrors the channel rename onto the underlying team", async () => {
+    const { t, world } = await setupSchedulingWorld();
+    const leaderToken = (await generateTokens(world.groupLeaderId)).accessToken;
+
+    await t.mutation(api.functions.messaging.channels.updateChannel, {
+      token: leaderToken,
+      channelId: world.channelId,
+      name: "  Renamed From Channel  ",
+    });
+
+    await t.run(async (ctx) => {
+      const channel = await ctx.db.get(world.channelId);
+      // Name is trimmed and applied to the channel...
+      expect(channel?.name).toBe("Renamed From Channel");
+
+      // ...and mirrored onto the linked team so the two stay in sync.
+      const team = await ctx.db.get(world.teamId);
+      expect(team?.name).toBe("Renamed From Channel");
+    });
+  });
+
+  it("rejects an empty / whitespace-only name with a ConvexError", async () => {
+    const { t, world } = await setupSchedulingWorld();
+    const leaderToken = (await generateTokens(world.groupLeaderId)).accessToken;
+
+    await expect(
+      t.mutation(api.functions.messaging.channels.updateChannel, {
+        token: leaderToken,
+        channelId: world.channelId,
+        name: "   ",
+      }),
+    ).rejects.toThrow(ConvexError);
+  });
+});
+
 describe("archiveTeam", () => {
   it("archives the team and its channel, purging synced members", async () => {
     const { t, world } = await setupSchedulingWorld();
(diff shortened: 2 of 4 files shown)


apps/convex/functions/messaging/channels.ts
```diff
@@ -2190,6 +2201,22 @@ export const updateChannel = mutation({
     }
 
     await ctx.db.patch(args.channelId, updates);
+
+    // Serving-team channels back a `teams` record. Keep the team name in sync
+    // with the channel name so the roster and the conversation stay labelled
+    // consistently everywhere (mirrors `updateTeam` in scheduling/teams.ts).
+    if (channel.isServingTeam && updates.name !== undefined) {
```
lilseyi, line 2208: **AUTH — verified, no privilege escalation (CONFIRMED, acceptable).**

This mutation now renames the linked `teams` record. I checked whether that lets a caller bypass the stricter permission on the dedicated `updateTeam` path:

- `updateChannel` auth (channels.ts:2159-2163): `isChannelAdmin` = channel membership `role === "admin"` **OR** `isGroupLeader` = `isLeaderRole(groupMembership?.role)` (LEADER_ROLES = `["leader"]`).
- `updateTeam` → `requireTeamScheduler` → `isTeamScheduler` (scheduling/permissions.ts): channel role in `SCHEDULER_CHANNEL_ROLES` = `{admin, moderator}` **OR** group leader **OR** community admin.

Every caller who passes `updateChannel` (channel `admin`, or group `leader`) is already inside `isTeamScheduler`'s allow-set (`admin` ⊆ `{admin,moderator}`; same `leader` check). So the channel-rename path grants **no** team-rename ability the caller didn't already have via `updateTeam`. If anything `updateChannel` is narrower (excludes channel moderators + community admins). **Not a privilege inconsistency.**

lilseyi, line 2208: **Nit (LOW, CONFIRMED): team is re-patched even when the name is unchanged.** The guard is `updates.name !== undefined`, which is true whenever `args.name` was passed — including when the new name equals the current channel/team name. The `updateTeam` mirror has the same shape, so this matches existing behavior; it's an idempotent write (bumps `updatedAt` with no real change), not a correctness bug. Fine to leave.

Everything else here is correct: `by_channel` `.first()` is the right 1:1 lookup, the no-team case is handled (`if (team)`), `updatedAt` is set, `channel.isServingTeam` is read off the fetched doc, and the mirrored `updates.name` is the already-trimmed/validated value — no invalid name can reach `teams.name`.

lilseyi (commented):
🤖 Subagent review (codex unavailable) — adversarial pass on rename + serving-inbox naming

**Verdict: LGTM.** I tried to break the auth/validation/mirroring story and it holds up. No blocking issues; one cosmetic nit.

**1. AUTH — the important one (CONFIRMED, safe).** `updateChannel` now also renames the linked team for `isServingTeam` channels. Checked against the dedicated `updateTeam` path (`requireTeamScheduler` → `isTeamScheduler`): `updateChannel`'s allow-set (channel `role==="admin"` OR group `leader`) is a **subset** of `isTeamScheduler`'s (`{admin,moderator}` OR leader OR community admin). So no caller gains team-rename power they lacked. Not an escalation. Details inline at channels.ts:2208.

**2. Name validation (CONFIRMED correct).** Applies to all renameable types (validation is type-agnostic, runs on any `args.name`). Rule (`trim`, 1-50, `/[a-zA-Z0-9]/`) matches `createCustomChannel` and `validateName` in scheduling/teams.ts exactly. Throws **before** `ctx.db.patch`, and the team mirror uses the already-trimmed `updates.name` — no path lets an invalid/blank name reach `teams.name`. Empty/whitespace → `ConvexError`.

**3. Serving-team mirroring (CONFIRMED correct, 1 nit).** `by_channel` `.first()` is the correct 1:1 lookup; no-team case handled (`if (team)`); `updatedAt` bumped; `isServingTeam` read off the fetched doc. Nit (LOW): re-patches the team even when the passed name equals the current one — idempotent, mirrors `updateTeam`, harmless. Inline at channels.ts:2208.

**4. cross_team rename (CONFIRMED fine).** `updateChannel` gate is just `channel.groupId` present, which cross_team has; group-leader auth applies; cross_team isn't `isServingTeam` so the team-mirror block is skipped. Name is independent of sync/selectors — no interaction.

**5. Frontend gating (CONFIRMED fine).** Tappable hero title only when `isRenameable && isLeader`; otherwise a plain `<Text>` with no dangling `onPress`. `openRename`/`handleRename`/`handleSaveHint` `useCallback`s are all declared **before** the early returns (channel undefined/null), so hooks order is stable. Pencil affordance uses `TouchableOpacity` + StyleSheet with layout on an inner `<View>` — avoids the RN-Web `Pressable` function-style pitfall. `#` prefix preserved: custom/serving-team → `#name`, cross_team → plain name.

**6. Inbox normalization (CONFIRMED fine).** Every `getServingChannelLabel` branch returns a channel name or a `humanize()`d team label — never a raw `#slug`. `in {group.name}` subtext is `servingMode`-gated (default false) and lives in the single-channel branch only, so normal inbox rendering is untouched; last-message preview row preserved. Multi-channel skip is justified (serving mode flattens to one channel per row). `slugifyChannelLabel` confirmed fully removed — no remaining referencers.

**7. Regression check (CONFIRMED).** Plain custom-channel rename still flows through the same validated path; normal inbox rows unchanged.

Severity summary: 0 blocking, 0 high, 1 low/nit (unchanged-name re-patch). Tests added cover both the mirror and the empty-name rejection.

PR #546, opened 2026-07-02
feat(tasks): one-line team/role cells in Event Tasks grid for density

lilseyi:
## Problem

In the Event Tasks grid, the **TEAM** and **ROLE** cells rendered their chip(s) on line 1 and a full-width **+ Team** / **+ Role** (and **+ Team-level**) text add-button on a *second* line. The `ChipCell` container was already `row` + `flexWrap`, but the wide text add-pill couldn't fit next to a chip in the 160px column, so it wrapped — **doubling every row's height** and cutting how many tasks fit on screen.

## Change

- **One line in the common case.** The chips and the add control now sit inline on one row. A single team/role chip plus a **compact icon-only `+` button** (Ionicons `add`, a small muted 24px circle with `hitSlop` for a comfortable touch target) sit on one line, wrapping only when multiple chips overflow the column.
- The compact `+` opens the **exact same** `onPickTeam` / `onPickRole` anchored picker as before (it still measures its own rect via `measureAnchor`). Keeps an `accessibilityLabel` ("Add team" / "Add role").
- **Team-level (empty role)** stays a single dashed **+ Team-level** chip that opens the role picker — one line, and still reads as "no role = team-level".
- Existing chip rendering (value chips, `×` remove control, colors, dots, picker menus) is unchanged — only the cell **layout** and resulting row height changed.
- **Row density:** `GridScrollList` gains an opt-in `dense` prop (Event Tasks passes it) that lowers the row min-height (46 → 38) and cell vertical padding (9 → 6). Rows still grow past the floor when a cell wraps (`minHeight` + `alignItems: stretch`), so wrapped/multi-line content is **never clipped**, and RunSheet's grid is untouched.

## Scope

Only the TEAM/ROLE cell layout + row height. No data-model, multi-select, picker-content, or backend changes. No `Pressable` function-style `style` (static styles only, RN-Web safe).

## Verification

Mobile `tsc --noEmit`: the two changed files are clean. Remaining errors are the known repo-wide `@togather/shared` module-resolution baseline (untouched import lines) plus a pre-existing unrelated tasks test — none from this change.

**Not visually verified** — I can't drive the simulator; the layout is reasoned from the flex/wrap rules (row + wrap + center + gap; compact 24px `+` fits beside a short chip in the ~140px usable column, wraps only on overflow).

### DIFF
--- a/apps/mobile/features/scheduling/components/GridScrollList.tsx
+++ b/apps/mobile/features/scheduling/components/GridScrollList.tsx
@@ -55,6 +55,9 @@ const GRIP_W = 34;
 // with content top-aligned (not vertically centered) so multi-line cells read
 // cleanly against the shorter ones. The header is compact above it.
 const ROW_MIN_HEIGHT = 46;
+// Tighter floor for one-line-per-cell grids (Event Tasks). Rows still grow past
+// this when a cell wraps, so wrapped content is never clipped.
+const DENSE_ROW_MIN_HEIGHT = 38;
 const HEADER_MIN_HEIGHT = 38;
 
 interface Props<T> {
@@ -74,6 +77,12 @@ interface Props<T> {
   ListFooterComponent?: React.ReactElement | null;
   /** Vertical padding only — never horizontal, or rows misalign with the header. */
   contentContainerStyle?: StyleProp<ViewStyle>;
+  /**
+   * Tighter row min-height + cell padding for cells that are one line in the
+   * common case (e.g. the Event Tasks grid). Rows still grow when a cell wraps,
+   * so this never clips multi-line content — it only lowers the floor.
+   */
+  dense?: boolean;
 }
 
 export function GridScrollList<T>({
@@ -85,9 +94,11 @@ export function GridScrollList<T>({
   ListHeaderComponent,
   ListFooterComponent,
   contentContainerStyle,
+  dense = false,
 }: Props<T>) {
   const { colors } = useTheme();
   const { primaryColor } = useCommunityTheme();
+  const rowMinHeight = dense ? DENSE_ROW_MIN_HEIGHT : ROW_MIN_HEIGHT;
 
   // The card's measured inner size. Width drives the slack distribution; height
   // bounds the inner content so the drag list's vertical scroll works while it
@@ -118,7 +129,7 @@ export function GridScrollList<T>({
   const cellFrame = (i: number): ViewStyle => ({
     width: effWidths[i],
     paddingHorizontal: 10,
-    paddingVertical: 9,
+    paddingVertical: dense ? 6 : 9,
     // Top-align content (events-os `items-start`) so a 2-line cell doesn't push
     // its 1-line neighbours off-centre. The row stretches every cell to the same
     // height, so the vertical dividers still run the full row.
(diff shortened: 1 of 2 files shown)


apps/mobile/features/scheduling/components/GridScrollList.tsx
```diff
@@ -191,6 +202,7 @@ export function GridScrollList<T>({
                     style={[
                       styles.row,
                       {
+                        minHeight: rowMinHeight,
```
lilseyi, line 205: 🤖 Subagent review (codex unavailable)

NIT (non-blocking): This inline minHeight: rowMinHeight always overrides the minHeight: ROW_MIN_HEIGHT already declared in styles.row. In the non-dense case both are 46, so the StyleSheet value is now effectively dead in every render. Harmless, but dropping minHeight from styles.row would keep a single source of truth for row height.

Verified GOOD here: minHeight (not a fixed height) plus alignItems: "stretch" on styles.row means dense only lowers the floor — rows still grow when a cell wraps, so multi-line chip/how-to content is never clipped.

lilseyi (commented):
🤖 Subagent review (codex unavailable) — Event Tasks grid density change

Verdict: LGTM. Clean, well-scoped density change. No bugs found; one non-blocking NIT posted inline. All 7 review criteria verified against the code.

1. Add-button wiring — CONFIRMED GOOD. Both the compact icon-only + and the team-level dashed placeholder use the same onPress={() => measureAnchor(addRef.current, onAdd)} on an addRef-tagged Pressable, unchanged from before (EventTasksGrid.tsx:497, :510). onAdd still routes to onPickTeam/onPickRole with the measured anchor. accessibilityLabel present on both (Add team/Add role, and the emptyPlaceholder label). Anchor now measures a 24px circle instead of a wide pill, but rect-based positioning is unaffected.

2. Row growth / no clipping — CONFIRMED GOOD. styles.row keeps alignItems: "stretch" and uses minHeight (not height); dense only lowers the floor (46→38) and cell paddingVertical (9→6). Rows still grow when the team/role or how-to cell wraps, so multi-line content is never clipped (GridScrollList.tsx:132, :205).

3. dense is opt-in, EventTasksGrid only — CONFIRMED GOOD. dense defaults to false; only EventTasksGrid passes it (EventTasksGrid.tsx:370). RunSheetScreen.tsx:704 GridScrollList usage is untouched — run sheet keeps 46/9. No run-sheet regression.

4. Team-level empty case — CONFIRMED GOOD. showPlaceholder = chips.length === 0 && !!emptyPlaceholder still renders the dashed "+ Team-level" chip (styles.chipAdd/chipAddText) and opens the role picker; one line, meaning preserved (EventTasksGrid.tsx:492).

5. Multi-chip wrap — CONFIRMED GOOD. chipWrap wraps; all chips render via chips.map, with the compact + inline after the last chip. Nothing hidden/clipped; chip labels use numberOfLines={1} as before.

6. RN-Web safety — CONFIRMED GOOD. All Pressable style props are static arrays/objects (styles.chipAddCompact + theme colors), not function-style — so they apply on web. No NativeWind/className.

7. Unused styles / control regressions — CONFIRMED GOOD. Old chipAdd/chipAddText are still used by the team-level placeholder (not dead). New chipAddCompact added and used. Drag handle, duplicate/delete actions, and column widths untouched.

NIT (non-blocking, posted inline): styles.row still declares minHeight: ROW_MIN_HEIGHT, now always overridden by the inline minHeight: rowMinHeight — dead in every render. Consider removing it for a single source of truth.

Not verified at runtime (static review only): visual density in the live app and picker-menu placement against the smaller 24px anchor. Recommend a quick Playwright/simulator glance, but nothing in the code suggests a problem.

PR #547, opened 2026-07-02
feat(events): template data model + CRUD (task + run-sheet templates)

lilseyi:
## Phase 1 of event templates — backend only

Adds reusable, **per-group** task templates and run-sheet templates plus their CRUD. This is the data-model + backend phase only: **no plan linkage, no propagation, and no UI** — those land in later phases.

### What's here
- **Schema** — 4 new tables:
  - `eventTaskTemplates` / `eventTaskTemplateItems` — the template mirror of `eventTasks` minus `planId`, using only the multi-assign array role model (`teamIds` / `roleIds`, no legacy single-value columns) and no completion state.
  - `runSheetTemplates` / `runSheetTemplateItems` — the template mirror of `eventItems` minus `planId` (segment + sequence ordering, notes, role assignments, library `songId` join).
  - Note: the task-template tables are prefixed `event*` because a **separate** `taskTemplates` table already exists in the schema (different feature) — using the bare name collided.
- **Backend CRUD** — `functions/scheduling/taskTemplates.ts` and `functions/scheduling/runSheetTemplates.ts`, mirroring the `eventTasks` / `eventItems` logic (sort/sequence ordering, resequencing, role-belongs-to-team + cross-community-song validation, hydration). Auth: `requireGroupScheduler` on writes, `requireGroupMember` on reads. Exported via `functions/scheduling/index.ts`.
- **Mobile API refs** — `apps/mobile/features/scheduling/api/eventTemplates.ts`, typed `makeFunctionReference` refs (mirrors `crossTeamChannels.ts`) so later phases can call the CRUD.
- **Tests** — `__tests__/scheduling/eventTemplates.test.ts` (13 tests): CRUD, reorder, delete-template cascade, itemCounts, role/song validation, and non-leader auth rejection.

### Verification
- `apps/convex` `tsc --noEmit`: clean (only the pre-existing `@togather/shared/config` baseline errors remain, all in untouched files).
- Scheduling test suite (incl. the new file): **301 passed / 22 files**; new file **13 passed**.
- `npx convex codegen` is not runnable here (no deployment); the generated `api.d.ts` was hand-patched to register the two new modules, and `api.js` resolves via `anyApi` at runtime so tests pass. CI regenerates on deploy.

### DIFF
--- a/apps/convex/_generated/api.d.ts
+++ b/apps/convex/_generated/api.d.ts
@@ -149,6 +149,8 @@ import type * as functions_scheduling_events from "../functions/scheduling/event
 import type * as functions_scheduling_index from "../functions/scheduling/index.js";
 import type * as functions_scheduling_mySchedule from "../functions/scheduling/mySchedule.js";
 import type * as functions_scheduling_people from "../functions/scheduling/people.js";
+import type * as functions_scheduling_runSheetTemplates from "../functions/scheduling/runSheetTemplates.js";
+import type * as functions_scheduling_taskTemplates from "../functions/scheduling/taskTemplates.js";
 import type * as functions_scheduling_permissions from "../functions/scheduling/permissions.js";
 import type * as functions_scheduling_presence from "../functions/scheduling/presence.js";
 import type * as functions_scheduling_publicAvailability from "../functions/scheduling/publicAvailability.js";
@@ -371,6 +373,8 @@ declare const fullApi: ApiFromModules<{
   "functions/scheduling/index": typeof functions_scheduling_index;
   "functions/scheduling/mySchedule": typeof functions_scheduling_mySchedule;
   "functions/scheduling/people": typeof functions_scheduling_people;
+  "functions/scheduling/runSheetTemplates": typeof functions_scheduling_runSheetTemplates;
+  "functions/scheduling/taskTemplates": typeof functions_scheduling_taskTemplates;
   "functions/scheduling/permissions": typeof functions_scheduling_permissions;
   "functions/scheduling/presence": typeof functions_scheduling_presence;
   "functions/scheduling/publicAvailability": typeof functions_scheduling_publicAvailability;
--- a/apps/convex/schema.ts
+++ b/apps/convex/schema.ts
@@ -3027,6 +3027,117 @@ export default defineSchema({
     .index("by_plan", ["planId"]),
 
   /**
+   * A reusable, per-GROUP event-task template — a named, saved checklist of
+   * `eventTaskTemplateItems` a leader can keep for a location and later apply to
+   * a plan's `eventTasks`. Named `eventTaskTemplates` (not `taskTemplates`, which
+   * is a separate feature) to avoid a table collision. Scoped to one campus group
+   * (mirrors how cross-team channels / teams are group-scoped); `communityId` is
+   * denormalized for the community read gate. Phase 1 stores the template only —
+   * there is no plan linkage or propagation yet.
+   */
+  eventTaskTemplates: defineTable({
+    groupId: v.id("groups"),
+    communityId: v.id("communities"),
+    name: v.string(),
+    createdById: v.id("users"),
+    createdAt: v.number(),
+    updatedAt: v.number(),
+  }).index("by_group", ["groupId"]),
+
+  /**
+   * A single item on an `eventTaskTemplates` template — the template mirror of
+   * `eventTasks` MINUS `planId` (keyed by `templateId` instead) and using ONLY
+   * the multi-assign array model (`teamIds` >= 0, `roleIds` empty => a
+   * team-level task). No legacy single-value `teamId` / `roleId` columns, and no
+   * completion records (templates are never "done" — they seed a plan's tasks).
+   * `sortOrder` orders items WITHIN a segment.
+   */
+  eventTaskTemplateItems: defineTable({
+    templateId: v.id("eventTaskTemplates"),
+    communityId: v.id("communities"),
+    teamIds: v.array(v.id("teams")),
+    roleIds: v.array(v.id("teamRoles")),
+    segment: v.union(
+      v.literal("before"),
+      v.literal("during"),
+      v.literal("after"),
+    ),
+    title: v.string(),
+    howToType: v.union(
+      v.literal("none"),
+      v.literal("text"),
+      v.literal("link"),
+      v.literal("media"),
+      v.literal("doc"),
+    ),
+    howToText: v.optional(v.string()),
+    howToUrl: v.optional(v.string()),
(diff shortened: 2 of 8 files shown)


apps/convex/schema.ts
```diff
@@ -3026,6 +3026,117 @@ export default defineSchema({
...
+      }),
+    ),
+    songId: v.optional(v.id("songs")),
+    createdById: v.id("users"),
+    createdAt: v.number(),
+    updatedAt: v.number(),
+  })
+    .index("by_template", ["templateId"])
+    // Scan items referencing a song so deleteSong can null them out (ADR-027).
+    .index("by_song", ["songId"]),
```
lilseyi, line 3138: **MEDIUM — CONFIRMED — data integrity (dangling reference).** This `by_song` index is added with the stated purpose "so deleteSong can null them out (ADR-027)", but `deleteSong` (`apps/convex/functions/scheduling/songs.ts:260-268`) only scans `eventItems.by_song` and nulls those — it never scans `runSheetTemplateItems`. So deleting a library song leaves a dangling `songId` on any run-sheet **template** item that referenced it. It is not a crash (`getHydratedSongForJoin` returns `null` for a missing id), but it violates the invariant this index was created to maintain, and the comment here promises behavior the PR does not implement. Fix: in `deleteSong`, also collect `runSheetTemplateItems` via `by_song` and patch `songId: undefined`.

apps/convex/_generated/api.d.ts
```diff
@@ -149,6 +149,8 @@ import type * as functions_scheduling_events from "../functions/scheduling/event
 import type * as functions_scheduling_index from "../functions/scheduling/index.js";
 import type * as functions_scheduling_mySchedule from "../functions/scheduling/mySchedule.js";
 import type * as functions_scheduling_people from "../functions/scheduling/people.js";
+import type * as functions_scheduling_runSheetTemplates from "../functions/scheduling/runSheetTemplates.js";
+import type * as functions_scheduling_taskTemplates from "../functions/scheduling/taskTemplates.js";
```
lilseyi, line 153: **LOW — CONFIRMED — cosmetic, CI-safe.** The hand-patched entries are inserted here (between `people` and `permissions`), but `convex codegen` emits this list in case-sensitive ASCII order — it would place `runSheetTemplates` next to `roster`/`serving` and `taskTemplates` next to `starterRoles`/`teamChannelSync`, not here. **CI risk assessment: LOW / not a blocker.** All three added module names are valid and every `typeof` resolves, so `npx convex typecheck` (ci.yml) passes regardless of ordering; there is no "generated files must be clean" git-diff gate in the workflow, and `convex typecheck` regenerates this file before checking. The only real consequence is that the next person to run codegen will get a noisy reorder diff. Optional: rerun codegen and commit the sorted result to avoid that churn.

lilseyi (commented):
🤖 Subagent review (codex unavailable) — Phase 1 event-templates data model + CRUD

Adversarial review against the 8 requested axes. Overall this is a clean, faithful mirror of `eventTasks.ts` / `eventItems.ts`. Two findings posted inline; no blockers.

**Findings (severe-first):**

1. MEDIUM — CONFIRMED — `deleteSong` does not null `runSheetTemplateItems.songId` (`songs.ts:260-268`; index at `schema.ts:3137-3138`). The new `by_song` index is added with the explicit comment "so deleteSong can null them out (ADR-027)", but `deleteSong` only scans `eventItems`, never `runSheetTemplateItems`. Deleting a song leaves a dangling `songId` on template items. Not a crash (`getHydratedSongForJoin` returns null), but the invariant the index promises is not wired up. See inline.

2. LOW — CONFIRMED — `api.d.ts` hand-patch is mis-ordered vs `convex codegen` (lines 152-153). CI RISK = LOW / not a blocker: all module names are valid so `npx convex typecheck` passes regardless of order, and there is no generated-files-clean gate in ci.yml. Only consequence is a noisy reorder diff on the next codegen. See inline.

**Verified CLEAN (per requested axis):**

- NAMING COLLISION (1): CONFIRMED clean. Existing CRM `taskTemplates` table (`schema.ts:1198`) + `functions/taskTemplates/index` are distinct from the new `eventTaskTemplates`/`eventTaskTemplateItems`. No ambiguous refs to the new names outside the new files. The event*/runSheet* asymmetry is cosmetic.
- SCHEMA (2): CONFIRMED. 4 all-new tables, additive-safe. Item tables use the array role model (`teamIds`/`roleIds`) only, no legacy `teamId`/`roleId`. Indexes `by_group`, `by_template`, `by_template_segment`, `by_song` all present and used by the queries.
- AUTH (3): CONFIRMED — the important one. Every mutation resolves the group FROM the template (`requireTaskTemplateScheduler` / `requireRunSheetTemplateScheduler` load the template, then authorize against `template.groupId`), never a caller-supplied group — a caller cannot edit another group's template. Writes gated by `requireGroupScheduler`, reads by `requireGroupMember`. Both delete mutations cascade all child items before deleting the template (no orphans).
- VALIDATION (4): CONFIRMED. Name rule (1-50, ≥1 alnum) applied on create + rename. Item mutations validate each team belongs to the template group and each role belongs to one of the item's teams (mirrors `eventTasks`); run-sheet assignments validated against the template group; segment/type validated. `reorderRunSheetTemplateItems` rejects stale/partial/duplicate sets and resequences per-segment. NOTE: `reorderTaskTemplateItems` does NOT reject partial/stale sets — it silently skips foreign ids — but that is a FAITHFUL mirror of `eventTasks.reorderTasks` (the asymmetry is inherited from the two different sources, not a regression).
- CORRECTNESS vs source (5): CONFIRMED faithful. sortOrder/sequence computation, duplicate-then-resequence (`duplicateRunSheetTemplateItem`), and hydration join shapes match `eventItems`/`eventTasks`. Completion machinery correctly dropped (templates are never completed). Cross-community song link rejected in `updateRunSheetTemplateItem`.
- MOBILE REFS (7): CONFIRMED. `eventTemplates.ts` ref strings (`functions/scheduling/taskTemplates:fn`) match the `crossTeamChannels` pattern exactly, every referenced export exists, and arg/return type assertions match the backend validators.
- TESTS (8): 13 tests cover create/rename/add/list-ordered/reorder, update/delete, role-in-team rejection, delete-cascade, itemCount, non-leader auth rejection, invalid name, cross-community song. Minor gaps (non-blocking): no negative test for `reorderRunSheetTemplateItems` stale/partial rejection, and no test for the deleteSong→template-item nulling (which is finding #1, currently broken).

**Verdict:** Approve-with-nits. Recommend fixing #1 (wire `deleteSong` to null `runSheetTemplateItems`, ideally + a test) before or shortly after merge; #2 is cosmetic. This is a review comment only — not an approval.

PR #549, opened 2026-07-02
feat(events): templates library UI + item editors (Phase 2)

lilseyi:
## Event templates — Phase 2 (frontend)

Builds the **frontend** for the per-location (per-GROUP) event templates
feature on top of the Phase 1 model + CRUD (no backend changes). Leaders can
create/rename/delete **task templates** and **run-sheet templates** and edit
each template's items — reusing the existing Event Tasks grid and run-sheet
editor rather than duplicating them.

### What's included
- **Templates library** (`TemplatesLibraryScreen`, route
  `/rostering/[group_id]/templates`): two sections (Task templates / Run-sheet
  templates), each listing templates with item counts, a "+ New …" row, and
  per-row rename/delete (⋯ / long-press). Leader-gated.
- **Task-template editor** (route `.../templates/task/[template_id]`): renders
  the plan **`EventTasksGrid`** wired to the task-template-item mutations, with
  the same phase/segment + multi-select team/role editing and a tap-to-rename
  header.
- **Run-sheet-template editor** (route `.../templates/runsheet/[template_id]`):
  reuses the run sheet editor's row/cell building blocks + shared
  `GridScrollList`, wired to the run-sheet-template-item mutations
  (add/edit/delete/duplicate/reorder, When/Dur/Owner-Role/Notes/Song). Clock
  times are omitted (templates hold durations only).
- **Entry point**: a "Templates" item in the roster grid's ⋯ overflow (beside
  Cross-team).

### How grids were reused vs extracted
- **Event Tasks grid — reused verbatim.** `EventTasksGrid` never reads `planId`,
  so template items are adapted to its `PlanTask` shape and its callbacks wired
  to the template mutations. The id-brand mismatch (template item vs plan task)
  is bridged with localized casts at the container boundary — **zero change to
  the shared grid**, so plan Event Tasks is untouched.
- **Run sheet editor — extracted shared parts (DRY).** The presentational,
  plan-agnostic row/cell editors (`WhenPill`, `DurationCell`, `NotesEditor`,
  Who/Notes/Song modal bodies, `AddButton`, segment types/`ItemPatch`) moved
  from `RunSheetScreen` into a new `RunSheetItemEditors` module used by **both**
  the plan run sheet and the template editor. `RunSheetScreen` now imports them
  — same props, no behavior change; plan-specific clock times / service ranges
  stay in the screen.

### Regression risk
- Main risk is the `RunSheetScreen` extraction. The move is mechanical and
  `tsc` is clean; the plan run sheet uses the extracted components with
  identical props. **Not runtime-verified** (no simulator) — worth a quick
  manual smoke of the plan run sheet.

### Explicitly deferred to Phase 3/4
Plan↔template linkage, propagation, save-to-template, and plan-side pickers —
this phase is library + item editors only.

### DIFF
--- a/apps/mobile/features/scheduling/components/RunSheetItemEditors.tsx
+++ b/apps/mobile/features/scheduling/components/RunSheetItemEditors.tsx
@@ -1,0 +1,601 @@
+/**
+ * RunSheetItemEditors
+ *
+ * The presentational, plan-agnostic building blocks of the run sheet editor —
+ * extracted from RunSheetScreen so BOTH the per-plan run sheet
+ * (`RunSheetScreen`) and the per-group run-sheet TEMPLATE editor
+ * (`RunSheetTemplateEditorScreen`) render identical row/cell UI (ADR-026 /
+ * event templates Phase 2).
+ *
+ * These components read only from a minimal `RunSheetItemLike` shape and write
+ * through an `ItemPatch` callback, so they are unaware of whether the row is a
+ * plan `eventItems` row or a `runSheetTemplateItems` row. Anything genuinely
+ * plan-specific (clock times, service ranges, event roster) stays in the
+ * containers.
+ */
+import React, { useState } from "react";
+import {
+  View,
+  Text,
+  StyleSheet,
+  Pressable,
+  Platform,
+  type TextStyle,
+} from "react-native";
+import { Ionicons } from "@expo/vector-icons";
+import { useTheme } from "@hooks/useTheme";
+import type { Id } from "@services/api/convex";
+import { DEFAULT_ROLE_COLOR } from "../utils/format";
+import { InlineText } from "./InlineText";
+import { OptionTag } from "./GridScrollList";
+import { measureAnchor, type AnchorRect } from "./AnchoredMenu";
+import { SongPicker } from "./SongPicker";
+import type { Song } from "@features/songs/types";
+
+/** When an item happens relative to the event's service times. */
+export type Segment = "before" | "during" | "after";
+
+export const SEGMENT_OPTIONS: Array<{
+  key: Segment;
+  label: string;
+  short: string;
+}> = [
+  { key: "before", label: "Before event", short: "Before" },
+  { key: "during", label: "During event", short: "During" },
+  { key: "after", label: "After event", short: "After" },
+];
+
+export type ItemAssignment = {
+  roleId: Id<"teamRoles">;
+  roleName: string;
+  roleColor: string | null;
+};
+
+export type RoleOption = {
+  roleId: Id<"teamRoles">;
+  roleName: string;
+  roleColor?: string;
+  people: string[];
+};
+
+/** Item field patch shape sent to the container's update mutation. */
+export type ItemPatch = {
+  type?: string;
+  title?: string;
+  segment?: Segment;
+  durationSec?: number;
+  description?: string;
+  notes?: Array<{ category: string; content: string }>;
+  assignments?: Array<{ roleId: Id<"teamRoles"> }>;
+  songDetails?: { key?: string; bpm?: number };
+  // Link / unlink a library song. `null` clears the link (ADR-027).
+  songId?: Id<"songs"> | null;
+};
+
+/**
+ * The minimal read surface the row editors need. Both `RunSheetItem` (plan) and
+ * a mapped `runSheetTemplateItems` row satisfy this structurally.
+ */
+export type RunSheetItemLike = {
+  segment: string;
+  description: string | null;
+  durationSec: number;
+  notes: Array<{ category: string; content: string }>;
+  songDetails: { key?: string; bpm?: number; author?: string } | null;
+  songId: Id<"songs"> | null;
+  song: Song | null;
+  assignments: ItemAssignment[];
+};
+
+/**
+ * The "When" phase pill. Measures its own window rect on press so the parent can
+ * anchor the dropdown next to it (the pill can't hold the menu itself — the
+ * table card clips overflow).
+ */
+export function WhenPill({
+  label,
+  colors,
+  primaryColor,
+  onOpen,
+}: {
+  label: string;
+  colors: ReturnType<typeof useTheme>["colors"];
+  primaryColor: string;
+  onOpen: (anchor: AnchorRect) => void;
+}) {
+  const ref = React.useRef<View>(null);
+  return (
+    <Pressable
+      ref={ref}
+      onPress={() => measureAnchor(ref.current, onOpen)}
+      style={styles.tagPressable}
+      accessibilityRole="button"
+      accessibilityLabel={`When: ${label}. Tap to change.`}
+    >
+      <OptionTag
+        label={label}
+        colors={colors}
+        primaryColor={primaryColor}
+        tinted
+        chevron
+      />
+    </Pressable>
+  );
+}
+
+export function AddButton({
+  label,
+  icon,
+  onPress,
+  primaryColor,
+  colors,
+}: {
+  label: string;
+  icon: keyof typeof Ionicons.glyphMap;
+  onPress: () => void;
+  primaryColor: string;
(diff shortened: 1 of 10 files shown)


apps/mobile/features/scheduling/components/RunSheetItemEditors.tsx
```diff
@@ -0,0 +1,601 @@
...
+    setLinked(next);
+    onPatch({
+      assignments: [...next].map((id) => ({ roleId: id as Id<"teamRoles"> })),
+    });
+  };
+
+  if (roleOptions.length === 0) {
+    return (
+      <Text style={[styles.emptyText, { color: colors.textSecondary }]}>
+        No roles are defined for this group yet.
```
lilseyi: 🤖 Subagent review (codex unavailable) — LOW / CONFIRMED

Copy regression in the **plan** run sheet: this empty-state string was `"No roles are defined for this **event** yet."` before the extraction and is now `"...for this **group** yet."`. Since this shared component also backs `RunSheetScreen` (the per-plan run sheet, where roles are event-scoped), the plan editor now shows the less-accurate "group" wording. Cosmetic only — flagging because it is the one observable behavior change in an otherwise byte-faithful extraction.

apps/mobile/features/scheduling/components/RunSheetTemplateEditorScreen.tsx
```diff
@@ -0,0 +1,909 @@
...
+    <View
+      style={[
+        styles.container,
+        { paddingTop: insets.top, backgroundColor: colors.surface },
+      ]}
+    >
+      {renderHeaderBar()}
+
+      {/* Load every team's roles so the Who picker has group-wide options. */}
+      {referencedTeamIds.map((tid) => (
```
lilseyi, line 643: 🤖 Subagent review — LOW / CONFIRMED (perf nit, non-blocking)

The Who picker is populated by mounting one `RoleLoader` (= one `roles.listRoles` query) for **every** team in the group (`referencedTeamIds` = all `teamsData`). For a group with many teams this fans out to N live queries just to fill the roles multiselect. The plan `RunSheetScreen` avoided this by reading roles off the single `event.roles` payload. Acceptable for now (Convex dedupes identical queries), but worth a follow-up if groups routinely have many teams.

apps/mobile/features/scheduling/components/RunSheetTemplateEditorScreen.tsx
```diff
@@ -0,0 +1,909 @@
...
+      <CustomModal
+        visible={whoLive !== null}
+        onClose={() => setWhoItem(null)}
+        title="Who's involved"
+      >
+        {whoLive ? (
+          <WhoModalBody
+            key={whoLive._id}
+            item={toItemLike(whoLive)}
+            roleOptions={roleOptions}
```
lilseyi, line 691: 🤖 Subagent review — LOW / PLAUSIBLE

Transient empty state: `roleOptions` is derived from `rolesByTeam`, which fills asynchronously as each `RoleLoader` resolves. If the user opens the Who modal before those queries return, `WhoModalBody` renders "No roles are defined for this group yet." even when roles exist. Self-corrects on the next render once roles load; minor UX flicker.

apps/mobile/features/scheduling/components/TaskTemplateEditorScreen.tsx
```diff
@@ -0,0 +1,658 @@
...
+      void updateItem({ itemId: asItemId(taskId), ...patch }).catch((e) =>
+        notifyError("Couldn't save", errMsg(e)),
+      );
+    },
+    [updateItem],
+  );
+
+  const handleAdd = useCallback(
+    async (segment: Segment) => {
+      const teamId = teams[0]?._id;
```
lilseyi, line 230: 🤖 Subagent review — LOW / PLAUSIBLE

`handleAdd` uses `teams[0]?._id` as the default team and, when absent, shows "Add a team first". `teamsData` loads independently of `items`, and the empty-state Add buttons render as soon as `items` resolves — so if a leader taps Add before `listTeams` returns, they can get the "Add a team first" error even though teams exist. Edge/race case; consider gating the Add buttons on `teamsData !== undefined`.

lilseyi (commented):
🤖 Subagent review (codex unavailable) — Phase 2 templates library UI + RunSheetItemEditors extraction

## Verdict: LGTM / non-blocking. No correctness bugs found. The RunSheetScreen extraction is behavior-preserving and the plan run sheet is safe to ship.

### 1. RunSheetScreen extraction — SAFE (this was the #1 risk)
Diffed every moved component against `RunSheetItemEditors.tsx`. The move is faithful (only line-wrapping/formatting differs):
- **No lost closures.** Each moved piece (`WhenPill`, `AddButton`, `DurationCell`/`parseDuration`, `WhoModalBody`, `NotesModalBody`, `SongModalBody`, `NotesEditor`, `webNoOutline`, `SEGMENT_OPTIONS`, `Segment`/`ItemAssignment`/`RoleOption`/`ItemPatch`/`RunSheetItemLike` types) was already prop-driven. Nothing captured plan-scope state (`clockTimes`, `earliestStart`, `event`, `roleOptions`, `duringTotalSec`) — those stay in `RunSheetScreen` and are still passed in as props. Duration parsing, notes editing, who/song modals, add button, and segment handling are unchanged.
- **Call sites pass identical props** (`WhenPill`, `DurationCell`, and the three modal bodies verified at RunSheetScreen.tsx:533/559/710/726/741).
- **Dead imports/styles confirmed gone:** dropped `type TextStyle`, `measureAnchor`, and `SongPicker` imports are no longer referenced in `RunSheetScreen` (grep-clean). Every removed style key was one of the moved ones; no dangling `styles.X` remains and all retained styles are still used.
- One cosmetic change: shared empty-state copy `"...for this event yet."` → `"...for this group yet."` now also shows in the plan run sheet (inline note). Non-blocking.

### 2. Task-template reuse / cast helpers — SAFE
`asTaskId`/`asItemId` are localized to the adapter boundary, used only to build the display `PlanTask` shape and to route grid callbacks to the `*TaskTemplateItem*` mutations. They are never handed to a plan (`eventTasks`) mutation. Add/update/delete/reorder/duplicate all target the template refs; arg shapes match `api/eventTemplates.ts` (howTo fields are `string | undefined`, so duplicate carries no `null`). Teams from `listTeams`, roles lazily per-team — correct group-scoped sources; `people` empty is expected.

### 3. Run-sheet template editor — SAFE
Add/edit/delete/duplicate/reorder/rename wired to the `runSheetTemplates` refs; song picker (`songId`), notes, durations, When phase all work. Time/clock-time column is correctly omitted (plan-only concept) — no leakage.

### 4. Library screen — SAFE
Create/rename/delete routed to the right task vs run-sheet mutations; item counts from `itemCount` on the list summaries; leader gate mirrors the other rostering leader screens (`getById.userRole` + `is_admin`); nav passes the right template ids; delete confirms cross-platform.

### 5. Conventions — OK
No NativeWind/className. All `Pressable`s use static `style` arrays (no RN-Web function-style pitfall). Hooks declared before every early return in all four new screens. `CustomModal` usage consistent. The 3 new `index.ts` exports do not collide; module-local `RoleLoader`/`RenameBody`/`NameModalBody` duplicates across files are fine (not exported).

### 6. RosterGridScreen overflow — SAFE
`onTemplates` threaded cleanly through `OverflowMenu`, reuses existing `menuRow`/`occupantName` styles; no impact on sibling menu items.

### 7. Crash surface — OK
Empty template, no items, and missing-name are all guarded (empty states + `canSave`); `setRolesForTeam` length-guards its `.every` compare.

### Low / non-blocking notes (inline)
- Plan run sheet empty-roles copy "event"→"group".
- Who picker fans out one `listRoles` query per group team.
- Transient "no roles" flash before RoleLoaders resolve.
- `handleAdd` "Add a team first" race if `listTeams` is still loading.

I did not run typecheck (branch is checked out in a sibling worktree); review is static + type-shape cross-check against the Phase 1 refs.

PR #550, opened 2026-07-02
feat(events): plan↔template linkage, forward propagation, save-as-template (Phase 3 backend)

lilseyi:
## Phase 3 (backend): plan ↔ template linkage, forward propagation, save-as-template

Backend-only. All UI (template pickers, link/switch/save/revert affordances) is deferred to **Phase 4**.

### The model
A plan can LINK to a task template and (independently) a run-sheet template. A linked plan still **owns** its own `eventTasks` / `eventItems` rows — they are **materialized** from the template's items, each carrying a `sourceTemplateItemId` back-pointer. Template edits propagate **forward** to **future** linked plans' still-synced rows; **past plans (eventDate < now) are frozen** and never touched. Per-row overrides + save-to-template are supported.

### Row semantics (both `eventTasks` and `eventItems`)
- `sourceTemplateItemId` set + `templateDetached` falsy → **SYNCED**: propagation updates/deletes it to match the template.
- `sourceTemplateItemId` set + `templateDetached` true → **OVERRIDDEN**: propagation leaves it alone.
- `sourceTemplateItemId` unset → **LOCAL** (plan-owned addition): propagation ignores it.
- template item id in the plan's `detached…ItemIds` → user removed it locally: propagation must not re-add it.

### Forward propagation (`propagateTaskTemplate` / `propagateRunSheetTemplate`)
Called at the end of **every** template ITEM mutation (add/update/delete/duplicate/reorder). For each plan linked to the template with `eventDate >= Date.now()`:
- template item exists + SYNCED row → patch the row to match (all fields, template ordering).
- template item exists + OVERRIDDEN row → leave it.
- template item exists + no row + id not in the plan's detached set → insert a new synced row.
- template item deleted + SYNCED row → delete the row (task completions cascaded via `cascadeTaskCompletions`).
- On a task's role↔team-level flip during a synced patch, stale completions are cleared.

### Freeze-on-date
`setPlan*Template` rejects a plan whose `eventDate < Date.now()`; propagation only considers plans with `eventDate >= Date.now()`.

### Override marking (in existing plan mutations, surgical / guarded on `sourceTemplateItemId`)
- `updateTask` / `updateItem`: editing a template-sourced row sets `templateDetached: true`.
- `deleteTask` / `deleteItem`: records the removed item's `sourceTemplateItemId` in the plan's `detached…ItemIds` so propagation won't re-add it.
- `reorderTasks` / `reorderItems`: detach only rows whose position **actually changes**.
- `createTask` / `createItem`: new rows are plain local (no source tag) — already ignored by propagation, no change needed.
- **No behavior change for plans with no template link** (all paths are guarded on `sourceTemplateItemId`).

### Link / switch / unlink (`setPlanTaskTemplate` / `setPlanRunSheetTemplate`)
- `templateId: null` → unlink: strip source tags from rows (become local), clear pointer + detached set.
- `templateId` set → link/switch: remove currently-synced rows, instantiate the new template's items as fresh synced rows, reset the detached set. `carryover`: `"discard"` drops overridden + local rows; `"copy"` keeps them as plain local rows alongside the new template rows. Same-group enforced. Roster/completions untouched.

### Save-as-template (`saveTaskTemplateFromPlan` / `saveRunSheetTemplateFromPlan`)
- `{kind:"new", name}` → create a template in the plan's group from the plan's current effective list, then link the plan to it (rows become synced).
- `{kind:"existing", templateId, strategy:"replace"|"merge"}` → replace clears + repopulates and relinks the source plan clean; merge appends the plan's list. Both propagate forward to the template's other future plans.

### Revert (`revertPlan*TemplateEdits`)
Drops all of the plan's rows (overridden + local, cascading completions), clears the detached set, and re-instantiates from the linked template. No-op if unlinked.

### Read contract for Phase 4 UI
`getPlanTemplateState({token, planId})` (group-member auth, with `returns` validator) → `{ taskTemplateId, taskTemplateName, hasTaskTemplateEdits, runSheetTemplateId, runSheetTemplateName, hasRunSheetTemplateEdits, isPast }`.

### duplicateEvent
Carries the template pointers, detached sets, and per-row `sourceTemplateItemId` / `templateDetached` so a duplicated upcoming plan stays linked.

### Tests
`apps/convex/__tests__/scheduling/planTemplates.test.ts` — 18 convex-test cases. Full scheduling suite: **321/321 green**. `apps/convex` `tsc --noEmit` clean.

### Notes / deferred
- All Phase 4 UI (pickers, banners, revert/save affordances) deferred.
- `_generated/api.d.ts` was hand-edited to register `planTemplates` (codegen not runnable offline; CI regenerates).
- Skipped the optional `by_plan_sourceTemplateItem` indexes — propagation scans a plan's rows via `by_plan` and maps in memory (plans have few rows).
- `deleteTaskTemplate` does not clear pointers on linked plans (their rows freeze at last-synced state); can be handled in a follow-up if desired.


apps/convex/functions/scheduling/eventTasks.ts
```diff
@@ -390,6 +423,13 @@ export const updateTask = mutation({
       patch.howToMediaPath = args.howToMediaPath;
     if (args.howToDoc !== undefined) patch.howToDoc = args.howToDoc;
 
+    // Event-templates linkage (Phase 3): a local edit to a template-sourced task
+    // detaches it so forward propagation stops overwriting the user's change.
```
lilseyi: 🤖 Subagent review (codex unavailable)

**[LOW / by-design — CONFIRMED]** `updateTask` detaches a synced row on *every* call, not only when a field actually changes. The block sets `templateDetached = true` whenever `task.sourceTemplateItemId && !task.templateDetached`, regardless of whether `patch` contains a real delta. If the Phase-4 editor writes a field back unchanged (autosave / focus-blur / re-saving the same `howToDoc`), a linked row silently stops receiving template updates. Consider gating on an actual field change. Note: non-linked plans are unaffected (guard is correct).

apps/convex/functions/scheduling/eventItems.ts
```diff
@@ -338,6 +338,13 @@ export const updateItem = mutation({
       }
     }
 
+    // Event-templates linkage (Phase 3): a local edit to a template-sourced item
+    // detaches it so forward propagation stops overwriting the user's change.
```
lilseyi: 🤖 Subagent review (codex unavailable)

**[LOW / by-design — CONFIRMED]** Same as the `updateTask` sibling: `updateItem` detaches a synced run-sheet row on *any* call, even a no-op edit. A Phase-4 UI that re-saves unchanged values would silently detach linked rows from template propagation. Consider detaching only when the patch actually changes a field. Non-linked plans unaffected.

apps/convex/functions/scheduling/planTemplates.ts
```diff
@@ -0,0 +1,962 @@
...
+      }
+      await ctx.db.patch(plan._id, {
+        taskTemplateId: templateId,
+        detachedTaskTemplateItemIds: [],
+        updatedAt: now,
+      });
+      await ctx.db.patch(templateId, { updatedAt: now });
+      // The source plan is now fully synced (no-op for it); this updates the
+      // template's OTHER future plans.
+      await propagateTaskTemplate(ctx, templateId, userId);
```
lilseyi, line 902: 🤖 Subagent review (codex unavailable)

**[MEDIUM / PLAUSIBLE]** `replace` deletes every template item and recreates them with **new ids**, then propagates. For the *source* plan this is a no-op. But for every **other** future plan linked to this template, propagation (a) sees no `bySource` match for the new ids → inserts a fresh synced row for each, and (b) in the delete loop finds each old synced row’s `sourceTemplateItemId` no longer in `itemIds` → deletes it, **cascading/destroying its completions**. So a save-as-replace from plan A silently wipes completion state on sibling future plans. Worse, a sibling plan’s **overridden** rows are skipped by the delete loop (kept, stale source id) *and* the fresh synced copies are inserted → **duplicated content** on that sibling. No UI yet (Phase 4), but this needs a guard/warning before it ships.

apps/convex/functions/scheduling/planTemplates.ts
```diff
@@ -0,0 +1,962 @@
...
+    // duplicate rows (they came from this plan). Merge does not relink.
+    if (plan.taskTemplateId === templateId) {
+      await ctx.db.patch(plan._id, {
+        detachedTaskTemplateItemIds: [
+          ...(plan.detachedTaskTemplateItemIds ?? []),
+          ...newItemIds,
+        ],
+      });
+    }
+    await propagateTaskTemplate(ctx, templateId, userId);
```
lilseyi, line 941: 🤖 Subagent review (codex unavailable)

**[LOW-MEDIUM / PLAUSIBLE]** The detached-set trick correctly prevents the appended copies from being re-added to the *source* plan. But `merge` appends the source plan's **entire current task list** — including rows that are already SYNCED to this same template — as brand-new template items. Those new items then propagate to every *other* future plan, which inserts them as duplicates (the other plans already have the originals). Merging a plan that is itself already linked to the target template therefore duplicates content across siblings. Likely an odd/rare operation, but worth either rejecting `merge` when `plan.taskTemplateId === templateId`, or merging only LOCAL (non-sourced) rows.

apps/convex/functions/scheduling/planTemplates.ts
```diff
@@ -0,0 +1,962 @@
...
+  const futurePlans = plans.filter((p) => p.eventDate >= now);
+  if (futurePlans.length === 0) return;
+
+  const items = await ctx.db
+    .query("eventTaskTemplateItems")
+    .withIndex("by_template", (q) => q.eq("templateId", templateId))
+    .collect();
+  const itemIds = new Set(items.map((i) => i._id as string));
+
+  for (const plan of futurePlans) {
```
lilseyi, line 271: 🤖 Subagent review (codex unavailable)

**[LOW / SCALE — PLAUSIBLE]** Propagation is O(futurePlans × rows) and runs **inside** the template-item mutation transaction, on *every* add/update/delete/duplicate/reorder of a template item. A widely-used template with many future linked plans (each with a full task list) turns a single item edit into a large fan-out of reads+patches+inserts in one transaction, which can approach Convex per-mutation read/write limits and add latency to routine template edits. Fine at current scale; flagging as a ceiling to watch (and a reason a future migration/backfill should batch rather than reuse this path).

Also note (separate, INFO): the delete loop at L308-313 intentionally skips OVERRIDDEN rows, so an overridden row whose template item is later deleted keeps a **dangling** `sourceTemplateItemId` forever. Harmless (always skipped by both loops) but a latent inconsistency that compounds the replace-duplication issue above.

lilseyi (commented):
🤖 Subagent review (codex unavailable) — Phase 3 backend: plan↔template linkage + forward propagation

Adversarial review focused on the two highest-risk surfaces: (1) whether non-linked plans get byte-identical behavior, and (2) propagation correctness. Traced all 6 edited existing-mutation branches, the propagation reconcilers, link/switch/unlink, save-as-template, revert, duplicateEvent, and schema.

## ✅ #1 VERDICT — NON-LINKED PLANS ARE UNAFFECTED (CONFIRMED SAFE)
Every added branch in `eventTasks.ts` / `eventItems.ts` is guarded on `sourceTemplateItemId` being set:
- `updateTask`/`updateItem` — detach block only runs when `task.sourceTemplateItemId && !task.templateDetached`.
- `deleteTask`/`deleteItem` — detached-set patch only when `sourceTemplateItemId` is set.
- `reorderTasks`/`reorderItems` — detach only when `sourceTemplateItemId && !templateDetached && position actually changed`.
- Return shapes unchanged (`{deleted}`, `{deletedCompletions}`, `{itemId}`, `{reordered}`).
- `cascadeTaskCompletions` extraction is **behavior-identical** to the old inline `deleteTask` cascade — same 3 tables (`eventTaskCompletions`, `sharedTaskCompletions`, `howToDocChecks`); `deletedCompletions` still = per-user completion count. `personalServingTasks` is a separate table not keyed to `eventTasks.taskId`, so (correctly) not part of this cascade — matches prior behavior.

A plan with no template link never has `sourceTemplateItemId` on any row, so all new code is a no-op for it. No unguarded patch, no completion-cascade change for non-template rows, no return-shape drift. The whole serving/run-sheet system is safe for the un-migrated majority.

## Propagation correctness — largely CONFIRMED CORRECT
Future-only filter (`eventDate >= now`) freezes past plans; `setPlan*` rejects past plans — consistent boundary. Per-plan reconcile is right: patches synced, skips overridden, inserts missing-and-not-in-detached-set, deletes synced-whose-template-item-vanished (+ cascades completions), and clears stale completions on a role↔team-level flip (mirrors `updateTask`). Delete loop iterates the pre-mutation snapshot so freshly-inserted rows aren't touched. Convex transactionality removes race concerns. Link/switch/unlink carryover (discard/copy), save-new, revert, and `duplicateEvent` linkage-carry all trace correctly; dangling pointer after a deleted template is handled gracefully everywhere (early returns / null name, no crash). `api.d.ts` hand-registration path is valid; schema additions are all optional/additive.

## Findings (all non-blocking; no correctness bug for the un-linked majority)
1. **[MED / PLAUSIBLE]** `saveTaskTemplateFromPlan` **replace** (L702) recreates template items with new ids, so propagation deletes+reinserts every synced row on **other** future linked plans — **destroying their completions** — and duplicates content onto siblings that hold overridden rows. Needs a guard/warning before Phase-4 UI ships.
2. **[LOW-MED / PLAUSIBLE]** **merge** (L736) appends the source plan's *entire* list incl. already-synced rows as new template items → duplicates across sibling plans. Consider rejecting merge when the plan is already linked to the target, or merging only LOCAL rows.
3. **[LOW / by-design]** `updateTask`/`updateItem` detach on *any* call, even a no-op edit — a Phase-4 autosave could silently detach linked rows. Gate on a real field delta.
4. **[LOW / SCALE]** Propagation is O(futurePlans × rows) inside the template-item mutation transaction, on every template item edit — a ceiling to watch for popular templates.
5. **[INFO]** Overridden rows keep a dangling `sourceTemplateItemId` if their template item is later deleted (delete loop skips them). Harmless but latent.
6. **[INFO / author-ack'd]** Plan-level `duplicateItem` correctly leaves the copy LOCAL, but resequences SYNCED siblings without detaching them → next propagation resets their sequence, possibly colliding with the local copy's sequence (ordering glitch only).

## Overall verdict
**Backend logic is sound and — most importantly — non-linked plans are provably unaffected.** No blocking correctness bug. The findings are edge cases in the not-yet-exposed save-as-template paths (completion loss / duplication on sibling plans) that should be guarded before the Phase-4 UI makes them reachable. Recommend addressing #1 and #2 (or explicitly deferring with a tracked note) before Phase 4; #3–#6 are polish. Not approving/requesting-changes — advisory review only.

PR #551, opened 2026-07-02
feat(events): plan-side template pickers + save-as-template + switch-carryover (Phase 4 UI)

lilseyi:
## Phase 4 (FRONTEND) - plan-side event templates

Adds the leader-facing template controls to the two plan authoring screens
(Event Tasks and Run Sheet). **UI only.**

> [!IMPORTANT]
> **Depends on the Phase 3 backend contract - merge Phase 3 first.** The
> `functions/scheduling/planTemplates` module (getPlanTemplateState +
> set/save/revert mutations) is being built in parallel and is not on `main`
> yet. These refs are addressed by **string path** via `makeFunctionReference`
> (same pattern as the Phase 1 refs already in `eventTemplates.ts`), so the UI
> typechecks on its own; runtime needs Phase 3 deployed.

### Controls added (to both screens)

- **Template picker** - `Task template: [ <name / None> v ]` (run sheet:
  `Run-sheet template: [...] v`). Opens an `AnchoredMenu` of the group's
  templates plus a **None (unlink)** row. Shown read-only when the event `isPast`.
- **Switch-with-carryover** - picking a *different* template while the plan
  `hasEdits` first prompts: **Copy my edits into `<name>`** (`carryover:"copy"`)
  / **Discard my edits** (`carryover:"discard"`) plus Cancel. No edits -> switches
  directly with `"discard"`.
- **"Edited for this event"** amber indicator (when `hasEdits`) with
  **Save to template** (confirm -> `save...FromPlan` existing/replace on the linked
  id; only when linked) and **Revert to template** (confirm -> `revert...Edits`).
- **Save as template v** (works linked or not): **New template...** (name prompt
  -> `mode:{kind:"new"}`) / **Add to existing...** (pick template -> Replace/Merge
  -> `mode:{kind:"existing"}`).

### Implementation notes

- One shared, presentational `PlanTemplateToolbar` drives both screens; it is
  string-typed on ids so it never juggles the two `Id` unions. Each screen wires
  its own kind's mutations and casts ids at the call boundary.
- Switch-carryover and Replace/Merge reuse one small radio `ChoiceModal`
  (`CustomModal`). Confirms use the shared `ConfirmModal`. Menus use
  `AnchoredMenu`.
- The new-template name prompt uses a plain controlled `TextInput` (not
  `InlineText`) - `InlineText`'s debounce/blur auto-save doesn't fit a
  Create-button prompt reliably.
- No NativeWind; `Pressable`/`TouchableOpacity` layout lives on static
  `StyleSheet` inner Views.

### Refs / functions called (reconcile if Phase 3 signatures differ)

- Query `functions/scheduling/planTemplates:getPlanTemplateState` `{planId}`
- `setPlanTaskTemplate` / `setPlanRunSheetTemplate` `{planId, templateId|null, carryover}`
- `saveTaskTemplateFromPlan` / `saveRunSheetTemplateFromPlan` `{planId, mode}`
- `revertPlanTaskTemplateEdits` / `revertPlanRunSheetTemplateEdits` `{planId}`
- Picker lists: `listTaskTemplates` / `listRunSheetTemplates` `{groupId}` (Phase 1, on main)

### Verification

- `packages/shared` builds clean; mobile `tsc --noEmit` clean for all four
  changed files (the repo-wide `@togather/shared` module-resolution errors are
  the known pre-existing baseline).
- **Not visually verified** - I cannot drive the simulator, and runtime needs
  Phase 3 merged.


apps/mobile/features/scheduling/components/RunSheetScreen.tsx
```diff
@@ -421,6 +529,16 @@ export function RunSheetScreen() {
                   {formatServiceRanges(times, duringTotalSec)}
                 </Text>
               ) : null}
+              <PlanTemplateToolbar
```
lilseyi: **[MEDIUM · CONFIRMED (no gate) / PLAUSIBLE (reachability)]** RunSheetScreen has **no internal leader gate**, unlike EventTasksScreen (which renders the toolbar only after its `if (!isLeader) return` at ~L512). Here the toolbar renders unconditionally for anyone who reaches this screen. The rostering routes (`app/rostering/_layout.tsx`, `[group_id]/_layout.tsx`) are plain `Stack` navigators with no route-level guard, and the roster grid renders for members — so a non-leader who reaches the run-sheet editor would see and be able to tap the template picker / Save-as / Revert controls.

Backend mutations are gated by `requirePlanScheduler` (Phase 3), so there is **no privilege escalation** — a non-leader who taps these just gets an error toast. But showing leader-only controls to non-leaders is a UX/consistency bug. Recommend gating the toolbar (and ideally the whole editor) on a leader check here, mirroring EventTasksScreen. The author already flagged this in the PR description; confirming it is real.

apps/mobile/features/scheduling/components/PlanTemplateToolbar.tsx
```diff
@@ -0,0 +1,647 @@
...
+                color={linkedId ? primaryColor : colors.textTertiary}
+              />
+            </TouchableOpacity>
+          </View>
+        )}
+
+        {/* Save as template */}
+        <View ref={saveAsRef} collapsable={false} style={styles.saveAsHost}>
+          <TouchableOpacity
+            onPress={openSaveAs}
```
lilseyi, line 251: **[MEDIUM · CONFIRMED]** `isPast` is only applied to the **picker** (rendered read-only above). The **Save-as template** button here, and the **Save to template / Revert to template** actions in the edited-indicator row, remain fully active for past events. Past events are otherwise treated as frozen — `setPlan{Task,RunSheet}Template` throws `"Past events are frozen"` in Phase 3. But `revertPlan…Edits` has **no isPast guard** in Phase 3 (it deletes all rows and re-instantiates from the template), so tapping Revert on a past event would mutate a supposedly-frozen plan. Save-as on a past event is arguably fine (harmless read into a new/existing template), but Revert is a destructive edit that contradicts the frozen model. Recommend hiding/disabling at least Revert (and probably Save-to-template) when `isPast`, consistent with the read-only picker.

apps/mobile/features/scheduling/components/PlanTemplateToolbar.tsx
```diff
@@ -0,0 +1,647 @@
...
+        visible={carryoverPrompt !== null}
+        title="Switch template"
+        message={`You've edited these ${itemNoun} for this event. Switching to "${
+          carryoverPrompt?.targetName ?? ""
+        }" will replace the list.`}
+        options={[
+          {
+            key: "copy",
+            label:
+              carryoverPrompt?.targetId == null
```
lilseyi: **[LOW · CONFIRMED]** On **unlink** (target = None) with edits, this modal offers "Keep my edited list" (`carryover:"copy"`) vs "Discard my edits" (`carryover:"discard"`). But the Phase 3 `setPlan{Task,RunSheet}Template` **null branch ignores `carryover` entirely** — it always strips the source tags and keeps the rows. So picking "Discard my edits" while unlinking silently does NOT discard; the list is kept regardless. Either the backend null-branch should honor `discard`, or the UI should not offer a discard choice when unlinking (targetId == null). Cross-PR note vs #550 — flagging so the contract stays honest.

apps/mobile/features/scheduling/api/eventTemplates.ts
```diff
@@ -286,3 +286,103 @@ export const reorderRunSheetTemplateItemsRef = makeFunctionReference<
...
+
+export const setPlanTaskTemplateRef = makeFunctionReference<
+  "mutation",
+  {
+    token: string;
+    planId: Id<"eventPlans">;
+    templateId: Id<"eventTaskTemplates"> | null;
+    carryover: TemplateCarryover;
+  },
+  null
```
lilseyi: **[LOW · CONFIRMED]** Return-type annotations on these mutation refs are inaccurate vs the Phase 3 backend: `setPlanTaskTemplate`/`setPlanRunSheetTemplate` actually return `{ planId, templateId }` and `revertPlan…Edits` returns `{ reverted: boolean }`, but both are typed `null` here. No runtime impact (the UI ignores return values, as the comment notes), but the annotations will mislead future callers. Reconcile when #550 lands.

apps/mobile/features/scheduling/api/eventTemplates.ts
```diff
@@ -286,3 +286,103 @@ export const reorderRunSheetTemplateItemsRef = makeFunctionReference<
...
+      strategy: SaveTemplateStrategy;
+    };
+
+export const getPlanTemplateStateRef = makeFunctionReference<
+  "query",
+  { token: string; planId: Id<"eventPlans"> },
+  PlanTemplateState
+>("functions/scheduling/planTemplates:getPlanTemplateState");
+
+export const setPlanTaskTemplateRef = makeFunctionReference<
```
lilseyi: **[LOW · INFO · CONFIRMED]** Coordination smell: #550 (Phase 3) **also** adds these exact seven refs in a new file `apps/mobile/features/scheduling/api/planTemplates.ts`, with identical string paths and (correct) return types. After both merge there will be two copies of the plan↔template refs (here in `eventTemplates.ts` and there in `planTemplates.ts`). Not a conflict (different files) but redundant. Cleanest resolution: drop these refs from this PR and import from #550's `api/planTemplates.ts` once it lands — the screens already import via `../api/eventTemplates`, so only the import source changes.

(Positive: the string paths `functions/scheduling/planTemplates:*` here match #550's backend module and its own refs **exactly** — the parallel-build contract is intact.)

lilseyi (commented):
🤖 Subagent review (codex unavailable) — Phase 4 UI, reviewed against the #550 Phase 3 contract.

## Contract match — CLEAN ✅ (the parallel-build risk)
All 7 refs in `api/eventTemplates.ts` match #550 **exactly** — verified module path + arg shape against the backend validators:
- `getPlanTemplateState {planId}` → return object matches `PlanTemplateState` field-for-field ✅
- `setPlan{Task,RunSheet}Template {planId, templateId Id|null, carryover}` — backend `carryoverValidator = "discard"|"copy"` ✅ (backend has it `v.optional`; UI always sends it — fine)
- `save{Task,RunSheet}TemplateFromPlan {planId, mode}` — `mode` = `{kind:"new",name}` | `{kind:"existing",templateId,strategy:"replace"|"merge"}` matches backend `taskSaveModeValidator`/`runSheetSaveModeValidator` ✅
- `revertPlan{Task,RunSheet}TemplateEdits {planId}` ✅
- Phase-1 `listTaskTemplates`/`listRunSheetTemplates {groupId}` return `{_id,name,itemCount}` ✅
- Token omitted at call sites, injected by `useAuthenticated*` hooks ✅
- String paths `functions/scheduling/planTemplates:*` are byte-identical to the paths #550 chose independently → **no ref-string mismatch.** No typos, no silent runtime failures.

## Flow correctness ✅
- **Switch-with-carryover:** `handlePick` → same-id no-op; `hasEdits` opens the copy/discard prompt FIRST (default selection "copy" = safe, never silently discards); no edits → direct switch with `"discard"`. `isPast` → picker is a static pill, no menu. No path silently drops edits without the prompt.
- **Save-as:** New → name prompt → `mode.kind:"new"`; Add-to-existing → pick → Replace/Merge → `mode.kind:"existing"`. Save-to-template (edited indicator) → confirm → existing+`"replace"` on the **linked** id, and is only shown when `linkedId` truthy. Revert → confirm → `revert`. Correct id in every case.
- **Per-slice wiring:** EventTasksScreen wires task fns + task slice; RunSheetScreen wires run-sheet fns + run-sheet slice. **No cross-wiring.**

## Conventions ✅
No NativeWind/className. Layout on static `StyleSheet` inner Views; `TouchableOpacity` for press feedback (no function-style Pressable `style`). All hooks in the toolbar, `ChoiceModal`, and both screens are declared **before** any early return (EventTasksScreen returns at ~L487/512 sit after the new hooks at L131–264). `ChoiceModal` re-seeds its default on each open via `useEffect([visible])` — no state leak between opens; the new-template `TextInput` resets `nameDraft` on open/submit. Empty/loading handled: `templates ?? []` → menu shows only "None (unlink)"; `state === undefined` → picker disabled, no crash on undefined.

## Findings (see inline)
1. **MEDIUM** — RunSheetScreen.tsx:532 — no internal leader gate (EventTasksScreen has one). No privilege escalation (backend `requirePlanScheduler`-gated), but non-leaders reaching the editor would see/tap leader-only controls. Author-flagged; confirmed real. Gate the toolbar here.
2. **MEDIUM** — PlanTemplateToolbar.tsx:241 — `isPast` disables only the picker; Save-as and (destructive) **Revert** stay active. #550's `revertPlan…Edits` has no isPast guard, so Revert can mutate a frozen past event. Hide/disable Revert (& Save-to) when `isPast`.
3. **LOW** — PlanTemplateToolbar.tsx:347 — unlink+"Discard my edits" is a no-op: #550's `setPlan…Template` null-branch ignores `carryover` and always keeps rows. UI offers a choice the backend won't honor.
4. **LOW** — eventTemplates.ts:352 — `setPlan*`/`revert*` ref return types annotated `null`; backend returns `{planId,templateId}` / `{reverted}`. Return values unused, so cosmetic.
5. **LOW/INFO** — eventTemplates.ts:344 — #550 ships the same 7 refs in `api/planTemplates.ts`; post-merge duplication. Prefer importing #550's file.

## No regressions
Toolbar is purely additive to both list headers (EventTasksScreen listHeader / RunSheetScreen listHeader) — grid/run-sheet rendering, layout, and scroll are untouched.

## Verdict
**Approve-with-nits (do not merge until #550 lands).** No blocking correctness bug and the contract is exact. Address the two MEDIUM UX/gating items (leader gate on RunSheetScreen; freeze Save-as/Revert when isPast) before or shortly after merge; the LOW items are cleanup. Not visually verified (runtime needs #550 deployed).

PR #569, opened 2026-07-07
Desktop two-pane layout for the dev dashboard at togather.nyc/dev

lilseyi:
## Summary

Gives the contributor dev dashboard its desktop shape and a stable public URL.

- **Two-pane split view on wide web windows** (`ContributeSplitView`): a 340px conversation sidebar — the existing list screen reused with selection highlighting — beside the active thread, with a "Select a conversation" placeholder. Follows the inbox's `useIsDesktopWeb` breakpoint convention. Phone behavior is pixel-identical: both screens are reused whole with optional props whose defaults reproduce today's exact flow.
- **Routes move from `/(user)/contribute` to `/(user)/dev`** so the public URL is `togather.nyc/dev`. The previous `/contribute` URL collides with the marketing page the Cloudflare worker serves at that path (a refresh mid-session would swap sites). Static dev-tool routes (`/dev/feature-flags` etc.) coexist with the new dynamic routes — static wins in Expo Router.
- **Worker**: `"dev"` added to `KNOWN_APP_ROUTES` so the path is served by the Expo web app instead of being treated as a community slug.
- Deep links keep working: `/dev/<id>` lands in the split view pre-selected on wide screens, the normal detail screen on phones.

## Test plan

- `features/contribute` + `features/profile` suites: 17/17 pass.
- Mobile typecheck and eslint clean; navigator-screens and native-imports guard scripts pass.
- No new dependencies, no `runtimeVersion` change.
- Worker change is additive (one entry in an allowlist); needs the usual worker deploy to take effect on togather.nyc.

https://claude.ai/code/session_01Qd3CXkKqqCscq4K3fxjtqP

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qd3CXkKqqCscq4K3fxjtqP)_


apps/mobile/app/(user)/dev/[id].tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useIsDesktopWeb } from "@hooks/useIsDesktopWeb";
+import type { Id } from "@services/api/convex";
+
+export default function ContributionDetailRoute() {
+  const isDesktopWeb = useIsDesktopWeb();
+  const params = useLocalSearchParams<{ id: string }>();
+  if (isDesktopWeb) {
+    return (
+      <ContributeSplitView
+        initialId={(params.id || null) as Id<"devBugs"> | null}
```
Comment (lilseyi) [inline apps/mobile/app/(user)/dev/[id].tsx:25]: **[verified | high] `/dev/<malformed-id>` crashes to the app-root error boundary**

`params.id` is cast to `Id<"devBugs">` unvalidated and handed to `v.id`-validated queries — a truncated shared URL or a staging id opened against prod throws `ArgumentValidationError`, `useQuery` rethrows during render, and the whole app is replaced by the generic error screen; the screen's own "We couldn't find this conversation" state is unreachable. The `/dev` namespace is now a public URL people will share, so garbage params are expected traffic.

**Fix:** validate/normalize the id before querying (e.g. pass `null` → skip + render the existing not-found state when the server rejects it, or use `normalizeId`-style client guard).

---

apps/mobile/features/contribute/components/ContributeSplitView.tsx
```diff
@@ -0,0 +1,106 @@
...
+  initialId?: Id<"devBugs"> | null;
+}
+
+export function ContributeSplitView({ initialId }: ContributeSplitViewProps) {
+  const { colors } = useTheme();
+  const [selectedId, setSelectedId] = useState<Id<"devBugs"> | null>(
+    initialId ?? null,
+  );
+
+  // Re-seed when the route param changes without a remount — e.g. the submit
```
lilseyi: **[verified | medium] Sidebar selection never syncs to the URL — resize or refresh silently swaps conversations and drops drafts**

Selecting conversation B in the sidebar leaves the URL at `/dev/A`; a resize below the breakpoint (or a refresh) re-derives the view from `params.id` and the user lands back on A with their B draft gone. The inbox split this mirrors derives the active item from `usePathname`, so URL and view never diverge there.

**Fix:** make selection navigate — `router.setParams`/`router.replace` to `/dev/<selectedId>` on sidebar clicks (keeping the split view mounted), so URL is always the source of truth.

---

apps/mobile/features/contribute/components/ContributionDetailScreen.tsx
lilseyi, line 1: **[verified | blocker on this PR's own surface] Approve spec / Archive are silent no-ops on desktop web**

The confirmations use `Alert.alert` (~line 250 `handleApprove`, ~line 339 `handleArchive`), which react-native-web doesn't implement — buttons never render, so `approveSpec`/archive are never called. The Alert code predates this PR, but this PR promotes desktop web to a first-class surface where the dashboard's core approval loop is dead. The repo already has the established workaround (`window.confirm` on web — see `EventPageClient.tsx:701` and `SongLibraryScreen.tsx:62`). Error-reporting `Alert.alert` calls in this screen and SubmitContributionScreen are likewise invisible on web.

**Fix:** platform-aware confirm/notify helper (web → `window.confirm`/inline error, native → Alert) used for approve, archive, and error surfaces.

---

apps/mobile/features/contribute/components/ContributeListScreen.tsx
lilseyi, line 1: **[verified | low] The embedded sidebar keeps its mobile back-chevron**

`ContributionDetailScreen` got an `embedded` prop to drop its back button, but the list screen didn't — so the persistent desktop sidebar shows a chevron (~line 259) that either does nothing (fresh-tab deep load, no history) or dismisses the entire dashboard while reading as sidebar chrome. Give the list screen the same `embedded` treatment.

---

lilseyi (commented):
**AI review cycle — 4 verified findings** (3 review angles, each candidate adversarially verified; refuted candidates included hook-order hazards, selection-reset semantics, worker slug handling, and stale route references — none survived).

Most severe: the split view's core actions (Approve spec / Archive) are dead on desktop web because `Alert.alert` doesn't render on react-native-web — pre-existing code, but this PR makes desktop the headline surface, so it blocks here. Also: unvalidated `/dev/<id>` params crash to the root error boundary, sidebar selection doesn't sync to the URL (drafts lost on resize/refresh), and the embedded sidebar keeps a misleading back-chevron.

Noted for follow-up (pre-existing, not blocking): `resolveNotificationNavigation` has no `dev_contribution_update` case, so dashboard pushes don't open the conversation on tap — the payload already carries `bugId`. Doc touch-ups: ADR-029 and route header comments still reference the old `/contribute` paths; the worker comment says "redirects" inaccurately.

Fixes incoming on this branch; threads resolve as they land.

---

PR #570, opened 2026-07-07
Fix keyboard covering the reply box on contributor conversations

lilseyi:
## What this fixes

On the contributor dashboard (**Profile → Contribute**), opening a conversation and tapping the **"Message @Togather…"** reply box slid the on-screen keyboard up **over the composer**, so you couldn't see what you were typing — the input sat hidden behind the keyboard. This is what the reporter hit while "doing maintenance" in the dashboard.

## Why it happened

The contributor dashboard opens as an iOS **sheet** (the `(user)` route group is presented as a modal). On `ContributionDetailScreen.tsx` the composer's `KeyboardAvoidingView` used `behavior="padding"` with **no `keyboardVerticalOffset`**. Inside a sheet presentation the view's measured frame is offset from the window, so with no offset the composer isn't lifted enough and ends up behind the keyboard.

A sibling screen in the **same** modal group — `app/(user)/dev/notifications.tsx` — already handles this with `keyboardVerticalOffset={Platform.OS === 'ios' ? 64 : 0}`. This screen was simply missing that offset.

## The change

Add `keyboardVerticalOffset={Platform.OS === "ios" ? 64 : 0}` to the one `KeyboardAvoidingView` in `ContributionDetailScreen.tsx`, mirroring the working sibling in the same sheet group.

- **iOS:** composer, attachment strip, and the composer hint line now ride up together above the keyboard.
- **Android:** left on its default resize behavior (unchanged), per the spec.
- Scope: a single line (plus an explanatory comment) in one file.

## A note on the spec's fallback

The spec offered a fallback of switching to `react-native-keyboard-controller`'s `KeyboardAvoidingView`, stating "`KeyboardProvider` is already mounted at the app root." **That is not the case** — `KeyboardProvider` was deliberately *removed* from `app/_layout.tsx` because it "was causing conflicts with Stream Chat's overlay system during interactive keyboard dismiss" (a comment marks where it used to be). Re-mounting it to enable that component would risk regressing chat and is out of scope, so the recommended fixed-offset approach is the right in-scope path here. The `64` is the empirically-working value from the sibling screen in the identical modal group, not a guess.

## Verification

- ✅ Native import gating, native fingerprint (no native change; `runtimeVersion` untouched), navigator-screen registration checks
- ✅ ESLint clean on the changed file
- ✅ Full mobile Jest suite (131 suites / 1227 tests) passes
- ⚠️ **Visual confirmation on a device/simulator is still pending** — this build environment is Linux, so no iOS simulator was available to screenshot the before/after. The offset value is carried over from a sibling screen that already renders correctly in the same sheet presentation, so behavior should match. Recommend a quick visual check on iOS staging before merge.

## Dashboard item

Contributor dashboard bug `x973r00983txvy4bssjtf51nq58a2qzf` — reported by **Seyi Olujide**.

https://claude.ai/code/session_01GQhTVQ92p3pLf2qS8iQ5Sg

---
_Generated by [Claude Code](https://claude.ai/code/session_01GQhTVQ92p3pLf2qS8iQ5Sg)_

### DIFF
--- a/apps/mobile/features/contribute/components/ContributionDetailScreen.tsx
+++ b/apps/mobile/features/contribute/components/ContributionDetailScreen.tsx
@@ -435,6 +435,11 @@ export function ContributionDetailScreen() {
     <KeyboardAvoidingView
       style={[styles.container, { backgroundColor: colors.background, paddingTop: insets.top }]}
       behavior={Platform.OS === "ios" ? "padding" : undefined}
+      // This screen opens as an iOS sheet (the `(user)` route group is a modal),
+      // where "padding" alone doesn't lift the composer above the keyboard — it
+      // stays hidden behind it. The sibling `dev/notifications.tsx`, in the same
+      // modal group, corrects the sheet inset with a 64pt offset; mirror it here.
+      keyboardVerticalOffset={Platform.OS === "ios" ? 64 : 0}
     >
       <View style={styles.headerRow}>
         <TouchableOpacity onPress={() => router.back()} style={styles.backBtn}>


lilseyi (commented):
**Automated review — verdict: approved (low risk).**

_Note on identity: a formal approving review from the reviewer bot account is blocked by this org's session policy, so this is posted as a review comment. The approval verdict is delivered to the dashboard via the pipeline callback._

Reviewed the one-file, +5-line diff across four lenses (correctness, security, spec-fidelity, tests/UX), then adversarially verified each candidate finding. **No surviving findings.**

- **Correctness:** `keyboardVerticalOffset={Platform.OS === "ios" ? 64 : 0}` only affects iOS — on Android `behavior` is `undefined`, so `KeyboardAvoidingView` is a pass-through `View` and the `0` offset is a no-op. No interaction/double-counting with the KAV's internal `paddingTop: insets.top` (internal padding doesn't change the view's measured window frame, which is what the offset math uses).
- **Spec-fidelity:** Implements exactly the spec's recommended fix, scoped to the single file. Good catch correcting the spec's fallback claim — `KeyboardProvider` was in fact deliberately removed from `_layout.tsx` (Stream Chat overlay conflict), so the fixed-offset path is the right in-scope choice.
- **Justification of `64`:** Confirmed both this screen (`app/(user)/contribute/[id].tsx`) and the sibling `app/(user)/dev/notifications.tsx` live in the same `(user)` `presentation: "modal"` group, so the offset is carried from an identical sheet presentation rather than guessed.
- **UX edge cases:** Composer, attachment strip, hint line, and the staging "Something's off" note input all sit inside this KAV / its `ScrollView`, so the uniform offset lifts them together; the note input stays scrollable via `keyboardShouldPersistTaps="handled"`.
- **Tests:** None warranted — a platform-conditional layout constant; a unit test would be tautological/brittle.

**One non-blocking caveat (already disclosed in the PR body):** the exact `64` value could not be visually confirmed on an iOS sheet from this Linux environment. It's the empirically-working value from the identical sibling presentation, so it should match — recommend a quick visual check on iOS staging before merge that the composer lands flush above the keyboard with no leftover gap.

---

PR #575, opened 2026-07-07
Remove manual billing override — active-member count is always automatic

lilseyi:
## What & why

The per-active-member bill had an admin/leader **"mark inactive"** toggle (on a member's admin person page) that excluded someone from the count regardless of app use. That's an abuse vector: a community could mark members inactive to zero out its bill while those members keep using the app. Removed entirely — the automatic **30-day activity rule is now the single source of truth**.

## Changes
- **`memberActivity.ts`**: deleted `setMemberBillingActive` and `canManageBillingActivity`; dropped the `billingInactive` skip from `countBillableActiveUsers`.
- **`admin/members.ts`**: `getCommunityMemberById` no longer returns `billingInactive`.
- **`PersonDetailScreen.tsx`**: removed the entire Billing row ("Counts as active member" / Mark inactive toggle), its handler, mutation, state, and styles.
- **Copy**: Go Live screen, the web guide callout (now "Nothing to manage"), ADR-030, and the cron/sync comments all state the count is fully automatic with no override.
- **Schema**: `userCommunities.billingInactive` is kept but marked **DEPRECATED/unread**. Dropping a Convex field that may already hold values requires a clear-migration first, so it stays inert (nothing reads or writes it) pending a follow-up drop. A test patches a stray `billingInactive: true` and asserts it's ignored, proving the abuse vector is gone.

## Testing
- 2,162 backend tests pass. The `member-activity` suite drops the manual-inactive/permission tests and adds "an active member always counts — the count is purely automatic" (including the stale-flag-ignored assertion).
- Mobile + web typecheck clean; web guide builds.

https://claude.ai/code/session_01VQLRNeL7gUToheWEhBnrLs

---
_Generated by [Claude Code](https://claude.ai/code/session_01VQLRNeL7gUToheWEhBnrLs)_

### DIFF
--- a/apps/convex/__tests__/member-activity.test.ts
+++ b/apps/convex/__tests__/member-activity.test.ts
@@ -222,120 +222,28 @@ describe("getBillableSummary", () => {
     expect(summary.billableActiveUsers).toBe(2);
   });
 
-  test("manually marked inactive members are excluded until re-activated", async () => {
+  test("an active member always counts — the count is purely automatic", async () => {
     const t = convexTest(schema, modules);
     const { communityId, adminToken } = await seedCommunity(t);
-    const memberId = await addMember(t, communityId, "+15555560006", { age: 0 });
+    await addMember(t, communityId, "+15555560006", { age: 0 });
 
-    await t.mutation(api.functions.memberActivity.setMemberBillingActive, {
-      token: adminToken,
-      communityId,
-      targetUserId: memberId,
-      active: false,
-    });
-    let summary = await t.query(api.functions.memberActivity.getBillableSummary, {
-      token: adminToken,
-      communityId,
-    });
-    expect(summary.billableActiveUsers).toBe(1); // admin only
-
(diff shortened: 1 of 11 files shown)


lilseyi (commented):
## 🔍 Self-review (Codex out of credits)

Ran the in-repo review over this diff (a small, deletion-heavy change: 10 files, net −244 lines). One finding, fixed in `ca83a2f`:

- **Stale user-facing copy in the pre-bill email** — the monthly `billing.monthly_preview` email still told admins "an admin or their group leader can mark them inactive from their person page before the 1st," an action this PR removes. Reworded to state the count updates automatically; fixed the matching `ee/billing.ts` comment and the test docstring too.

Otherwise clean: no dangling references to the deleted `setMemberBillingActive` / `canManageBillingActivity` / `billingInactive` (typecheck + a codebase grep both confirm), imports are all still used, and the dormant schema field is proven inert by a test that patches a stray `billingInactive: true` and asserts the member still bills. 2,162 backend tests pass; mobile + web typecheck clean.

---

PR #596, opened 2026-07-11
Emoji reactions on prayers, updates & praise reports

lilseyi:
## What this adds

A lightweight way to respond to a prayer without completing a full 3-minute pray session. Members can now react with a small, curated set of emoji to:

1. **Prayer requests** (the original post)
2. **Updates** (author follow-ups)
3. **Praise reports** (author follow-ups celebrating answered prayer)

Reactions appear as badge pills under each card on both prayer **detail** screens (`PrayedPrayerDetailScreen`, `MyPrayerDetailScreen`). A **＋** button opens a popover of the 6 curated emoji; tapping a badge you gave toggles it off; long-pressing a badge shows who reacted. The "pray now" deck (`PrayerScreen`) intentionally gets no reactions — you react from the detail view.

**Curated set (6):** ❤️ love · 🙏 praying/amen · 🎉 celebrate · 🙌 praise · 🕊️ peace · 🥹 moved. This is the one product-flavored choice and it's a single constant (`PRAYER_REACTION_EMOJIS`) in both `apps/convex/functions/prayers/reactions.ts` (server allowlist) and `apps/mobile/features/prayer/components/PrayerReactions.tsx` (picker) — trivial to adjust.

## Rendered mock (not a device capture)

This run executes on a headless Linux runner with no iOS simulator, so the image below is a **faithful mock rendered from the real component styles**, not a screenshot from a device.

[Prayer reactions — before/after](https://images.togather.nyc/dev-assistant/f50d8198-5235-405b-800a-f89584408c8c-prayer-reactions-after.png)

## How it works

**Backend** — mirrors the chat reactions pattern:
- New polymorphic `prayerReactions` table (`targetType: "prayer" | "followUp"`, `targetId`, `userId`, `emoji`, denormalized `communityId`), indexed `by_target` and `by_target_user`.
- `prayers.reactions.toggleReaction({ targetType, targetId, emoji })` — inserts/deletes one row per (target, user, emoji). Guards: the **emoji must be on the server-side allowlist**, and the caller must pass the **same visibility gate `getDetail` enforces** (community member who authored the prayer or already has a `prayerResponses` row).
- `prayers.reactions.getReactionDetails` — the "who reacted" list.
- `getDetail` now **folds aggregated `{ emoji, count, hasReacted }` reactions** onto the request card and every follow-up in one round-trip (no per-card N+1).
- **No notifications** are sent on reaction — silent, like chat.

**Frontend** — self-contained prayer components (the chat reaction hooks are hard-wired to `chatMessages` IDs, so duplicating the small presentational pieces was cleaner than generalizing chat internals):
- `PrayerReactions` — badge pills (count hidden at 1, own reaction highlighted blue like chat), the ＋ picker popover, and toggle wiring. Convex reactivity refreshes `getDetail` after each toggle.
- `PrayerReactionDetailsModal` — the long-press "who reacted" modal.

## Edge cases handled

- **Anonymity** — reacting is a social action and is attributed to the reactor; the prayer author's own anonymity is untouched.
- **Access** — reactions can only be fetched/created for prayers the caller can already see; hidden/rejected prayers can't be reacted to (they never reach `getDetail`).
- **Toggle idempotency** — double-tapping the same emoji nets to add-then-remove, never two rows (enforced via `by_target_user`).

## Tests

- New `apps/convex/__tests__/prayers/reactions.test.ts` (8 tests): toggle add/remove, double-add idempotency, allowlist rejection, non-member rejected, member-who-hasn't-prayed rejected, author + prayed-for allowed on request and follow-up, `getDetail` folding with shared cross-user counts + per-viewer `hasReacted`, and `getReactionDetails`.
- Existing chat reactions tests still pass (21/21 across both suites). Convex + mobile prayer files typecheck; native-import gating and lint pass.

## Docs

- Updated `apps/web/src/pages/guides/Prayer.tsx` ("Answered prayers & follow-ups") to describe the new reactions.

---

Dev dashboard item: `x978hjbfpafkgh2hpe4vxjm1xs8a8f9h` (requested by Seyi Olujide).

---
_Generated by [Claude Code](https://claude.ai/code/session_01TX9Xp1cx7kMVJo7F5bTR5W)_

### DIFF
--- a/apps/convex/__tests__/prayers/reactions.test.ts
+++ b/apps/convex/__tests__/prayers/reactions.test.ts
@@ -1,0 +1,377 @@
+/**
+ * Tests for emoji reactions on prayers and prayer follow-ups.
+ *
+ * Covers: toggle add/remove, the curated-emoji allowlist, access control
+ * (author + prayed-for allowed, non-member and non-prayer rejected), follow-up
+ * targets, folding into getDetail (no extra round-trip), cross-user
+ * aggregation, and the "who reacted" details query.
+ */
+
+import { convexTest } from "convex-test";
+import { expect, test, describe } from "vitest";
+import schema from "../../schema";
+import { modules } from "../../test.setup";
+import { api } from "../../_generated/api";
+import { generateTokens } from "../../lib/auth";
+import type { Id } from "../../_generated/dataModel";
+
+process.env.JWT_SECRET = "test-jwt-secret-for-unit-tests-minimum-32-chars";
+
+const HEART = "❤️";
+const PRAY = "🙏";
+
+interface Seeded {
+  communityId: Id<"communities">;
+  prayerId: Id<"prayers">;
+  followUpId: Id<"prayerFollowUps">;
+  authorId: Id<"users">;
+  authorToken: string;
+}
+
+async function makeUser(
+  t: ReturnType<typeof convexTest>,
+  communityId: Id<"communities">,
+  firstName: string,
+  phone: string,
+  { member = true }: { member?: boolean } = {},
+): Promise<{ userId: Id<"users">; token: string }> {
+  const userId = await t.run(async (ctx) => {
+    return await ctx.db.insert("users", {
+      firstName,
+      lastName: "Tester",
+      phone,
+      phoneVerified: true,
+      activeCommunityId: communityId,
+      createdAt: Date.now(),
+      updatedAt: Date.now(),
+    });
+  });
+  if (member) {
+    await t.run(async (ctx) => {
+      await ctx.db.insert("userCommunities", {
+        userId,
+        communityId,
+        status: 1,
+        createdAt: Date.now(),
+        updatedAt: Date.now(),
+      });
+    });
+  }
+  const { accessToken } = await generateTokens(userId);
+  return { userId, token: accessToken };
+}
+
+async function seed(t: ReturnType<typeof convexTest>): Promise<Seeded> {
+  const communityId = await t.run(async (ctx) => {
+    return await ctx.db.insert("communities", {
+      name: "Test Community",
+      subdomain: "test",
+      slug: "test",
+      timezone: "America/New_York",
+      churchFeatures: { prayerEnabled: true },
+      createdAt: Date.now(),
+      updatedAt: Date.now(),
+    });
+  });
+
+  const { userId: authorId, token: authorToken } = await makeUser(
+    t,
+    communityId,
+    "Author",
+    "+15555550001",
+  );
+
+  const prayerId = await t.run(async (ctx) => {
+    return await ctx.db.insert("prayers", {
+      communityId,
+      authorUserId: authorId,
+      isAnonymous: false,
+      bodyText: "Please pray for my mom's surgery",
+      status: "active",
+      prayedForCount: 0,
+      moderationStatus: "approved",
+      createdAt: Date.now(),
+      updatedAt: Date.now(),
+    });
+  });
+
+  const followUpId = await t.run(async (ctx) => {
+    return await ctx.db.insert("prayerFollowUps", {
+      prayerId,
+      authorUserId: authorId,
+      kind: "praise_report",
+      bodyText: "Surgery went perfectly!",
+      createdAt: Date.now(),
+    });
+  });
+
+  return { communityId, prayerId, followUpId, authorId, authorToken };
+}
+
+/** Give a user a prayerResponses row so they pass the "hasPrayed" gate. */
+async function recordPrayed(
+  t: ReturnType<typeof convexTest>,
+  prayerId: Id<"prayers">,
+  communityId: Id<"communities">,
+  userId: Id<"users">,
+) {
+  await t.run(async (ctx) => {
+    await ctx.db.insert("prayerResponses", {
+      prayerId,
+      userId,
+      communityId,
+      prayedAt: Date.now(),
+    });
+  });
+}
+
+describe("prayer reactions", () => {
+  test("author can add a reaction; toggling again removes it", async () => {
+    const t = convexTest(schema, modules);
+    const { prayerId, authorToken } = await seed(t);
+
+    await t.mutation(api.functions.prayers.reactions.toggleReaction, {
+      token: authorToken,
+      targetType: "prayer",
+      targetId: prayerId,
+      emoji: HEART,
+    });
+
+    let rows = await t.run(async (ctx) =>
+      ctx.db.query("prayerReactions").collect(),
+    );
+    expect(rows).toHaveLength(1);
+    expect(rows[0].emoji).toBe(HEART);
+    expect(rows[0].targetType).toBe("prayer");
+
+    // Toggle off
+    await t.mutation(api.functions.prayers.reactions.toggleReaction, {
+      token: authorToken,
+      targetType: "prayer",
+      targetId: prayerId,
+      emoji: HEART,
+    });
+    rows = await t.run(async (ctx) =>
+      ctx.db.query("prayerReactions").collect(),
+    );
+    expect(rows).toHaveLength(0);
+  });
+
+  test("double add of the same emoji nets to one row then none", async () => {
+    const t = convexTest(schema, modules);
+    const { prayerId, authorToken } = await seed(t);
+
+    // Two toggles = add then remove = zero rows, never two.
+    await t.mutation(api.functions.prayers.reactions.toggleReaction, {
+      token: authorToken,
+      targetType: "prayer",
+      targetId: prayerId,
+      emoji: PRAY,
+    });
+    await t.mutation(api.functions.prayers.reactions.toggleReaction, {
+      token: authorToken,
+      targetType: "prayer",
+      targetId: prayerId,
+      emoji: PRAY,
+    });
+    const rows = await t.run(async (ctx) =>
+      ctx.db.query("prayerReactions").collect(),
+    );
+    expect(rows).toHaveLength(0);
+  });
+
+  test("rejects an emoji outside the curated allowlist", async () => {
+    const t = convexTest(schema, modules);
+    const { prayerId, authorToken } = await seed(t);
+
+    await expect(
+      t.mutation(api.functions.prayers.reactions.toggleReaction, {
+        token: authorToken,
+        targetType: "prayer",
+        targetId: prayerId,
+        emoji: "💩",
+      }),
+    ).rejects.toThrow(/invalid_reaction_emoji/);
+
+    const rows = await t.run(async (ctx) =>
+      ctx.db.query("prayerReactions").collect(),
+    );
+    expect(rows).toHaveLength(0);
+  });
+
+  test("a member who has not prayed cannot react", async () => {
+    const t = convexTest(schema, modules);
+    const { prayerId, communityId } = await seed(t);
+    const { token } = await makeUser(
+      t,
+      communityId,
+      "Bystander",
+      "+15555550002",
+    );
+
+    await expect(
+      t.mutation(api.functions.prayers.reactions.toggleReaction, {
+        token,
+        targetType: "prayer",
+        targetId: prayerId,
+        emoji: HEART,
+      }),
+    ).rejects.toThrow(/prayer_not_accessible/);
+  });
+
+  test("a non-member cannot react even after (impossibly) prayed", async () => {
+    const t = convexTest(schema, modules);
+    const { prayerId } = await seed(t);
+    // User in NO community.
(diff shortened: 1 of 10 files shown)


apps/convex/functions/prayers/reactions.ts
```diff
@@ -0,0 +1,273 @@
...
+          "Unknown";
+        return {
+          userId: reaction.userId,
+          displayName,
+          profilePhoto: user.profilePhoto ?? null,
+        };
+      }),
+    );
+
+    return users.filter((u): u is NonNullable<typeof u> => u !== null);
```
lilseyi, line 303: **⛔ Blocking — anonymous prayer authors can be de-anonymized through the "who reacted" list.**

`getReactionDetails` returns every reactor's real `displayName` + `profilePhoto` with no exception for the author of an **anonymous** prayer. Combined with two facts, this defeats the prayer's anonymity guarantee:

- `resolveAccessiblePrayer` grants the author access unconditionally (`reactions.ts:93`), so an anonymous author *can* react to their own request/follow-up.
- `MyPrayerDetailScreen` renders the reaction row (incl. the ＋) on the author's own card with no `isAnonymous` guard (`MyPrayerDetailScreen.tsx:215,275`).

**Scenario:** User A posts a prayer with `isAnonymous: true` and taps ❤️ on it from their own detail screen. User B (who prayed for it and sees only "Anonymous" on the card) long-presses the ❤️ badge → `getReactionDetails` returns `{ displayName: "A. Tester" }` — B now knows who authored the "anonymous" prayer. Note the name returned here is the **full** first+last name, not the "First L." form used elsewhere, so the leak is complete.

The spec's anonymity note ("hides the author, not reactors") didn't consider the author reacting to their *own* anonymous prayer, and there's no warning to the author that doing so unmasks them. In this tender context (crisis-flagged prayers included) this should be closed before merge. Suggested fixes, any one of:
1. Suppress the author from the reactor list **and** from the aggregated count/`hasReacted` when `prayer.isAnonymous` (so a lone count doesn't leak by inference), or
2. Hide the reaction UI on an anonymous author's *own* card, or
3. Block self-reaction on anonymous prayers.

Please add a regression test for whichever approach you take.

---

apps/convex/functions/prayers/reactions.ts
```diff
@@ -0,0 +1,273 @@
...
+  ctx: QueryCtx,
+  userId: Id<"users">,
+  targetType: PrayerReactionTargetType,
+  targetId: string,
+): Promise<Doc<"prayers"> | null> {
+  let prayer: Doc<"prayers"> | null;
+  if (targetType === "prayer") {
+    prayer = await ctx.db.get(targetId as Id<"prayers">);
+  } else {
+    const followUp = await ctx.db.get(targetId as Id<"prayerFollowUps">);
```
lilseyi: **Non-blocking (robustness):** `targetId` is a bare `v.string()`, so `ctx.db.get(targetId as Id<…>)` throws Convex's "invalid id" error (an unhandled 500) on a malformed string, instead of falling through to the intended clean deny (`prayer_not_accessible` in the mutation, `[]` here in `getReactionDetails`). Not reachable through the current UI — both detail screens always pass a real `detail.id`/`f.id` — so this is low priority. Chat uses `v.id("chatMessages")`; since the target is polymorphic you can't do that directly, but consider wrapping the two `db.get` calls so a bad id resolves to `null`/deny rather than surfacing a raw error.

---

lilseyi (commented):
## Automated review — Prayer reactions (4 parallel reviewers: correctness, security, spec-fidelity, tests)

**Verdict: changes requested** — one blocking privacy issue, everything else is solid.

**What's good:** The feature faithfully implements the spec's "Done when" checklist. Access control is genuinely equivalent to `getDetail`'s gate, the polymorphic `targetId` fails *safe* (a wrong-table/foreign id yields an undefined `communityId` → access denied), toggle idempotency holds via the `by_target_user` uniqueness path, aggregation counts + per-viewer `hasReacted` are correct, no push is sent, reactions fold into `getDetail` in one round-trip, and the emoji allowlist is enforced server-side. Tests run green (8/8 new, 13/13 existing chat — no regressions), and the "who reacted" and count-hidden-at-1 UX match chat.

**Blocking (1):** Anonymous prayer authors can be de-anonymized through the "who reacted" list — see the inline comment on `prayers/reactions.ts`. Two independent reviewers surfaced this. Please close it before merge.

**Non-blocking suggestions:**
- Robustness on unvalidated `targetId` — inline comment on `prayers/reactions.ts`.
- Test coverage is strong on the spec-required paths; a few worthwhile additions with the fix round: (1) a follow-up whose parent prayer is in a **different community** than the caller — the cross-community gate leans on the parent's `communityId` and is currently untested; (2) add two reactions, remove one, then re-read `getDetail` to assert the aggregated count goes 2→1 (only the increment path is covered today); (3) two *different* emojis on one target to lock in the curated-order rendering.

_This review was posted as inline comments + this summary rather than a formal approve/request-changes review, which routine sessions can't submit. Verification ran on a headless Linux runner (no iOS simulator): backend logic was exercised via the Convex test suite; UI was reviewed statically against the component styles._

---

apps/convex/functions/prayers/reactions.ts
lilseyi: Fixed in 7c2c269 with approach 3 (block self-reaction on anonymous prayers). `toggleReaction` now rejects with `anonymous_author_cannot_react` when `prayer.isAnonymous && prayer.authorUserId === userId`, so no author reaction row can ever exist on an anonymous prayer — closing the leak at the source (no row → nothing in the reactor list, count, or `hasReacted` to infer from). As defense in depth, the mobile UI hides the ＋/toggle on an anonymous author's own cards (`PrayerReactions` gains a `canReact` prop; `MyPrayerDetailScreen` passes `canReact={!detail.isAnonymous}`). Regression tests added: author blocked on their own anonymous request and follow-up, and a prayed-for viewer can still react to an anonymous prayer (only the author is blocked).

---

apps/convex/functions/prayers/reactions.ts
```diff
@@ -0,0 +1,273 @@
...
+  const isAuthor = prayer.authorUserId === userId;
+  if (isAuthor) return prayer;
+
+  const hasPrayed = !!(await ctx.db
+    .query("prayerResponses")
+    .withIndex("by_prayer_user", (q) =>
+      q.eq("prayerId", prayer!._id).eq("userId", userId),
+    )
+    .first());
+  return hasPrayed ? prayer : null;
```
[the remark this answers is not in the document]
lilseyi: Fixed in 7c2c269. `resolveAccessiblePrayer` now requires `moderationStatus === "approved"` for non-author viewers, so a prayed-for user reaching a later-rejected prayer via a stale detail screen or deep link is denied (`prayer_not_accessible`) for both `toggleReaction` and `getReactionDetails`. The author keeps access to their own prayer regardless of moderation state, matching `getDetail`. Added a regression test where a viewer who prayed for a since-rejected prayer is rejected.

---

apps/convex/functions/prayers/reactions.ts
lilseyi: Fixed in 7c2c269. `resolveAccessiblePrayer` now resolves the target via `ctx.db.normalizeId("prayers"/"prayerFollowUps", targetId)`, returning `null` for a malformed id instead of letting `ctx.db.get` throw the raw "invalid id" error. So a bad id now denies cleanly — `prayer_not_accessible` in the mutation, `[]` in `getReactionDetails` — as intended. Added a regression test passing `"not-a-real-id"`.

---

apps/convex/functions/prayers/reactions.ts
```diff
@@ -0,0 +1,305 @@
...
+import { query, mutation, type QueryCtx } from "../../_generated/server";
+import type { Doc, Id } from "../../_generated/dataModel";
+import { requireAuth } from "../../lib/auth";
+import { now } from "../../lib/utils";
+
+/**
+ * Curated reaction set for prayers — a warm, tight 6, narrower than chat's 14
+ * because prayer is a more tender context than group chat. This is the single
+ * source of truth for both the picker bar (frontend imports it) and the
+ * server-side allowlist. Adjusting the product choice is a one-line edit here.
```
lilseyi, line 28: Non-blocking (maintainability): this comment says the list is "the single source of truth for both the picker bar (frontend imports it) and the server-side allowlist," but the frontend does **not** import this constant — `PrayerReactions.tsx` hardcodes its own duplicate copy of the six emojis. The two are byte-identical today (variation selectors included), so there's no live bug, but nothing pins them together: editing one list without the other would silently ship a picker that offers an emoji the server rejects (or hides a curated one). Suggest either (a) correcting this comment to say "kept in sync with the frontend copy," and/or (b) adding a tiny test that imports both arrays and asserts equality so drift fails CI.

---

apps/convex/__tests__/prayers/reactions.test.ts
```diff
@@ -0,0 +1,567 @@
...
+      api.functions.prayers.reactions.getReactionDetails,
+      {
+        token: authorToken,
+        targetType: "prayer",
+        targetId: "not-a-real-id",
+        emoji: HEART,
+      },
+    );
+    expect(reactors).toEqual([]);
+  });
```
lilseyi, line 528: Non-blocking (test coverage): `getReactionDetails` is the most privacy-sensitive surface in the PR — it returns reactors' **real display names**, unlike the "First L." masking the rest of the module is careful about. Its authorization is only exercised here via the *malformed-id* path (returns `[]`). There's no test that a valid target with an **ineligible viewer** (a member who hasn't prayed, a non-member, or a viewer of a later-rejected prayer) gets `[]`. The code is correct today because it shares `resolveAccessiblePrayer`, but that coupling is by convention only — a future refactor that split or loosened the query's gate would leak the reactor roster with no test failing. Worth one test asserting a bystander gets `[]` from `getReactionDetails` on a real prayer (and ideally one against a `followUp` target, which is never tested).

---

apps/mobile/features/prayer/components/PrayerReactions.tsx
```diff
@@ -0,0 +1,257 @@
...
+    [toggleReaction, targetType, targetId],
+  );
+
+  const openPicker = useCallback(() => {
+    const node = addButtonRef.current;
+    if (node) {
+      node.measureInWindow((x, y) => {
+        // Anchor the bar just above the ＋ button, nudged left so the 6-emoji
+        // row stays on-screen.
+        setPickerPos({ top: Math.max(y - 60, 40), left: Math.max(x - 120, 12) });
```
lilseyi, line 102: Non-blocking (UX nit): the picker anchor only clamps the top and left edges (`Math.max(y-60,40)` / `Math.max(x-120,12)`). There's no clamp against the right or bottom edge, so a ＋ button near the right edge of a narrow screen can push the ~6-emoji bar partly off-screen — and the centered fallback only triggers when `measureInWindow` fails, not when it succeeds but lands off-screen. Low risk given the fixed bar width, but worth clamping `left` against screen width (and it can't be verified here without a simulator).

---

lilseyi (commented):
**Automated review — prayer emoji reactions (mode: review, high risk)**

Reviewed the diff against the approved spec with parallel correctness / security / spec-fidelity / test subagents, then adversarially verified each finding. **Verdict: approved — no blocking findings.**

The feature faithfully mirrors the existing chat reactions pattern applied to a second surface:
- **Correctness ✓** — polymorphic `targetId` guarded by `normalizeId`, toggle idempotent under Convex's serializable mutations, `(detail as any).isAnonymous` degrades safely (author-only screen), emoji bytes match frontend↔backend.
- **Security ✓** — `resolveAccessiblePrayer` faithfully reproduces `getDetail`'s gate and is actually *stricter* (adds `moderationStatus === "approved"` for non-authors); `communityId` is derived from the prayer, not user input; the anonymity guard (`isAnonymous && author === caller → throw`) is sound and `isAnonymous` is immutable, so an anonymous author can never surface in the reactor list. Cross-community `targetId` fails the membership lookup. No substantiated vulnerabilities.
- **Spec-fidelity ✓** — reactions on both detail screens (request + every follow-up), exactly the 6 curated emojis, count hidden at 1, own reaction highlighted, ＋ picker, long-press "who reacted", single-query fold into `getDetail` (no N+1), no push on react, deck untouched, onboarding guide updated.
- **Tests ✓** — 13/13 pass locally; solid coverage of toggle, allowlist, access control, cross-user aggregation, and both anonymity guards.

Three **non-blocking** notes posted inline (a drifting-emoji-list maintainability risk, a test-coverage gap on `getReactionDetails`' authz-denial path, and a picker off-screen-clamp UX nit). None gate the merge.

---

PR #602, opened 2026-07-12
Show cross-team groups in the serving-mode inbox

lilseyi:
## What this fixes

**Serving Mode** is the focused, day-of view a volunteer switches into on the day they're serving — it replaces the normal Inbox with a slimmed-down list of just the group chats that matter for *today's* event.

A **cross-team group** is a special group chat whose members are pulled from *several* serving teams at once (e.g. a "Service Leads" chat that automatically includes the leaders of the Worship Team and the Tech Team). Its membership is auto-synced from who's rostered — nobody adds people by hand.

Today these cross-team groups **did not appear in the serving-mode inbox**, even when the volunteer belonged to one and it was tied to the event they were serving — leaving them without a chat they expected to see on the busiest part of their day. This PR makes the cross-team group show up as one more row in the serving inbox, styled exactly like the other team-chat rows. **No new UI** — a row that should have been there now is.

## Why it happened — two independent backend gaps, both closed

1. **Candidate never included** — `getInboxChannels` (`apps/convex/functions/messaging/channels.ts`) builds the inbox from the groups the user belongs to, then supplements it with channels the user is a member of but whose owning group they're *not* in. That supplement had branches for `event` and `isShared` channels but **none for `cross_team`**, so a same-group cross-team channel the volunteer wasn't a group member of never entered the candidate list — and the serving flatten (which only re-filters that list) couldn't bring it back. Added a `cross_team` branch **gated on serving mode**, so the normal (non-serving) inbox is left byte-for-byte unchanged.

2. **Allow-set was group-scoped** — `resolveServingChannelIds` (`apps/convex/functions/scheduling/serving.ts`) built the set of channels allowed into the serving inbox by scanning cross-team channels **scoped to the plan's group**. A cross-team channel can be created in one group while sourcing a team in *another* group, so such a cross-group channel was never in the allow-set and got filtered out. Now resolved **community-wide** by the channel's `crossTeamSync` selectors' `sourceTeamId` matching a team on the plan — mirroring the sibling `getServingUpcomingChannels` helper.

## Safety / edge cases

- **Membership gating unchanged** — the serving flatten only ever iterates channels the user has an active `chatChannelMembers` row for, so a non-member still can't see the chat.
- **Archived** cross-team channels stay hidden (existing `isArchived` guard).
- **No normal-inbox regression** — the new branch is gated on serving mode; outside serving mode a cross-team channel still falls through to the `isShared` branch exactly as before. Verified by the pre-existing normal-inbox test (which caught an early version of this change and now passes).

## Tests

Added serving-mode `getInboxChannels` coverage in `crossTeamChannels.test.ts`:
- same-group cross-team channel surfaced to a volunteer **not** in the owning group,
- cross-**group** cross-team channel surfaced (channel lives in another group),
- non-member negative case (does not see it).

All convex suites pass (865 tests) and type-check is clean; the mobile suite ran clean on push (1273 tests).

## Verification

This is a backend change to the inbox query; the tests drive the exact `getInboxChannels` serving-mode paths end-to-end for both topologies. The environment has no iOS simulator, so no device capture — the before/after of the row appearing matches the originator's attached mock (a rendered approximation of the serving-inbox layout, not a device screenshot).

Dashboard item: `x97dwheqxp6b6tg52f7tymdk198acrga`

---
_Generated by [Claude Code](https://claude.ai/code/session_01AM9v2CJy58xqSX5q6ns657)_

### DIFF
--- a/apps/convex/functions/messaging/channels.ts
+++ b/apps/convex/functions/messaging/channels.ts
@@ -1622,9 +1622,11 @@ export const getInboxChannels = query({
     // Find shared channels from user's memberships that aren't already in allChannels.
     // These are channels owned by other groups but shared with one of the user's groups.
     const allChannelIds = new Set(allChannels.map((ch) => ch._id));
-    // Track event channels picked up from chatChannelMembers so we can ensure
-    // their owning group appears in validGroups / groupRoleMap below.
-    const eventChannelsToInclude: Doc<"chatChannels">[] = [];
+    // Channels picked up from chatChannelMembers whose owning group the user is
+    // NOT in (event channels always; cross-team channels only in serving mode),
+    // so we can ensure that owning group appears in validGroups / groupRoleMap
+    // below and the grouping step picks the channel up.
+    const nonGroupChannelsToInclude: Doc<"chatChannels">[] = [];
     for (const membership of userChannelMemberships) {
       if (allChannelIds.has(membership.channelId)) continue;
       const candidateChannel = await ctx.db.get(membership.channelId);
@@ -1644,7 +1646,26 @@ export const getInboxChannels = query({
         const lastMessageAt = candidateChannel.lastMessageAt ?? 0;
         if (lastMessageAt <= 0) continue;
         if (lastMessageAt < inboxQueryNow - INBOX_EVENT_HIDE_AFTER_MS) continue;
-        eventChannelsToInclude.push(candidateChannel);
+        nonGroupChannelsToInclude.push(candidateChannel);
+        continue;
+      }
+
+      // Cross-team channels in SERVING MODE: membership is auto-synced from
+      // serving-role assignments (a rostered volunteer gets a chatChannelMembers
+      // row without being a member of the channel's home group), so a same-group
+      // channel the volunteer isn't a group member of never arrives via the
+      // group-owned fetch above. When it's one of this plan's resolved serving
+      // channels, pull it into the candidate list so the serving flatten can
+      // surface it. This branch is gated on serving mode (`servingChannelIds`)
+      // so the normal inbox is left byte-for-byte unchanged — outside serving
+      // mode a cross-team channel still falls through to the isShared branch
+      // below exactly as before. The flatten still filters by servingChannelIds
+      // and this loop only ever sees channels the user has an active membership
+      // row for, so nothing leaks to a non-member.
+      if (candidateChannel.channelType === "cross_team" && servingChannelIds) {
+        if (servingChannelIds.has(candidateChannel._id)) {
+          nonGroupChannelsToInclude.push(candidateChannel);
+        }
         continue;
       }
 
@@ -1680,11 +1701,11 @@ export const getInboxChannels = query({
       }
     }
 
-    // Event channels: surface every enabled event channel the user is a member
-    // of, even when they're not a member of the owning group. We fetch the
-    // owning group (and its group type) on demand and add it to validGroups
-    // so the normal grouping step picks the channel up.
-    for (const eventChannel of eventChannelsToInclude) {
+    // Surface each collected channel (enabled event channels always; serving
+    // cross-team channels) even when the user isn't a member of the owning
+    // group. We fetch the owning group (and its group type) on demand and add
+    // it to validGroups so the normal grouping step picks the channel up.
+    for (const eventChannel of nonGroupChannelsToInclude) {
       if (!eventChannel.groupId) continue; // Skip ad-hoc channels (DM/group_dm)
       const ecGroupId = eventChannel.groupId;
       const owningGroup = allGroups.find((g) => g && g._id === ecGroupId)
--- a/apps/convex/functions/scheduling/serving.ts
+++ b/apps/convex/functions/scheduling/serving.ts
@@ -94,18 +94,19 @@ export async function resolveServingChannelIds(
   // (c) Cross-team channels that draw from any team on this plan. Their
   // membership is auto-synced (syncSource "event_plan") from the same
   // roleAssignments, so a rostered member belongs in the serving inbox too.
-  // Scoped to the plan's group via `by_group_type` (cross-team channels live
-  // in the same group as the plan's teams) to avoid a full-table scan. This is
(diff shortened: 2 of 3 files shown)


apps/convex/functions/messaging/channels.ts
```diff
@@ -1644,7 +1646,26 @@ export const getInboxChannels = query({
...
+      // channel the volunteer isn't a group member of never arrives via the
+      // group-owned fetch above. When it's one of this plan's resolved serving
+      // channels, pull it into the candidate list so the serving flatten can
+      // surface it. This branch is gated on serving mode (`servingChannelIds`)
+      // so the normal inbox is left byte-for-byte unchanged — outside serving
+      // mode a cross-team channel still falls through to the isShared branch
+      // below exactly as before. The flatten still filters by servingChannelIds
+      // and this loop only ever sees channels the user has an active membership
+      // row for, so nothing leaks to a non-member.
+      if (candidateChannel.channelType === "cross_team" && servingChannelIds) {
```
lilseyi, line 1665: **Non-blocking (test coverage).** This `&& servingChannelIds` gate is exactly what keeps the normal (non-serving) inbox byte-for-byte unchanged — good. But nothing in the new tests locks that guarantee in for the *synced-but-not-a-group-member* topology. The existing normal-inbox tests only cover the cross-group `isShared` case where the volunteer **is** a group member; test #1's `techUserId` (member via auto-sync only, not in the owning group) is never asserted against the *normal* inbox.

If someone dropped the `&& servingChannelIds` guard here, a same-group cross-team channel would start leaking into the **normal** inbox for that synced-only volunteer, and no test would catch the regression. Consider one more assertion reusing the test #1 fixture: call `getInboxChannels` **without** `servingPlanId` for `techUserId` and expect the channel to be **absent**. Cheap, and it pins the one guarantee this branch exists to protect.

---

apps/convex/functions/messaging/channels.ts
```diff
@@ -1680,11 +1701,11 @@ export const getInboxChannels = query({
...
-    // Event channels: surface every enabled event channel the user is a member
-    // of, even when they're not a member of the owning group. We fetch the
-    // owning group (and its group type) on demand and add it to validGroups
-    // so the normal grouping step picks the channel up.
-    for (const eventChannel of eventChannelsToInclude) {
+    // Surface each collected channel (enabled event channels always; serving
+    // cross-team channels) even when the user isn't a member of the owning
+    // group. We fetch the owning group (and its group type) on demand and add
+    // it to validGroups so the normal grouping step picks the channel up.
+    for (const eventChannel of nonGroupChannelsToInclude) {
```
lilseyi, line 1708: **Nit (non-blocking).** The loop variable is still named `eventChannel`, but the collection it iterates (`nonGroupChannelsToInclude`) was just renamed away from `eventChannelsToInclude` precisely because it now also holds cross-team channels. Renaming this to `channel` / `nonGroupChannel` keeps the loop consistent with the rename and avoids implying these are all event channels.

---

apps/convex/functions/scheduling/serving.ts
```diff
@@ -94,18 +94,19 @@ export async function resolveServingChannelIds(
...
+  // membership-agnostic like the team-channel branch above: the serving flatten
+  // in messaging/channels.ts only iterates channels already in the per-user,
+  // active-membership-gated result, so widening this Set leaks nothing — only
+  // users with an active chatChannelMembers row actually see the channel.
   const crossTeamChannels = await ctx.db
     .query("chatChannels")
-    .withIndex("by_group_type", (q) =>
-      q.eq("groupId", plan.groupId).eq("channelType", "cross_team"),
-    )
+    .filter((q) => q.eq(q.field("channelType"), "cross_team"))
```
lilseyi, line 109: **Non-blocking observation (no change required).** Two small things worth flagging on this newly community-wide resolution:

1. This is now an unindexed full scan of `chatChannels` filtered to `cross_team`, executed on every serving-mode `getInboxChannels` (a hot, reactive query). It's intentional, mirrors the sibling `getServingUpcomingChannels`, and is what the spec asked for — so no change here — but its cost grows with total channel count rather than per-plan/group. Fine given cross-team channels are rare; just noting it for future scale.

2. `resolveServingChannelIds` reads the plan via a raw `ctx.db.get(planId)` with no caller/plan-membership assertion. The result is fully membership-gated downstream (the flatten only keeps channels the user has an active `chatChannelMembers` row for), so there's **no** leak today. An explicit plan-scope / `requireGroupMember` check at the `getInboxChannels` entry would be cheap defense-in-depth if you want belt-and-suspenders, but it's not required for this fix.

---

lilseyi (commented):
**Automated review — verdict: approved (no blocking findings).**

Reviewed the diff against the approved spec with four parallel lenses (correctness, security, spec-fidelity, tests), then adversarially verified each candidate finding. Summary:

- **Correctness ✅** — the surfaced `cross_team` candidate provably survives every downstream filter (group-access gate, `isCustomChannel`/leader-enabled gate — which don't apply to `cross_team`, shared-dedup, and the serving flatten). Unread counts / last-message come through the normal pipeline. The `resolveServingChannelIds` selector match (`sourceTeamId` only, vs `getServingUpcomingChannels` also matching `roleId`) is a safe superset because membership gating — synced via a role-aware sync — is the real gate.
- **Security ✅** — no data leak. `servingChannelIds` is only ever used to *filter down* an already-membership-gated list (`userChannelIds`, active `chatChannelMembers` rows); it never adds a channel the user isn't a member of. Community boundary re-checked at the owning-group step.
- **Spec-fidelity ✅** — both backend gaps closed exactly as specified, backend-only, no scope creep; the "normal inbox unchanged" guarantee holds (branch gated on `servingChannelIds`). No guide-doc update warranted (surfaces a pre-existing chat in a pre-existing list; no UI/label/flow change).
- **Tests ✅** — the same-group and cross-group tests each fail against reverted prod code, so they genuinely prove both fixes.

Three **non-blocking** inline notes posted: a worthwhile normal-inbox regression-guard test to add, a cosmetic loop-variable rename, and an optional defense-in-depth/perf observation on the community-wide scan. None block merge.

---

PR #611, opened 2026-07-14
fix(chat): make thread-reply echo read as a reference, not a duplicate

lilseyi:
## What this fixes

Two related chat glitches reported on the dev dashboard (dev-assistant bug `x979q1acywv2aw2hx1tvpg2kq18ag4qs`, reported by Seyi Olujide):

**1. Thread replies looked like the message was posted twice.** When a message has a thread reply, the timeline floats a small echo of the original message (`GhostThreadPointer`) at the thread's latest-activity slot. That echo was drawn with the *exact same* styling as a real message — full-strength bubble, same author header, same text size — so it read as an accidental duplicate.

**2. The "Replying to …" bar went blank** when replying to a message you'd *just sent*. A just-sent message is still optimistic; its id is a synthetic `optimistic-…` string the server query can't resolve, so the fetched preview came back empty (no name, no snippet).

Both are mobile-only, look-and-feel + local-data changes. No Convex schema/function changes.

## What changed

**`GhostThreadPointer.tsx`** — the echo now reads as a quiet reference to an older message:
- Bubble + avatar dimmed to ~60% opacity (opacity-only, so it stays theme-safe across all four themes — no hard-coded light-theme color).
- Preview text shrunk 14 → 12.5, bubble padding tightened, avatar 24 → 20 and name shrunk/muted — physically smaller than a real bubble.
- Added a muted `↪ Original message` label so its purpose is explicit.
- All behavior preserved: tap-to-jump-to-original, the "N replies" pill opens the thread, own-vs-other alignment, and the deleted/attachment-only placeholder treatment.

**`ConvexChatRoomScreen.tsx`** — the reply bar is never blank:
- Capture the tapped message's already-in-hand `content`/`senderName` at reply time (`replyToLocal`) and use them as the immediate fallback in the "Replying to …" preview.
- Skip the parent fetch entirely for optimistic ids (it could only ever return empty). Real messages still fetch, so edits/latest content override the local snapshot.

## Verification

- `GhostThreadPointer` tests extended for the new label (own/other/deleted cases); full suite green: **138 suites, 1292 tests passing** (ran via the pre-push hook).
- Type-check: **0 new errors** introduced (pre-existing monorepo `@togather/shared` resolution errors are unchanged at 93 on both base and this branch).
- No iOS simulator is available on this runner, so the screenshot below is a **rendered mock built from the component's actual style values** (not a device capture), shown in light and dark to confirm the dimmed echo stays legible:

[after — rendered mock, light & dark](https://images.togather.nyc/dev-assistant/253bd7b4-afbb-4906-ae74-2c953fc0e439-ghost-after-rendered.png)

---
_Generated by [Claude Code](https://claude.ai/code/session_01KexsjV9krBBXhygKBhuFWk)_


apps/mobile/features/chat/components/GhostThreadPointer.tsx
```diff
@@ -174,7 +195,7 @@ export function GhostThreadPointer({
             alignRight ? styles.bubbleOwn : styles.bubbleOther,
             {
               backgroundColor: alignRight ? colors.chatBubbleOwn : colors.chatBubbleOther,
-              opacity: pressed ? 0.7 : 1,
+              opacity: pressed ? DIM_OPACITY * 0.8 : DIM_OPACITY,
```
lilseyi: **Blocking (spec fidelity — dark themes): the 0.6 bubble opacity double-dims the deleted/attachment placeholder text into illegibility.**

`opacity: DIM_OPACITY` sits on the `<Pressable>`, and RN opacity composites onto all children — including the placeholder `<Text>` (deleted / "📷 Photo") whose color is already `colors.textTertiary` (line ~139). So placeholder text renders at *textTertiary × 0.6 over the bubble*, exactly the "double-dimming into illegibility" the spec explicitly warned against.

Measured effective text-vs-bubble contrast (WCAG):
- **Dark theme** (`textTertiary #667781` on `chatBubbleOther #1f2c34`, composited at 0.6 over background `#0b141a`): drops from **~3.1:1 → ~1.9:1**.
- **Knicks-dark**: `#6e8aa8` on `#01294d` → **~2.3:1**.

Both land well below a readable floor for the deleted/attachment echo in the two dark themes — the `DIM_OPACITY` doc comment claims it "stays legible," but the math disagrees. Normal (non-placeholder) text is fine because `chatBubbleOwnText/OtherText` are near-white/near-black and survive 0.6.

Fix: dim the bubble background + normal text only, not the placeholder — e.g. keep the 0.6 on the background but lift the placeholder `<Text>` out from under it (its own compensating opacity, or `textSecondary`), so the deleted/attachment case keeps its pre-PR contrast.

---

apps/mobile/features/chat/components/ConvexChatRoomScreen.tsx
```diff
@@ -514,11 +514,25 @@ const ConvexChatRoomScreenInner: React.FC = () => {
...
+  //
+  // A just-sent message is still optimistic: its id is a synthetic
+  // `optimistic-…` string the server query can't resolve, so we skip the fetch
+  // for those ids (it would only ever come back empty) and rely on the local
+  // preview captured at reply time. Real messages still fetch as before so
+  // edits/latest content override the local snapshot.
+  const isOptimisticReplyTarget =
+    typeof replyToMessageId === "string" && replyToMessageId.startsWith("optimistic-");
+  const { message: replyParentMessage } = useParentMessage(
+    isOptimisticReplyTarget ? null : replyToMessageId
```
lilseyi, line 534: **Non-blocking (pre-existing, but it undermines this fix's goal — worth a heads-up):** this PR makes the "Replying to …" banner render correctly for a just-sent optimistic message, but *sending* that reply still fails end-to-end.

`replyToMessageId` holds the synthetic `optimistic-…` string, which flows through `MessageInput`'s `replyToMessage._id` → `parentMessageId` into the `sendMessage` mutation, whose validator is `parentMessageId: v.optional(v.id("chatMessages"))` (`messaging/messages.ts:659`). A non-Id `optimistic-…` value is rejected with an `ArgumentValidationError`, so the reply lands in `error` status.

Repro: send a message → immediately long-press it → Reply → type → Send. The banner now looks correct (this PR), but the send throws. The offending send line predates this PR and the spec scoped the fix to the banner only, so this isn't a regression and doesn't block on spec terms — but the banner fix makes the flow *look* functional when it isn't, so please confirm this is knowingly deferred (a follow-up could resolve `replyToMessageId` to the real id once the optimistic parent settles, or disable Send while the parent is unresolved).

---

apps/mobile/features/chat/components/ConvexChatRoomScreen.tsx
```diff
@@ -1343,8 +1369,13 @@ const ConvexChatRoomScreenInner: React.FC = () => {
                   replyToMessageId
                     ? {
                         _id: replyToMessageId,
-                        content: replyParentMessage?.content ?? "",
-                        senderName: replyParentMessage?.senderName ?? "",
+                        // Prefer the fetched parent (authoritative, reflects
+                        // edits) but fall back to the locally-captured values so
+                        // the banner is never blank — e.g. when replying to a
+                        // just-sent optimistic message the server can't resolve.
+                        content: replyParentMessage?.content ?? replyToLocal?.content ?? "",
```
lilseyi: **Non-blocking (test coverage):** the trickier half of this PR — the optimistic-id guard and the two-level `replyParentMessage?.x ?? replyToLocal?.x ?? ""` fallback — ships with no test, while the simpler `GhostThreadPointer` label change gets three. Given `riskLevel: medium`, a `ConvexChatRoomScreen` test for (a) optimistic id → skips the fetch, uses `replyToLocal`; (b) real id → fetched value overrides the local snapshot; (c) cancel/tab-switch clears `replyToLocal`, would lock in the behavior. The harness already supports it — `LeaderChatNavigation.test.tsx` renders this screen with `useQuery`/`useStoredAuthToken` mocked, so `useParentMessage` can be driven to `undefined`/`null` to simulate the unresolved-optimistic case. Recommend adding it; not a merge blocker on its own.

---

lilseyi (commented):
Automated review (dev-assistant, `changes_requested`). Reviewed the diff against the approved spec with parallel correctness / security / spec-fidelity / test subagents, then adversarially verified each finding.

**The reply-banner fix and most of the `GhostThreadPointer` restyle are correct and faithful** — opacity within the 55–65% band, label added, text/padding shrunk, tap-to-jump, "N replies" pill, own-vs-other alignment, and the deleted/attachment placeholders all preserved; the optimistic fallback resets on every reply-clear path; no security issues.

**One blocking item** (inline): the 0.6 bubble opacity composites onto the *placeholder* text (deleted / "📷 Photo"), dropping its contrast to ~1.9:1 in dark theme and ~2.3:1 in Knicks-dark — the exact "double-dim into illegibility" the spec called out. Dim the bubble background/normal text but keep the placeholder legible.

Two non-blocking notes inline: (1) replying to a just-sent optimistic message fixes the banner but the *send* still fails validation (pre-existing; please confirm it's knowingly deferred), and (2) the reply-banner logic ships untested despite reusable screen-test infra.

_Formal request-changes isn't available to this automated reviewer; the changes-requested verdict is recorded on the dashboard._

---

apps/mobile/features/chat/components/GhostThreadPointer.tsx
lilseyi: Fixed in d4ca84a. Moved the dim off the `Pressable`'s `opacity` (which composited onto the placeholder `Text`) and onto the bubble **background** alpha (new `withAlpha` helper) plus the **normal** text only. The deleted/attachment placeholder now renders at full `textTertiary` strength.

Re-measured effective placeholder contrast with the same method:

| theme | pre-PR | this PR (opacity bug) | after fix |
| --- | --- | --- | --- |
| dark | 3.08 | 1.92 | **3.48** |
| knicks-dark | 4.11 | 2.35 | **4.43** |
| light | 2.27 | 1.55 | **2.48** |
| knicks-light | 2.23 | 1.57 | **2.27** |

Placeholder is restored to at/above the pre-PR baseline in every theme. Added a `GhostThreadPointer` regression test asserting the placeholder carries no reduced opacity while normal text does. Rendered before/after mock (dark + knicks-dark) is in the PR comments.

---

apps/mobile/features/chat/components/ConvexChatRoomScreen.tsx
lilseyi: Added in d4ca84a. Rather than stand up the full `ConvexChatRoomScreen` harness, I extracted the two pieces of logic into a pure `replyPreview` helper (`isOptimisticMessageId` + `resolveReplyPreview`) and unit-tested them directly — covering (a) optimistic id → skips the fetch, uses the local snapshot; (b) real id → fetched value overrides local; (c) neither available → empty strings, never `undefined`; plus per-field fallback. The screen now imports these instead of the inline `??` chains, so the tested code path is the shipped one.

---

apps/mobile/features/chat/components/ConvexChatRoomScreen.tsx
```diff
@@ -514,11 +514,25 @@ const ConvexChatRoomScreenInner: React.FC = () => {
...
+  // A just-sent message is still optimistic: its id is a synthetic
+  // `optimistic-…` string the server query can't resolve, so we skip the fetch
+  // for those ids (it would only ever come back empty) and rely on the local
+  // preview captured at reply time. Real messages still fetch as before so
+  // edits/latest content override the local snapshot.
+  const isOptimisticReplyTarget =
+    typeof replyToMessageId === "string" && replyToMessageId.startsWith("optimistic-");
+  const { message: replyParentMessage } = useParentMessage(
+    isOptimisticReplyTarget ? null : replyToMessageId
+  );
```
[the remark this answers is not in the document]
lilseyi: Knowingly deferred — this is out of scope for the approved spec, which is explicitly a "look-and-feel + local-data change only. No Convex schema, function, or `parentMessageId` model changes." The spec's edge-case list also anticipates this exact case: *"Reply to an optimistic message that then fails to send: the reply bar should still show the local preview; if the parent send fails the user can cancel the reply as normal (no crash, no stuck blank bar)."*

I verified the failure is graceful, not a crash: the `optimistic-…` id doesn't match any known soft-fail string in `classifyChatSendError`, so it falls through to `unknown` → **hard fail**, which keeps the optimistic row in `_status: "error"` (retry/dismiss available). No `ArgumentValidationError` reaches the user as an Alert, and it doesn't hit the navigator update-loop path (that's the soft-fail branch).

Mapping the synthetic id to the real row would mean threading the server id returned by `sendMessage` back into the reply state — a change to the send hook's contract that the spec deliberately keeps out. Leaving as-is per spec.

---

apps/mobile/features/chat/components/ConvexChatRoomScreen.tsx
lilseyi: Confirming this is knowingly deferred. The approved spec scopes the fix to the banner only ("look-and-feel + local-data change, no Convex function / `parentMessageId` model changes") and its edge-case list explicitly accepts the failed send: *"if the parent send fails the user can cancel the reply as normal (no crash, no stuck blank bar)."*

I checked the failure path is graceful: the `optimistic-…` value fails the `v.id("chatMessages")` validator → `unknown`/hard-fail in `classifyChatSendError` → the row stays in `error` (dismissable), no Alert, no navigator loop. So this PR's banner fix doesn't introduce a crash — it just makes the (still-failing) send *look* right, which matches the spec's stated intent.

The proper follow-up (resolve `replyToMessageId` to the real id once the optimistic parent settles, or block Send while unresolved) touches the send-hook contract and should be its own spec'd item. Filed mentally as follow-up; not changing it here to stay within scope.

---

apps/mobile/features/chat/components/ConvexChatRoomScreen.tsx
```diff
@@ -514,11 +515,24 @@ const ConvexChatRoomScreenInner: React.FC = () => {
...
-  const { message: replyParentMessage } = useParentMessage(replyToMessageId);
+  //
+  // A just-sent message is still optimistic: its id is a synthetic
+  // `optimistic-…` string the server query can't resolve, so we skip the fetch
+  // for those ids (it would only ever come back empty) and rely on the local
+  // preview captured at reply time. Real messages still fetch as before so
+  // edits/latest content override the local snapshot.
+  const isOptimisticReplyTarget = isOptimisticMessageId(replyToMessageId);
+  const { message: replyParentMessage } = useParentMessage(
+    isOptimisticReplyTarget ? null : replyToMessageId
```
lilseyi, line 534: Non-blocking (test coverage): The core bug-fix wiring here — `isOptimisticReplyTarget` gating the parent fetch, `handleMessageReply` capturing `localPreview`, and the `...resolveReplyPreview(replyParentMessage, replyToLocal)` spread into the banner object (line ~1376) — has no end-to-end test. The pure helpers in `replyPreview.ts` are well-covered in isolation, but nothing asserts they're actually plumbed together so that replying to a just-sent optimistic message yields a non-blank banner (the exact scenario this PR fixes). A small component/integration test around the reply banner would lock the regression in. Not a merge blocker given the helper coverage.

---

apps/mobile/features/chat/components/__tests__/GhostThreadPointer.test.tsx
```diff
@@ -142,6 +189,41 @@ describe('GhostThreadPointer', () => {
...
+        originalContent=""
+        originalSenderId={SOMEONE_ELSE}
+        senderName="Samuel Baker"
+        attachments={[{ type: 'image' }]}
+      />,
+    );
+
+    const placeholderStyle = flatten(getByText('📷 Photo').props.style);
+    // No dimming opacity on placeholder text (undefined or a full 1).
+    expect(placeholderStyle.opacity ?? 1).toBe(1);
```
lilseyi, line 210: Non-blocking (nit): This test's comment frames the regression as parent-`Pressable`-`opacity` compositing onto children, but the assertion only inspects the placeholder `<Text>`'s own flattened style (`opacity ?? 1 === 1`). If someone reintroduced `opacity` on the parent `Pressable`, this test would still pass — the dimming would live on the parent, not the Text's style. It does correctly guard the `isPlaceholder ? placeholderText : { opacity }` ternary, so it's not useless, just weaker than the comment implies. Consider either softening the comment or additionally asserting the `Pressable` carries no `opacity` prop.

---

lilseyi (commented):
**Automated review — dev-assistant (review mode), medium risk.** Reviewed the diff against the approved spec across four lenses (correctness, security, spec-fidelity/UX, tests), each finding adversarially verified.

**Verdict: approved — no blocking findings.**

- **Correctness:** Clean. The optimistic-id skip (`optimistic-…` prefix matches `useConvexSendMessage`), the `resolveReplyPreview` fallback order (`fetched ?? local`), and all preserved behaviors (tap-to-jump, N-replies pill, own/other alignment, deleted/attachment placeholder) hold up. `withAlpha` is safe — all four themes define `chatBubbleOwn/Other` as 6-digit hex, and it only ever receives theme tokens, never message content.
- **Spec-fidelity/UX:** Every "Done when" item met; dimming (0.6) sits in the ≈55–65% band; label, sizing, and padding match. No scope creep (no schema/function/`parentMessageId` changes). Edge cases (deleted, own-side, attachment-only, optimistic-then-failed, all four themes) handled. Dimming the bubble *background alpha* rather than a parent `opacity` is a deliberate, spec-permitted choice that satisfies the "don't double-dim the placeholder into illegibility" requirement.
- **Security:** No findings — UI/local-state only, no untrusted value reaches `withAlpha`.
- **Tests:** Pure helpers well-covered and regression-sensitive. Two non-blocking notes left inline: the ConvexChatRoomScreen wiring has no end-to-end test, and the "not double-dimmed" test is slightly weaker than its comment implies.

Note: this is a review-bot comment, not a formal GitHub approval — routine sessions can't submit approve/request-changes. The approved verdict reaches the dashboard via callback.

---

Issue #14, opened 2026-03-05
Bug: Role preview truncated when filtering by position during channel creation


dwalkerjr7:
## What happened?

During new channel creation, when selecting roles via the "filter by position" option, the preview list is truncated. It only shows a portion of the matching roles instead of the full list.

## Steps to reproduce

1. Start creating a new channel
2. Use the "filter by position" option to select roles
3. The preview list appears but is cut off
4. You can't see all the matching roles

## What should have happened?

The role preview should show the full list of matching roles, either by scrolling or expanding to fit all results.

## Screenshots

![image](https://github.com/user-attachments/assets/1c39e464-de38-4bb4-850b-76ab32fe25a1)

## Device

Unknown

## Frequency

Every time

PR #618, opened 2026-07-14
Pick which events a "Collect availability" link asks about

lilseyi:
## What & why

Tapping **Collect availability** in the rostering grid's ⋯ menu used to instantly build a shareable link covering **every** upcoming event and open the OS share sheet. There was no way to say "only ask about these two Sundays."

Now tapping **Collect availability** opens a small picker sheet listing the group's upcoming events, each with a checkbox. **All upcoming events start checked** (so the default matches today's behavior). The leader unchecks any they don't want, then taps **Share link · N events** — which creates the link for exactly those events and opens the share sheet as before.

Addresses dashboard item / issue #617 (reported by **Seyi Olujide** / @lilseyi).

## Before / after

The mock below is a **rendered mock built from the component's real modal styles** — not a device screenshot (this CI runner has no iOS simulator).

[Before / after — Collect availability event picker](https://images.togather.nyc/dev-assistant/3c5f8d99-df0f-4bff-b99f-c9d8b3ad7286-availability-picker-mock.png)

## Scope: frontend-only

The backend already supports this. `createAvailabilityLink` (`apps/convex/functions/scheduling/publicAvailability.ts`) already takes an optional `planIds` array, validates each belongs to the group, and snapshots exactly those events onto the request. Every downstream reader — the chat card (`getAvailabilityRequestByToken`) and the public `/a/` page (`getPublicAvailabilityRequest`) — already renders only the request's snapshotted `planIds`. The only gap was that the mobile button never passed `planIds`. **No backend, schema, or migration work.**

## Changes

`apps/mobile/features/scheduling/components/RosterGridScreen.tsx`:
- New `AvailabilityPicker` sheet (mirrors the existing `PublishMenu` inside `ModalShell`) — checkbox rows for each upcoming event, seeded all-checked, with a primary **Share link · N events** action.
- `handleShareLink` now takes the picked `planIds` and passes them to `createAvailabilityLink({ groupId, planIds })`.
- New `handleCollectAvailability` opens the picker; when the group has **no upcoming events** it keeps the existing "add an event first" error instead of showing an empty picker.
- `upcomingEvents` derives the list from the in-scope `events` (`eventDate >= today`, sorted ascending — matching the backend's snapshot order).

`apps/mobile/features/scheduling/components/__tests__/AvailabilityPicker.test.tsx`:
- Unit tests: all events checked by default; sharing passes exactly the checked ids in order; **Share** is disabled and shows a `0` count when nothing is checked.

## Legacy links

No migration. Existing links already snapshotted their event set at creation time, and the readers render that stored snapshot — so old links keep showing the events they were created with, unchanged.

## Verification

- `AvailabilityPicker.test.tsx` — 3 new tests pass.
- Full mobile suite green via the pre-push hook (140 suites, 1323 passed).
- Lint clean on the changed file (0 errors).
- No iOS simulator on the runner, so the UI is shown as a rendered mock above (built from the real `ModalShell` / `menuRow` / `addBtn` styles).

## Done when

- [x] Tapping **Collect availability** opens a picker listing upcoming events, each with a checkbox, all checked by default.
- [x] Leader can check/uncheck individual events.
- [x] **Share link** is disabled when nothing is checked and shows the selected count.
- [x] Sharing creates a link whose `planIds` are exactly the checked events, then opens the share sheet.
- [x] The resulting `/a/` link (chat card + public page) shows only the picked events (already handled by the snapshot readers — unchanged).
- [x] Links created before this change are unaffected.
- [x] Groups with no upcoming events still show the existing "add an event first" error.

Co-authored-by: Seyi Olujide <lilseyi@users.noreply.github.com>

https://claude.ai/code/session_011UxgmNy65AESKCVxTjNEJN

---
_Generated by [Claude Code](https://claude.ai/code/session_011UxgmNy65AESKCVxTjNEJN)_

### DIFF
--- a/apps/mobile/features/scheduling/components/RosterGridScreen.tsx
+++ b/apps/mobile/features/scheduling/components/RosterGridScreen.tsx
@@ -199,6 +199,13 @@ function monthDay(ms: number): string {
   });
 }
 
+/** Local midnight today — the cutoff for "upcoming" events (mirrors the backend). */
+function startOfTodayMs(): number {
+  const d = new Date();
+  d.setHours(0, 0, 0, 0);
+  return d.getTime();
+}
+
 /**
  * Web-only right-click prop bag. RN-Web forwards unknown DOM props onto the
  * host node, so `onContextMenu` lands on the underlying <div>. The RN types
@@ -515,6 +522,10 @@ export function RosterGridScreen() {
   // the in-flight states for its async actions + the "＋ Add date" column.
   const [overflowOpen, setOverflowOpen] = useState(false);
   const [sharingLink, setSharingLink] = useState(false);
+  // Event picker for "Collect availability": lists the group's upcoming events
+  // with checkboxes so the leader chooses exactly which ones the link asks
+  // about, before the OS share sheet opens.
+  const [availPickerOpen, setAvailPickerOpen] = useState(false);
   const [creatingEvent, setCreatingEvent] = useState(false);
   const [settingUp, setSettingUp] = useState(false);
 
@@ -906,30 +917,61 @@ export function RosterGridScreen() {
   // rostering home now, so these actions live here).
   // -------------------------------------------------------------------------
 
-  // Generate a public, app-optional availability link and hand it to the OS
-  // share sheet. People open it in a browser (no app needed) and their
-  // response is matched to their account when they later sign up.
-  const handleShareLink = useCallback(async () => {
-    if (sharingLink) return;
-    setSharingLink(true);
-    try {
-      const { publicToken } = await createAvailabilityLink({ groupId });
-      const url = DOMAIN_CONFIG.availabilityLinkUrl(publicToken);
-      await Share.share({
-        message: `Let us know when you can serve: ${url}`,
-      });
-    } catch (e) {
-      const err = e as { data?: { message?: string }; message?: string };
+  // The group's upcoming events (today onward), soonest first — the picker's
+  // list and the snapshot order the backend uses for a link's events.
+  const upcomingEvents = useMemo(
+    () =>
+      events
+        .filter((e) => e.eventDate >= startOfTodayMs())
+        .sort((a, b) => a.eventDate - b.eventDate),
+    [events],
+  );
+
+  // Open the event picker for "Collect availability". If the group has no
+  // upcoming events, keep today's behavior: surface the "add an event first"
+  // error instead of an empty picker.
+  const handleCollectAvailability = useCallback(() => {
+    if (upcomingEvents.length === 0) {
       Alert.alert(
         "Couldn't create link",
-        err?.data?.message ??
-          err?.message ??
-          "Add an upcoming event plan first, then try again.",
+        "Add an upcoming event plan first, then try again.",
       );
-    } finally {
-      setSharingLink(false);
+      return;
     }
-  }, [sharingLink, createAvailabilityLink, groupId]);
+    setAvailPickerOpen(true);
+  }, [upcomingEvents]);
+
+  // Generate a public, app-optional availability link for the picked events and
+  // hand it to the OS share sheet. People open it in a browser (no app needed)
+  // and their response is matched to their account when they later sign up.
+  const handleShareLink = useCallback(
+    async (planIds: Id<"eventPlans">[]) => {
+      if (sharingLink) return;
+      setSharingLink(true);
+      try {
+        const { publicToken } = await createAvailabilityLink({
+          groupId,
+          planIds,
+        });
+        const url = DOMAIN_CONFIG.availabilityLinkUrl(publicToken);
+        setAvailPickerOpen(false);
+        await Share.share({
+          message: `Let us know when you can serve: ${url}`,
+        });
+      } catch (e) {
+        const err = e as { data?: { message?: string }; message?: string };
+        Alert.alert(
+          "Couldn't create link",
+          err?.data?.message ??
+            err?.message ??
+            "Add an upcoming event plan first, then try again.",
+        );
+      } finally {
+        setSharingLink(false);
+      }
+    },
+    [sharingLink, createAvailabilityLink, groupId],
+  );
 
   // Create a new draft plan at the neutral default date (next Sunday 9 AM, same
   // as the editor's default). The reactive rosterMatrix self-refreshes → a new
@@ -2307,7 +2349,7 @@ export function RosterGridScreen() {
           }}
           onCollectAvailability={() => {
             setOverflowOpen(false);
-            void handleShareLink();
+            handleCollectAvailability();
           }}
           onChangePastLimit={(delta) =>
             setPastLimit((n) => Math.max(0, Math.min(n + delta, MAX_PAST_DATES)))
@@ -2331,6 +2373,16 @@ export function RosterGridScreen() {
             void publishAllDrafts();
           }}
           onClose={() => setPublishMenuOpen(false)}
+        />
+      )}
+
+      {availPickerOpen && (
+        <AvailabilityPicker
+          events={upcomingEvents}
+          sharingLink={sharingLink}
+          colors={colors}
+          onShare={(planIds) => void handleShareLink(planIds)}
+          onClose={() => setAvailPickerOpen(false)}
         />
       )}
     </View>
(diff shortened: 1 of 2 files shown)


apps/mobile/features/scheduling/components/RosterGridScreen.tsx
```diff
@@ -3181,6 +3233,119 @@ function PublishMenu({
...
+  const toggle = useCallback((id: string) => {
+    setSelected((prev) => {
+      const next = new Set(prev);
+      if (next.has(id)) next.delete(id);
+      else next.add(id);
+      return next;
+    });
+  }, []);
+  const count = selected.size;
+  const disabled = count === 0 || sharingLink;
```
lilseyi: **Correctness (blocking): `count`/`disabled` read the stale seed set, not the current events.**

`selected` is seeded once by the `useState` initializer and never re-synced, but `AvailabilityPicker` stays mounted while its `events` prop (`upcomingEvents`, a reactive-query derivative) can change. `count = selected.size` is computed off the stale set, while the rendered rows (`events.map`) and the shared payload (`events.filter(e => selected.has(...))` at line 3313) are computed off the *current* `events`. They diverge whenever a selected event leaves the upcoming list (concurrent delete/reschedule by a co-leader, or an event aging past today's cutoff while the sheet is open).

- **Cosmetic case:** one selected event drops out → button still reads "Share link · 2 events" but only 1 id is shared.
- **Worst case:** if *all* currently-selected events leave the list, `selected.size` stays > 0, so `disabled = count === 0 || sharingLink` stays `false`. Tapping Share calls `onShare([])`, and in `handleShareLink` the backend's `args.planIds && args.planIds.length > 0` check is false, so it falls through to the "snapshot **all** upcoming events" branch — the opposite of "nothing selected," and a defeat of the spec's "Share is disabled when nothing is checked" guarantee.

One-line fix: derive count from the intersection with the live list, e.g.
```ts
const selectedIds = events.filter((e) => selected.has(e._id as string));
const count = selectedIds.length;
```
and share `selectedIds.map((e) => e._id)`.

---

apps/mobile/features/scheduling/components/RosterGridScreen.tsx
```diff
@@ -906,30 +917,61 @@ export function RosterGridScreen() {
...
-      const { publicToken } = await createAvailabilityLink({ groupId });
-      const url = DOMAIN_CONFIG.availabilityLinkUrl(publicToken);
-      await Share.share({
-        message: `Let us know when you can serve: ${url}`,
-      });
-    } catch (e) {
-      const err = e as { data?: { message?: string }; message?: string };
+  // The group's upcoming events (today onward), soonest first — the picker's
+  // list and the snapshot order the backend uses for a link's events.
+  const upcomingEvents = useMemo(
```
lilseyi, line 940: **Nit (non-blocking): the count can overstate what's actually shared when a group has >12 upcoming events.**

`upcomingEvents` filters + sorts but never caps, so the picker lists and counts all upcoming events. The backend `createAvailabilityLink` sorts asc and `.slice(0, MAX_EVENTS=12)`, silently dropping the latest ones. Order is preserved (both sort asc), so this is only a display-honesty issue: a leader with 15 upcoming events, all checked, sees "Share link · 15 events" while the link covers only the soonest 12. The spec explicitly delegates the cap to the backend, so this is within spec — flagging only in case you'd prefer to cap/hint the count so it's never misleading.

---

apps/mobile/features/scheduling/components/__tests__/AvailabilityPicker.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+        onClose={jest.fn()}
+      />,
+    );
+    fireEvent.press(getByLabelText("Sunday Service, Sun Jul 19"));
+    fireEvent.press(getByLabelText("Sunday Service, Sun Jul 26"));
+    // With nothing checked the Share button reports 0 and is disabled, so a tap
+    // can't create a link (the backend also rejects an empty set).
+    const shareBtn = getByLabelText("Share link for 0 events");
+    expect(shareBtn.props.accessibilityState.disabled).toBe(true);
+    expect(onShare).not.toHaveBeenCalled();
```
lilseyi, line 201: **Tests (non-blocking): this `onShare` assertion is vacuous.** The test presses the two checkboxes but never presses the Share button, so `expect(onShare).not.toHaveBeenCalled()` is trivially true and the actual press-guard (`if (disabled) return;` at RosterGridScreen.tsx:3311) is left unverified — which matters because RNTL's `fireEvent.press` doesn't reliably honor `disabled` on `Pressable` (the very reason that internal guard exists). Add `fireEvent.press(shareBtn)` before the assertion so the guard is actually exercised. The `accessibilityState.disabled === true` check above is good and does cover the render-level state.

Two other gaps worth a follow-up (not blocking): no case with `sharingLink={true}` covers the spinner + double-submit guard, and multi-id order preservation / re-checking a toggled-off event aren't exercised (Test 2's single-id result makes "in order" trivial).

---

lilseyi (commented):
## Automated review — Collect-availability event picker

Reviewed against the approved spec with parallel correctness / security / spec-fidelity / test lenses; every finding below survived an adversarial verify pass. **Verdict: changes requested** — one confirmed correctness issue, the rest non-blocking.

**Overall:** the happy path is correct, faithful to the spec (picker in `ModalShell`, all-checked default, "Share link · N events" with count, disabled-when-empty, `createAvailabilityLink({ groupId, planIds })`, no-upcoming-events keeps the existing error, no scope creep), and the 3 tests pass. Security is clean: the backend still enforces the group-scheduler gate and filters `p.groupId === args.groupId`, so client-supplied `planIds` can't leak another group's events.

**Blocking**
- **Stale selection state** (RosterGridScreen.tsx:3270) — `count`/`disabled` are derived from the once-seeded `selected` set instead of its intersection with the live `events` prop. If a selected event leaves the reactive upcoming list while the sheet is open, the count desyncs, and in the worst case (all selected events drop out) the Share button stays enabled and shares `[]`, which the backend interprets as "snapshot **all** upcoming events." One-line fix in the inline comment.

**Non-blocking (nits / follow-ups)**
- Count can overstate the shared set when a group has >12 upcoming events (backend caps at 12). Within spec — noted inline.
- Test 3's `onShare not called` assertion is vacuous (Share is never pressed); the `disabled` press-guard is unverified. Also no `sharingLink={true}` loading/double-submit coverage. Noted inline.
- Cleanup: the `sharingLink` spinner/disabled on the `OverflowMenu` "Collect availability" row (RosterGridScreen.tsx:3414) is now unreachable — the overflow closes (`setOverflowOpen(false)`) before `sharingLink` ever flips true, so that state lives in the picker's Share button now. Consider dropping the prop.

_Posted as inline comments + this summary; formal approve/request-changes isn't available to this session, so the verdict travels via the dashboard._

---

apps/mobile/features/scheduling/components/RosterGridScreen.tsx
```diff
@@ -906,30 +917,61 @@ export function RosterGridScreen() {
...
+    () =>
+      events
+        .filter((e) => e.eventDate >= startOfTodayMs())
+        .sort((a, b) => a.eventDate - b.eventDate),
+    [events],
```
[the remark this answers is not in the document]
lilseyi: Fixed. The picker now sources its upcoming-events list from `scheduling.events.listEvents` (upcoming-only, sorted ascending, uncapped) instead of the grid's `events`, which comes from `rosterMatrix` and is capped at `MAX_PAST_DATES` columns with past plans consuming the cap when "Previous dates" is set. The list is sliced to `MAX_AVAILABILITY_EVENTS` (12) to mirror the backend's `MAX_EVENTS` slice, so a group with 11+ upcoming plans no longer silently omits later dates and setting Previous dates no longer hides upcoming events.

---

apps/mobile/features/scheduling/components/RosterGridScreen.tsx
lilseyi: Fixed as suggested. `count`, `disabled`, and the shared ids now derive from the intersection with the live `events` prop:

```ts
const selectedIds = useMemo(
  () => events.filter((e) => selected.has(e._id as string)).map((e) => e._id),
  [events, selected],
);
const count = selectedIds.length;
```

`onShare(selectedIds)` shares that live set, so if every selected event leaves the list the count drops to 0 and Share disables — no more sharing `[]` (which the backend read as "all upcoming"). Added two tests that rerender with a shrunk / empty `events` prop to cover both the cosmetic desync and the worst case.

---

apps/mobile/features/scheduling/components/RosterGridScreen.tsx
lilseyi: Addressed alongside the P2 fix: `upcomingEvents` is now sliced to `MAX_AVAILABILITY_EVENTS` (12), matching the backend's `MAX_EVENTS` cap. The picker lists and counts exactly the events a link can cover, so the count can no longer overstate for a group with 12+ upcoming events.

---

apps/mobile/features/scheduling/components/__tests__/AvailabilityPicker.test.tsx
lilseyi: Fixed. The "nothing checked" test now presses the disabled Share button before asserting, so the `if (disabled) return;` press-guard is actually exercised. Also added the two follow-up cases you flagged: a `sharingLink={true}` case covering the spinner + double-submit guard, and a multi-id case asserting both ids share in ascending order (Test 2's single-id result no longer being the only "in order" coverage).

Note: pressing a disabled `Pressable` in this RNTL version routes the press up to `ModalShell`'s backdrop handler, which called `e.stopPropagation()` on an undefined event — I null-guarded that (`e?.stopPropagation?.()`) so the bubbled press can't throw.

---

apps/mobile/features/scheduling/components/RosterGridScreen.tsx
```diff
@@ -906,30 +925,68 @@ export function RosterGridScreen() {
...
+  // at MAX_PAST_DATES and can lead the grid with past plans when "Previous
+  // dates" is set — either of which would silently drop upcoming events from
+  // the link. Capped to the backend's MAX_AVAILABILITY_EVENTS so the picker
+  // never lists (or counts) more events than a link can actually cover.
+  const upcomingData = useAuthenticatedQuery(
+    api.functions.scheduling.events.listEvents,
+    groupId ? { groupId } : "skip",
+  ) as UpcomingEvent[] | undefined;
+  const upcomingEvents = useMemo(
+    () => (upcomingData ?? []).slice(0, MAX_AVAILABILITY_EVENTS),
```
lilseyi, line 941: Non-blocking, but worth flagging as a slight deviation from the spec ("no new cap is needed"). Slicing `upcomingData` to `MAX_AVAILABILITY_EVENTS` (12) means events 13+ are never *listable* in the picker. That's the right default for the "all checked" count (the old all-upcoming link already backend-sliced to the earliest 12, so a 13th event was never covered), so this is not a regression.

The one subtlety the new per-selection path introduces: the backend caps the *selected* set to the earliest 12, so a leader could legitimately want to share only event #14 — but with the client slicing to the earliest 12 before render, #14 can never be selected. This only affects groups with >12 upcoming events, so it's an acceptable trade for count integrity — just noting the picker is now strictly "earliest 12," not "the group's upcoming events."

---

apps/mobile/features/scheduling/components/__tests__/AvailabilityPicker.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+        colors={COLORS}
+        onShare={onShare}
+        onClose={jest.fn()}
+      />,
+    );
+    fireEvent.press(getByLabelText("Sunday Service, Sun Jul 19"));
+    fireEvent.press(getByLabelText("Sunday Service, Sun Jul 26"));
+    // With nothing checked the Share button reports 0 and is disabled. Actually
+    // press it: RNTL's fireEvent.press ignores `disabled` on Pressable, so this
+    // exercises the internal `if (disabled) return;` guard, not just the render.
```
lilseyi, line 122: Non-blocking (tests pass 7/7): this comment's claim is inverted, and it makes the following assertion tautological. On the pinned stack (RNTL 12.4 / jest-expo 54), `fireEvent.press` on a **disabled** `Pressable` does *not* invoke `onPress` at all — RNTL swallows it. So `fireEvent.press(shareBtn); expect(onShare).not.toHaveBeenCalled()` passes because RNTL never dispatches the press, **not** because the component's internal `if (disabled) return;` guard ran — that line is never actually exercised by any test. The load-bearing assertion here is the preceding `accessibilityState.disabled === true`, which does correctly cover the behavior. Suggest either dropping the redundant press + comment, or exercising the guard directly (e.g. call the `onPress` prop) if you want real coverage of `if (disabled) return;`. Same pattern repeats in the `sharingLink` and drop-all tests.

---

lilseyi (commented):
**Automated review — verdict: approved (no blocking findings).**

Reviewed the diff against the approved spec across four lenses (correctness, security, spec-fidelity/UX, tests), adversarially verifying each finding. Nothing blocking survived.

- **Correctness** — clean. The empty-`planIds`→"all upcoming" backend fallback is unreachable from the UI (`selectedIds` is derived from the live `events` prop, so a drained selection disables Share). Seed-once `selected` set is safe because the picker only mounts after the non-empty check. No stale-closure/dep-array issues.
- **Security** — none. Frontend-only change; the backend re-validates every `planId` against the group and requires scheduler role, so client-chosen ids can't leak another group's events.
- **Spec-fidelity** — every acceptance criterion met. Sourcing upcoming events from `listEvents` (rather than the grid's `rosterMatrix`-derived `events`) is a justified improvement — `rosterMatrix` caps/leads with past dates and would silently drop upcoming events. One non-blocking note left inline about the client-side 12-event cap.
- **Tests** — `AvailabilityPicker.test.tsx` runs 7/7 green and asserts the actual `planIds` (order-sensitive), including the live-events reactivity cases. Two non-blocking notes left inline (a tautological disabled-press assertion; the `handleCollectAvailability` empty-state branch is untested since it lives in the un-tested `RosterGridScreen`).

All findings are non-blocking polish. Good, focused, spec-matching change.

---

PR #624, opened 2026-07-15
Make post-update notice a silent "Updating…/Updated" pill

lilseyi:
## What & why

Seyi asked that the post-OTA-update notice feel like a **silent** update rather than an alarm. Today, the first launch on a new bundle slides down a full-width dark banner reading _"App just updated — if anything feels stuck, force-close and reopen."_ That's intrusive, and the "force-close/refresh" instruction is redundant: by the time this component runs, the new OTA bundle is **already live** (updates apply only on a cold start), so there's nothing to refresh.

This replaces the banner with a small, centered **toast pill** near the top of the screen:

- Shows a small **spinner + "Updating…"** for a brief beat (~1s),
- swaps to a **green check + "Updated"**,
- then **fades away on its own** after a short hold (~3s total on screen).

No dismiss control, no "force-close/reopen" copy anywhere in the component.

## Before / after

Rendered mock built from the component's real styles — **not** a device capture (this flow is `__DEV__`-gated and only fires on a real OTA on staging/prod, so CI can't exercise it):

[before/after](https://images.togather.nyc/dev-assistant/aff05ab9-f906-4000-af99-42f8dd22ef43-post-update-pill-before-after.png)

## What stayed the same

Only presentation and copy changed — the **trigger logic is untouched**:

- One-shot per bundle: compares `Updates.updateId` against `@togather/last_seen_update_id` in AsyncStorage.
- Bails on `isEmbeddedLaunch` (fresh install / App Store update / embedded fallback).
- `__DEV__`-gated (never shows in development).
- Top-safe-area aware (floats over the notch, does **not** push content down).
- Fails open (no pill) if storage throws.

Additional robustness: the pill is non-interactive (`pointerEvents="none"`) so it never captures taps on the screen beneath it, and unmount runs on its own timer rather than the fade's completion callback, so it always tears down with no user action.

## Tests

`PostUpdateRecoveryBanner.test.tsx` updated for the new copy/markup — the old assertions (banner text, ✕ dismiss, safe-area padding) tracked to the new pill:

- detection cases unchanged (dev-gate, embedded launches, first-launch-with-this-code, matching id, differing id, fail-open),
- **"Updating…" → "Updated"** phase transition,
- **auto-dismiss** over the full timeline with no user action,
- **non-interactivity** (`pointerEvents="none"`),
- top-inset offset.

All 12 tests pass locally.

## Verification note

Still needs a **staging OTA smoke test** on a device to confirm the pill appears, animates, and self-dismisses — this path is device-only and `__DEV__`-gated, so it can't be exercised in CI.

---

Dashboard item: bug `x97671grvhvcstrqk60g8dbg9d8ajr6b` · GitHub issue #622

---
_Generated by [Claude Code](https://claude.ai/code/session_01XEZLPJb878zNCBTJd3ZjXU)_

### DIFF
--- a/apps/mobile/components/ui/PostUpdateRecoveryBanner.tsx
+++ b/apps/mobile/components/ui/PostUpdateRecoveryBanner.tsx
@@ -1,42 +1,50 @@
 /**
  * PostUpdateRecoveryBanner
  *
- * Renders a one-shot informational banner the first time the app boots
- * onto a new OTA bundle. Backstop for the iOS + Fabric touch-reattach
- * race that can wedge the UI after a foreground reloadAsync: the JS
- * thread keeps running so this banner still appears, telling the user
- * how to recover (force-close + reopen) even when taps are dead.
+ * Renders a small, unobtrusive "toast" pill the first time the app boots
+ * onto a new OTA bundle. By the time this component runs the new bundle is
+ * already live (OTA updates apply only on a cold start / refresh), so the
+ * pill simply reassures the user that the app updated: it shows a brief
+ * spinner + "Updating…", settles on a green check + "Updated," then fades
+ * away on its own. It never blocks interaction and carries no "force-close"
+ * or "refresh" copy — the update has already been installed.
  *
- * Detection: bail entirely on embedded launches (Updates.isEmbeddedLaunch
- * = true — fresh installs, App Store updates, emergency fallback to the
- * embedded bundle); none of those carry a reloadAsync race. On OTA
- * bundles, compare Updates.updateId against the value persisted to
- * AsyncStorage from the previous OTA launch and show the banner if it's
- * absent (first OTA launch with the banner code on this user's device,
- * or first OTA after a native install) or different (a real OTA
- * transition). The current id is then persisted so the banner is
- * one-shot per bundle.
+ * Detection (unchanged one-shot-per-bundle logic): bail entirely on embedded
+ * launches (Updates.isEmbeddedLaunch = true — fresh installs, App Store
+ * updates, emergency fallback to the embedded bundle). On OTA bundles,
+ * compare Updates.updateId against the value persisted to AsyncStorage from
+ * the previous OTA launch and show the pill if it's absent (first OTA launch
+ * with this code on the device, or first OTA after a native install) or
+ * different (a real OTA transition). The current id is then persisted so the
+ * pill is one-shot per bundle. Fails open (no pill) if storage throws.
  *
- * Auto-dismisses after AUTO_DISMISS_MS (wedged users can't tap, but JS
- * setTimeout still fires). Tap-X works for non-wedged users.
+ * Presentation timeline (all cosmetic): fade in → "Updating…" for
+ * UPDATING_PHASE_MS → swap to "Updated" → hold UPDATED_HOLD_MS → fade out.
+ * Auto-dismisses with no user action (the whole point — silent); there is no
+ * manual dismiss control.
  */
 import React, { useState, useEffect, useRef } from 'react';
-import { View, Text, StyleSheet, Pressable, Platform } from 'react-native';
+import { View, Text, StyleSheet, Animated, ActivityIndicator, Platform } from 'react-native';
 import { Ionicons } from '@expo/vector-icons';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import * as Updates from 'expo-updates';
 import AsyncStorage from '@react-native-async-storage/async-storage';
 
 export const STORAGE_KEY = '@togather/last_seen_update_id';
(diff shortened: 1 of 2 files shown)


apps/mobile/components/ui/__tests__/PostUpdateRecoveryBanner.test.tsx
```diff
@@ -158,79 +162,98 @@ describe('PostUpdateRecoveryBanner', () => {
...
         });
 
+        expect(screen.getByText(UPDATED_TEXT)).toBeTruthy();
+        expect(screen.queryByText(UPDATING_TEXT)).toBeNull();
+      } finally {
+        jest.useRealTimers();
+      }
+    });
+
+    it('auto-dismisses (fades out) after the full timeline with no user action', async () => {
```
lilseyi, line 207: Non-blocking (test coverage): the two timeline tests only bound the *dismiss* as an upper bound — this one advances the full `UPDATING_PHASE_MS + UPDATED_HOLD_MS + FADE_MS` (3300ms) and the phase test pins the swap at exactly `UPDATING_PHASE_MS`. Nothing asserts the pill is still on screen *during* the "Updated" hold, so a regression that dismissed the pill early (e.g. at ~1500ms, after the phase swap but before the intended fade) would pass both tests. Consider an intermediate assertion — pill present at `UPDATING_PHASE_MS + UPDATED_HOLD_MS - 1`, gone shortly after — to pin the ~2s hold. Not a blocker.

---

apps/mobile/components/ui/PostUpdateRecoveryBanner.tsx
```diff
@@ -1,42 +1,50 @@
...
- * race that can wedge the UI after a foreground reloadAsync: the JS
- * thread keeps running so this banner still appears, telling the user
- * how to recover (force-close + reopen) even when taps are dead.
+ * Renders a small, unobtrusive "toast" pill the first time the app boots
+ * onto a new OTA bundle. By the time this component runs the new bundle is
+ * already live (OTA updates apply only on a cold start / refresh), so the
+ * pill simply reassures the user that the app updated: it shows a brief
+ * spinner + "Updating…", settles on a green check + "Updated," then fades
+ * away on its own. It never blocks interaction and carries no "force-close"
+ * or "refresh" copy — the update has already been installed.
```
lilseyi, line 10: Non-blocking (nit): the spec's "no 'refresh'/'force-close'/'reopen' wording remains anywhere in this component" is satisfied for all *rendered* copy — only "Updating…"/"Updated" reach the screen. The words survive here only in the JSDoc, and as meta-documentation stating the copy was *removed*. Fine in spirit; flagging so a future literal grep doesn't get a false hit. Optional tidy-up, not a blocker.

---

lilseyi (commented):
**Review — approved (no blocking findings).** riskLevel: low.

Reviewed the diff against the approved spec across four lenses (correctness, security, spec-fidelity/UX, tests), adversarially verifying each candidate finding.

**Spec fidelity ✓** — every "Done when" item is met: the full-width `#1F2937` banner (icon + two-line copy + ✕) is fully replaced by a small centered dark pill that shows `ActivityIndicator` + "Updating…" → green `checkmark-circle` + "Updated", then fades out on its own via `setTimeout`s. No "refresh/force-close/reopen" copy reaches the screen. The pill floats over the top safe-area (`top: 8 + insets.top`) without pushing content down and is non-interactive (`pointerEvents="none"`).

**Trigger logic unchanged ✓** — verified line-by-line against `origin/main`: `__DEV__` gate, `isEmbeddedLaunch` bail, `updateId` vs AsyncStorage `@togather/last_seen_update_id` compare, one-shot `setItem`, and fail-open on storage throw are behaviorally identical. Only presentation/copy changed — no scope creep.

**Correctness ✓** — the presentation effect (deps `[isVisible, opacity]`) can't get stuck or re-run mid-timeline; unmount runs on its own timer rather than the animation callback, so teardown is guaranteed even if the native anim callback never fires. `phase` non-reset after unmount is harmless (one-shot per bundle).

**Security ✓** — purely presentational; the persisted `updateId` is a non-sensitive build id, never logged/rendered/sent, and rendered text is a fixed literal.

**Tests ✓** — updated suite covers detection (dev-gate, embedded-launch incl. the App-Store-with-prior-id case, first-launch-no-stored-id, matching/differing id, fail-open) plus the new phase transition, full-timeline auto-dismiss, non-interactivity, and the safe-area offset math (`top: 55` = 8 + 47).

Two **non-blocking** notes left inline (a test-coverage gap on pinning the "Updated" hold, and a JSDoc-only mention of the removed copy). Neither gates merge.

Reminder from the spec: the `__DEV__`-gated, device-only OTA path can't be exercised in CI — a staging OTA smoke test on a real device is still needed to confirm the pill appears, animates, and self-dismisses.

---

PR #630, opened 2026-07-19
Link previews for every page: registry-driven OG metadata + unified preview resolver

lilseyi:
## Summary

Overhauls the link-preview system so **every page ships a real share preview by construction**, and adding future pages requires exactly one step.

### Static pages (`apps/web`)
- New site-wide route registry `src/routes.tsx` — `main.tsx` now generates the router from it, so a page cannot exist without title/description metadata. Guide entries derive their copy from the existing guides registry (no duplication).
- New post-build step (`scripts/generate-static-pages.tsx`, wired into `pnpm build`): writes `dist/<path>/index.html` per route with page-specific `<title>`/description/OG/Twitter tags, and renders a branded 1200×630 card (satori + resvg, Plus Jakarta Sans, `#D4A574`) at `/og/<slug>.png` for every page without a bespoke image. Cloudflare Pages serves these static files before the SPA fallback — bots get correct meta with zero runtime infrastructure.
- Previously all landing/guide pages shared the homepage's generic OG block; now all 19 pages have their own preview (18 generated cards + the homepage's bespoke image).

### Dynamic routes (`apps/convex` + `apps/link-preview`)
- One typed resolver (`functions/linkPreviewMeta.ts`) behind `GET /link-preview/meta?url=…` now assembles all preview content: title formats, rich descriptions, timezone-aware date formatting, and image fallback chains (event cover → group image → community logo). The five per-type `/link-preview/*` endpoints are removed (`/api/link-preview` chat unfurling is untouched).
- The Cloudflare worker shrinks 1,573 → 742 lines: a `PREVIEW_ROUTES` table plus one shared OG template and one shared error template. Its routing, `.well-known`, and redirect jobs are unchanged. Adding a dynamic preview = one resolver case + one table row, no worker redeploy for content changes.

### Docs
- ADR-009 rewritten for the new architecture (amendment dated 2026-07-19); CLAUDE.md documents the one-step conventions for future pages.

## Test plan
- [x] Worker: `node --test` — 34/34 passing
- [x] Convex: 45 new resolver tests (dispatch, date formatting, fallback chains, slug canonicalization); full suite green (2,361 tests / 145 files); `convex typecheck` clean
- [x] Web: type-check, lint, and full build clean; verified per-route HTML head tags and generated PNGs (1200×630, ~69–93 KB)
- [ ] After staging deploy: paste a guide link + an event link into Slack/iMessage to confirm cards render

## Deploy note
The old per-type endpoints are removed in the same change (remove-don't-deprecate). If Convex deploys before the worker, bots briefly get the generic error-fallback preview on dynamic links — seconds on staging where both deploy on merge.

https://claude.ai/code/session_013VHVFzj5sj8uq91cP7389D

---
_Generated by [Claude Code](https://claude.ai/code/session_013VHVFzj5sj8uq91cP7389D)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -313,6 +313,14 @@ What to update when a guide is affected:
 If you're unsure whether a change is "user-facing enough" to need a guide
 update, it probably is — err on the side of updating.
 
+### Link Previews / OG Metadata
+
+**Adding a static marketing page:** Add an entry to `apps/web/src/routes.tsx` with `path`, `component`, `title`, `description`, and optional `image` and `emoji`. The build script automatically generates OG metadata and images at compile time (satori + resvg for branded cards). Routes are **the only way to add a page** — the router is generated from the registry.
+
+**Adding a dynamic shareable app route:** Implement a resolver case in `apps/convex/functions/linkPreviewMeta.ts` (typed, unit-tested) to assemble preview metadata (title, description, image fallback chains, timezone formatting, etc.), then add a row to `PREVIEW_ROUTES` in `apps/link-preview/cloudflare-worker.js` to route the pattern. The worker fetches the metadata endpoint and renders the shared HTML template — **no per-type logic in the worker itself.**
+
+See ADR-009 for full architecture.
+
 ## File and Project Hygiene
 
 ### Keep the Codebase Clean
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -315,7 +315,7 @@ update, it probably is — err on the side of updating.
 
 ### Link Previews / OG Metadata
 
-**Adding a static marketing page:** Add an entry to `apps/web/src/routes.tsx` with `path`, `component`, `title`, `description`, and optional `image` and `emoji`. The build script automatically generates OG metadata and images at compile time (satori + resvg for branded cards). Routes are **the only way to add a page** — the router is generated from the registry.
+**Adding a static marketing page:** Add an entry to `apps/web/src/routes.tsx` with `path`, `element` (JSX component like `<EventsGuide />`), `title`, `description`, and optional `image` and `emoji`. The build script automatically generates OG metadata and images at compile time (satori + resvg for branded cards). Routes are **the only way to add a page** — the router is generated from the registry. For a **new top-level path** (not nested under an existing prefix like `/guides/`), you must also add it to `LANDING_PAGE_PATHS` or `LANDING_PAGE_PREFIXES` in `apps/link-preview/cloudflare-worker.js`, or the Cloudflare worker will misroute it; paths under already-listed prefixes (like `/guides/`) need no worker change.
 
 **Adding a dynamic shareable app route:** Implement a resolver case in `apps/convex/functions/linkPreviewMeta.ts` (typed, unit-tested) to assemble preview metadata (title, description, image fallback chains, timezone formatting, etc.), then add a row to `PREVIEW_ROUTES` in `apps/link-preview/cloudflare-worker.js` to route the pattern. The worker fetches the metadata endpoint and renders the shared HTML template — **no per-type logic in the worker itself.**
 
--- a/apps/convex/http.ts
+++ b/apps/convex/http.ts
@@ -165,21 +165,24 @@ http.route({
   method: "GET",
   handler: httpAction(async (ctx, request) => {
     const url = new URL(request.url);
+    // `searchParams.get` already URL-decodes once; the worker single-encodes
+    // the target URL, so decoding again here would mangle reserved chars
+    // (e.g. `%26` -> `&`) that are legitimately part of the target's query string.
     const targetUrl = url.searchParams.get("url");
 
     if (!targetUrl) {
       return jsonResponse({ error: "Missing url parameter" }, 400);
     }
 
-    let decodedUrl: string;
+    // Validate it's a well-formed URL before handing it to the resolver.
     try {
-      decodedUrl = decodeURIComponent(targetUrl);
+      new URL(targetUrl);
     } catch {
       return jsonResponse({ error: "Invalid url encoding" }, 400);
     }
 
     try {
-      const result = await resolveLinkPreviewMeta(ctx, decodedUrl);
+      const result = await resolveLinkPreviewMeta(ctx, targetUrl);
       return jsonResponse(result.body, result.status);
     } catch (error) {
       console.error("Error resolving link preview meta:", error);
--- a/apps/convex/http.ts
+++ b/apps/convex/http.ts
@@ -12,10 +12,11 @@
 
 import { httpRouter } from "convex/server";
 import { httpAction } from "./_generated/server";
-import { api, internal } from "./_generated/api";
+import { internal } from "./_generated/api";
 import { Id } from "./_generated/dataModel";
 import { verifySlackSignature } from "./functions/slackServiceBot/slack";
 import { hashApiKey } from "./lib/apiKeys";
+import { resolveLinkPreviewMeta } from "./functions/linkPreviewMeta";
 import { registerRoutes } from "@supa-media/dev-assistant";
 import "./functions/devAssistant/config"; // side-effect: sets config first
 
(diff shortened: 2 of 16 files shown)


apps/convex/http.ts
```diff
@@ -141,303 +142,55 @@ async function verifyStripeSignature(
...
-http.route({
-  path: "/link-preview/channel",
-  method: "GET",
-  handler: httpAction(async (ctx, request) => {
-    const url = new URL(request.url);
-    const shortId = url.searchParams.get("shortId");
-
-    if (!shortId) {
-      return jsonResponse({ error: "Missing shortId parameter" }, 400);
+      decodedUrl = decodeURIComponent(targetUrl);
```
lilseyi: **[CONFIRMED — correctness] Double-decode corrupts URLs with encoded reserved characters.**

`url.searchParams.get("url")` already percent-decodes once (WHATWG URL spec), and the worker only encodes once (`encodeURIComponent(requestUrl)` in `fetchPreviewMeta`). This extra `decodeURIComponent` is a second decode.

Verified repro: original URL `/nearme?type=r%26b` → after this double decode becomes `/nearme?type=r&b`, which `new URL()` splits into `type=r` + bogus `b=` — the group-type lookup silently gets the wrong slug. Any `%`-encoded reserved char (`&`, `=`, `#`, `+`, `%`) in the original query corrupts; a stray `%` makes this line throw `URIError` (caught → 400, so fails closed, but on valid URLs).

Fix: drop the `decodeURIComponent` and use the value from `searchParams.get("url")` directly.

---

apps/link-preview/cloudflare-worker.js
```diff
@@ -578,461 +427,26 @@ function generateHomepageOgHtml(config) {
...
+ * @param {string|null} meta.image
+ * @param {string} meta.url - Absolute canonical destination for the meta-refresh redirect
+ * @param {string} meta.siteName
+ * @param {string} [meta.imageAlt]
+ */
+function renderOgHtml(meta) {
+  const title = escapeHtml(meta.title || "");
+  const description = escapeHtml(meta.description || "");
+  const siteName = escapeHtml(meta.siteName || BRAND_NAME);
+  const imageAlt = escapeHtml(meta.imageAlt || title);
```
lilseyi: **[CONFIRMED — correctness] Double-escaped `twitter:image:alt` for nearme/community previews.**

`title` on the line above is already `escapeHtml(meta.title)`; using it as the fallback input to a second `escapeHtml()` double-escapes entities. `buildNearmeMeta` and `buildCommunityMeta` never set `imageAlt`, so every /nearme and /:slug preview with an `&`, `<`, `>`, or quote in the community name renders mangled alt text (`Arts &amp; Crafts` → `Arts &amp;amp; Crafts`).

```suggestion
  const imageAlt = escapeHtml(meta.imageAlt || meta.title || "");
```

---

apps/link-preview/cloudflare-worker.js
```diff
@@ -1083,153 +498,82 @@ function generateCommunityOgHtml(data, slug, config) {
...
 /**
- * Generate community error HTML (fallback when API fails)
+ * Render the shared fallback HTML for bot-preview routes when the Convex
+ * meta endpoint returns a 404 (unknown entity) or the fetch itself fails.
+ * @param {string} pathname - Original request path (e.g. "/e/abc123")
+ * @param {Object} config - Environment-specific configuration
  */
-function generateCommunityErrorHtml(slug, config) {
-  const pageUrl = `${config.baseUrl}/${slug}`;
+function renderPreviewErrorHtml(pathname, config) {
```
lilseyi: **[CONFIRMED — behavior regression] Error page drops the query string and the `/c/` canonicalization.**

Two regressions vs the old per-type error pages:
1. Callers pass only `url.pathname`, so `/nearme?type=book-club` error-falls-back to a redirect targeting bare `/nearme` — the old `generateNearmeErrorHtml(url.href)` preserved the full URL including query.
2. For the community-slug route, the old `generateCommunityErrorHtml` redirected to `${config.baseUrl}/c/${slug}`; this now targets the bare `/${slug}`, forcing an extra client-side hop through CommunityRedirect.

Fix: accept the full original URL (pathname + search), and for the community route pass the canonical `/c/<slug>` path. Worth a test for the slug-route error path — nothing covers it today.

---

apps/link-preview/cloudflare-worker.js
```diff
@@ -1041,31 +455,32 @@ function generateCommunityOgHtml(data, slug, config) {
...
   <meta property="og:title" content="${title}">
   <meta property="og:description" content="${description}">
-  <meta property="og:site_name" content="${BRAND_NAME}">
+  <meta property="og:site_name" content="${siteName}">
   ${imageUrl ? `<meta property="og:image" content="${imageUrl}">` : ""}
   ${imageUrl ? `<meta property="og:image:secure_url" content="${imageUrl}">` : ""}
   ${imageUrl ? `<meta property="og:image:type" content="image/jpeg">` : ""}
-  ${imageUrl ? `<meta property="og:image:width" content="400">` : ""}
-  ${imageUrl ? `<meta property="og:image:height" content="400">` : ""}
+  ${imageUrl ? `<meta property="og:image:width" content="1200">` : ""}
```
lilseyi: **[CONFIRMED — behavior regression] Square community logos now advertised as 1200×630.**

The old `generateNearmeOgHtml`/`generateCommunityOgHtml` emitted `og:image:width/height` of **400×400** for square community-logo images; the shared template hardcodes 1200×630 for every route. Crawlers (Facebook/Twitter/Slack) pre-allocate the card aspect ratio from these declared dimensions, so square logos on /nearme and /:slug previews will render stretched.

Fix: add optional `imageWidth`/`imageHeight` to the meta contract (set 400×400 in `buildNearmeMeta`/`buildCommunityMeta`) and use them here with 1200×630 as the default.

---

apps/web/src/routes.tsx
```diff
@@ -0,0 +1,150 @@
...
+  "check-in": <CheckIn />,
+  prayer: <Prayer />,
+};
+
+const guideRoutes: RouteEntry[] = guides.map((guide) => ({
+  path: `/guides/${guide.slug}`,
+  title: `${guide.title} | Togather Guides`,
+  description: guide.summary,
+  emoji: guide.emoji,
+  element: guideComponents[guide.slug],
```
lilseyi: **[CONFIRMED — latent footgun] Unchecked `guideComponents[guide.slug]` lookup ships a silent blank page.**

Without `noUncheckedIndexedAccess` this types as `ReactElement` (not `| undefined`), so a future guide added to `guides/registry.ts` without a matching entry here registers a route that renders nothing — correct meta tags, blank body, no type error, no test failure.

Fix: throw on a missing key in the map (module-scope throw means the OG build script — which imports this file — fails the build in CI, converting the silent blank page into a loud failure):

```ts
const element = guideComponents[guide.slug];
if (!element) throw new Error(`No guide component registered for slug "${guide.slug}"`);
```

---

docs/architecture/ADR-009-link-preview-system.md
```diff
@@ -39,59 +48,113 @@ Implement a link preview system using a single **Cloudflare Worker** that:
...
 | Component | Location | Purpose |
 |-----------|----------|---------|
-| Cloudflare Worker | `apps/link-preview/cloudflare-worker.js` | Bot detection, routing, OG HTML generation |
-| Wrangler Config | `apps/link-preview/wrangler.toml` | Cloudflare Worker routes |
+| Route Registry | `apps/web/src/routes.tsx` | Site-wide routes with preview metadata (title, description, image, emoji) |
+| Build Script | `apps/web/scripts/generate-static-pages.tsx` | Generates prebuilt HTML + OG image cards at build time (satori + resvg) |
+| Web App | `apps/web/` | Vite SPA deployed to Cloudflare Pages; routes come from registry |
+| Link Preview Resolver | `apps/convex/functions/linkPreviewMeta.ts` | Typed resolver: all preview assembly logic (title formats, descriptions, fallback chains) |
+| Cloudflare Worker | `apps/link-preview/cloudflare-worker.js` | Thin layer: bot detection, routing, PREVIEW_ROUTES table, fetches meta endpoint, renders shared template |
+| HTML Template | `apps/link-preview/templates/preview.html` | Shared template for dynamic preview OG HTML |
```
lilseyi: **[CONFIRMED — docs accuracy] ADR references components that don't exist in the implementation.**

Three phantom references introduced in this rewrite:
- Line 90: `apps/link-preview/templates/preview.html` — no such file; the template is the inline `renderOgHtml()` in `cloudflare-worker.js`.
- Line 135: example `RouteEntry` uses `component: EventsGuide` — the real field is `element` (a `ReactElement`).
- Lines 384/386: troubleshooting references `isStaticPath()` — the actual function is `isLandingPagePath()`.

Since this ADR is the designated source of truth for the system, these should be corrected before merge.

---

CLAUDE.md
```diff
@@ -313,6 +313,14 @@ What to update when a guide is affected:
 If you're unsure whether a change is "user-facing enough" to need a guide
 update, it probably is — err on the side of updating.
 
+### Link Previews / OG Metadata
+
+**Adding a static marketing page:** Add an entry to `apps/web/src/routes.tsx` with `path`, `component`, `title`, `description`, and optional `image` and `emoji`. The build script automatically generates OG metadata and images at compile time (satori + resvg for branded cards). Routes are **the only way to add a page** — the router is generated from the registry.
```
lilseyi: **[CONFIRMED — altitude/docs] "Only way to add a page" overstates — the worker's path lists are a required second step, and the field name is wrong.**

Two issues:
1. The registry is *not* sufficient for a new **top-level** path: the worker's hand-maintained `LANDING_PAGE_PATHS`/`LANDING_PAGE_PREFIXES` decide Pages-vs-EAS routing. A new single-segment page (e.g. `/pricing`) added only to `routes.tsx` would match the worker's community-slug rule and 302-redirect to `/c/pricing`. (All current registry paths are covered — verified — but the drift risk is structural.) Either derive the worker lists from the registry at build time, or document the second step here and in `routes.tsx`'s doc comment.
2. This section says entries take `component` — the actual `RouteEntry` field is `element`.

---

lilseyi (commented):
## Claude review — 8 finder angles × adversarial verification

**9 confirmed findings.** 7 are inline comments on the diff; these 3 are in files *outside* the diff, which inline comments can't attach to:

### 1. [CONFIRMED — breaks production previews] `apps/mobile/app/(landing)/nearme/index+api.ts:223`
Still fetches the deleted `GET /link-preview/community?communitySubdomain=...`. Every bot hitting the mobile web build's `/nearme` route now gets `response.ok === false` and silently falls back to the generic error OG card. This and #2 are the only two remaining call sites of the five deleted endpoints (repo-wide grep). Fix: call `GET /link-preview/meta?url=<encoded original URL>` and consume the flat `{title, description, image, url, siteName}` shape.

### 2. [CONFIRMED — breaks dev tooling] `apps/mobile/scripts/serve-with-og.js:156`
The local OG-preview server builds `/link-preview/{event,group}` URLs — both deleted. Every simulated bot request now serves error HTML, masking real regressions during manual QA. Same fix as #1.

### 3. [CONFIRMED — deploy gap] `.github/workflows/deploy-landing.yml`
The build step never sets `SITE_ORIGIN`, so staging deploys bake `https://togather.nyc` into every page's `og:url`/`og:image`. Verified this is an *inherited* limitation (the old single `index.html` also hardcoded prod URLs) — but the new per-route canonical URLs make it far more consequential. Fix: set `SITE_ORIGIN` per environment in the workflow.

### Verified-but-not-blocking (noted for the record)
- Origin echo: `/link-preview/meta` derives the canonical `url` from the caller-supplied origin with no allowlist. The only production caller is the worker passing its own `request.url`, so exploitation is theoretical today; a hostname allowlist is cheap hardening if a second consumer ever appears.
- The escaped `og:description` can now exceed the intended length cap (truncation moved before escaping) — unobservable in practice since crawlers self-truncate well below it.
- Cleanup opportunities (non-blocking): `escapeHtml` now exists in 3 places; the OG tag block is hand-built in both the web build script and the worker; the shortId route regexes are duplicated worker↔resolver; `PageMeta` is redefined in the build script instead of type-imported; `emoji` is declared/populated but never rendered on cards.

All 9 confirmed findings are being fixed on this branch; threads will be resolved as each lands.

---

apps/web/scripts/generate-static-pages.tsx
```diff
@@ -0,0 +1,349 @@
...
+}
+
+async function loadFonts() {
+  const fontDir = path.join(
+    webRoot,
+    "node_modules/@fontsource/plus-jakarta-sans/files",
+  );
+  // satori needs raw ttf/otf/woff font data — NOT woff2 — so we read the
+  // .woff files @fontsource ships alongside its .woff2 ones.
+  const read = (weight: number) => fs.readFileSync(path.join(fontDir, `plus-jakarta-sans-latin-${weight}-normal.woff`));
```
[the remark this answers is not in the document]
lilseyi: Checked against the installed package: `@fontsource/plus-jakarta-sans@5.2.8` ships `.woff` alongside `.woff2` (56 `.woff` files in `files/`, including `plus-jakarta-sans-latin-{400,600,700}-normal.woff` — exactly what `loadFonts()` reads). The build runs green end-to-end and renders valid 1200×630 PNGs (`generate-static-pages: wrote 19 route HTML file(s) and 18 OG card PNG(s)`), which would be impossible if `loadFonts()` threw ENOENT. Resolving as a false positive.

---

apps/web/scripts/generate-static-pages.tsx
```diff
@@ -0,0 +1,349 @@
...
+import path from "node:path";
+import fs from "node:fs";
+import esbuild from "esbuild";
+import satori from "satori";
+import { Resvg } from "@resvg/resvg-js";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const webRoot = path.resolve(__dirname, "..");
+const distDir = path.join(webRoot, "dist");
+const SITE_ORIGIN = (process.env.SITE_ORIGIN || "https://togather.nyc").replace(/\/$/, "");
```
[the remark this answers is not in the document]
lilseyi: Fixed on this branch — `deploy-landing.yml`'s build step now sets `SITE_ORIGIN` from the workflow's existing environment target (`production` → `https://togather.nyc`, otherwise `https://staging.togather.nyc`), so staging pages bake staging canonical/OG URLs. (This matched finding #3 in the Claude review body as well.)

---

PR #632, opened 2026-07-19
Include unconfirmed volunteers in serving mode

lilseyi:
## What & why

When a leader assigns someone to serve, that person starts **unconfirmed** until they tap "accept". Today the volunteer-facing **serving-mode** surfaces only showed and admitted *confirmed* people — so an assigned-but-unconfirmed volunteer was invisible on the Team roster and **couldn't even enter Serving Mode**, even though (as the reporter put it) they're "more than likely to still come, they just haven't accepted."

This brings the serving surfaces onto the same **`status !== "declined"`** predicate the rest of the app already uses (team-chat membership, leader fill counts): include everyone who hasn't declined, and clearly mark the unconfirmed ones. **Declined stays hidden. Leader-facing counts are untouched.**

Dashboard item / issue: #631

## Before / after (Team roster)

> ⚠️ **Rendered mock, not a device capture** — built from the real `ServingTeamScreen` styles and the light-theme color tokens (`colors.warning` `#FF9500`, surface/border/text). This environment is a headless Linux runner with no iOS simulator.

[Serving Team roster before/after](https://images.togather.nyc/dev-assistant/2512c0c7-5021-4ddb-8ef8-93d7173a24d4-serving-unconfirmed-before-after.png)

Before, the unconfirmed teammate is absent. After, they appear with an amber **"Unconfirmed"** pill (clock icon), reusing the leader roster's "awaiting" vocabulary.

## Changes

**Backend — `apps/convex/functions/scheduling/`**
- `serving.ts`
  - `getServingEligibility`: an assigned-but-**unconfirmed** volunteer is now *eligible* to enter, but **`autoEnter` is still computed from confirmed assignments only** — eligible-but-not-auto-entered, so the nudge to accept stays meaningful without locking them out.
  - `getServingTeamRoster`: includes non-declined people and returns each person's accept `status` so the client can badge unconfirmed ones (one card per (user, role), so a person confirmed on one role and unconfirmed on another reads correctly).
  - `getServingUpcomingChannels`: resolves "coming soon" channels for non-declined assignments too (team chat already mirrors unconfirmed members into channels).
- `eventTasks.ts`
  - `getMyServingTasks`: an unconfirmed volunteer sees the tasks for the role they're assigned to (they can now enter and serve).
  - `getCrewTasks`: includes non-declined teammates and carries their `status`.
  - `myConfirmedAssignments` → renamed/widened to **`myServingAssignments`** (non-declined), so Shared/Crew reads and the shared-task-toggle gate admit unconfirmed teammates.
  - **Unchanged:** leader-facing readiness/fill math (`assigneesForTask`, `roster.ts`, `events.ts`) stays confirmed-only.

**Frontend — `apps/mobile/features/serving/`**
- `ServingTeamScreen.tsx`: a small **"Unconfirmed"** pill on unconfirmed person cards; empty-state copy no longer says "confirmed" ("No one else is on the team yet.").
- `ServingTasksScreen.tsx`: an **"Unconfirmed"** chip on unconfirmed crew rows.
- `ChatInboxScreen.tsx`: **no change needed** — the serving chip already keys off `eligible` and the auto-enter effect off `autoEnter`, so newly-eligible unconfirmed users get the manual chip but are never auto-entered.

> Note: these serving screens have no separate `.web.tsx` — they're single cross-platform files (rendered on web via react-native-web), so "web parity" is automatic. `ServingRunsheetScreen` is a launcher into the Rostering run sheet and renders no teammate list of its own, so it needed no per-row change; unconfirmed users reach it via the widened eligibility.

## Tests
- `serving.test.ts`: unconfirmed people appear on the roster (badged), declined stay hidden, unconfirmed callers reach their own plan, and `getServingEligibility` is `eligible: true / autoEnter: false` for unconfirmed, `true/true` for confirmed, `false/false` for declined.
- `eventTasks.test.ts`: an unconfirmed volunteer sees their role's tasks in "Mine"; `getCrewTasks` includes an unconfirmed teammate tagged `unconfirmed` and excludes a declined one.
- Full `apps/convex/__tests__/scheduling` suite green (358 passing); convex typecheck clean; mobile eslint/typecheck clean on touched files.

## Review call-outs
- Phone-number exposure in the Team roster already applies to teammates on the same event; this extends it to **unconfirmed** teammates on the same event — consistent with the existing behavior, but worth a look.
- The shared-task-toggle gate now admits unconfirmed teammates (they can flip a team-wide shared task). Intentional per the "reach the same resources" goal.

https://claude.ai/code/session_0156x8CChUz3vD1mq2tYwima

---
_Generated by [Claude Code](https://claude.ai/code/session_0156x8CChUz3vD1mq2tYwima)_

### DIFF
--- a/apps/convex/functions/scheduling/eventTasks.ts
+++ b/apps/convex/functions/scheduling/eventTasks.ts
@@ -601,8 +601,8 @@ export const reorderTasks = mutation({
  * only for "during" tasks (one completion per service time).
  *
  * Team-level tasks (roleId == null) are REJECTED here — they are team-wide and
- * must go through `toggleSharedTeamTask` (which gates on confirmed team
- * membership); this per-user path only checks group membership.
+ * must go through `toggleSharedTeamTask` (which gates on serving/non-declined
+ * team membership); this per-user path only checks group membership.
  *
  * Auth: any authenticated user (completion is personal). We still verify the
  * caller can see the plan.
@@ -621,8 +621,8 @@ export const toggleTaskCompletion = mutation({
       throw new ConvexError("Task not found");
     }
     // Team-level tasks (no roles) are TEAM-WIDE and must be completed through
-    // `toggleSharedTeamTask`, which gates on confirmed TEAM membership. This
-    // per-user path only checks GROUP membership, so accepting a team-level
+    // `toggleSharedTeamTask`, which gates on serving (non-declined) TEAM
+    // membership. This per-user path only checks GROUP membership, so accepting a team-level
     // completion here would let a non-teammate (or a stale/direct caller) mark it
     // done and bypass the team gate. The mobile client already routes team-level
     // tasks to the Shared tab and excludes them from "Mine".
@@ -883,14 +883,17 @@ export const getMyServingTasks = query({
       ServingTaskItem[]
     > = { before: [], during: [], after: [] };
 
-    // The roles / teams the user is confirmed for on this plan.
+    // The roles / teams the user is on (non-declined) for this plan. An
+    // unconfirmed volunteer can now enter serving mode, so they see the tasks
+    // for the role they're assigned to regardless of their own accept state —
+    // only a *declined* assignment drops its role's tasks.
     const myAssignments = await ctx.db
       .query("roleAssignments")
       .withIndex("by_plan", (q) => q.eq("planId", args.planId))
       .filter((q) => q.eq(q.field("userId"), userId))
       .collect();
-    const myConfirmed = myAssignments.filter((a) => a.status === "confirmed");
-    const confirmedRoleIds = new Set(myConfirmed.map((a) => a.roleId as string));
+    const myServing = myAssignments.filter((a) => a.status !== "declined");
+    const myRoleIds = new Set(myServing.map((a) => a.roleId as string));
 
     // (a) Assigned template tasks.
     const tasks = await ctx.db
@@ -915,11 +918,11 @@ export const getMyServingTasks = query({
       // Team-level tasks (no roles) are team-wide, single-source on
       // `sharedTaskCompletions`, and belong to the Shared surface only — they
       // are no longer listed in this personal "Mine" view. A role task appears
-      // here when the user is confirmed for ANY of its roles (plus their
-      // personal tasks).
+      // here when the user is assigned (non-declined) to ANY of its roles (plus
+      // their personal tasks).
       const roleIds = taskRoleIds(task);
       if (roleIds.length === 0) continue;
-      if (!roleIds.some((r) => confirmedRoleIds.has(r as string))) continue;
+      if (!roleIds.some((r) => myRoleIds.has(r as string))) continue;
 
       const base = {
         title: task.title,
@@ -1003,8 +1006,14 @@ async function resolveUserName(
   return `${u?.firstName ?? ""} ${u?.lastName ?? ""}`.trim() || "Someone";
 }
 
-/** The current user's confirmed role assignments on a plan. */
-async function myConfirmedAssignments(
+/**
+ * The current user's *serving* role assignments on a plan: everything they
+ * haven't declined (unconfirmed or confirmed). An assigned-but-unconfirmed
+ * volunteer can now enter serving mode and reach the same day-of resources, so
+ * the serving surfaces gate on "non-declined" — the same predicate team chat and
+ * leader fill counts already use — not on "confirmed".
+ */
+async function myServingAssignments(
   ctx: QueryCtx | MutationCtx,
   planId: Id<"eventPlans">,
   userId: Id<"users">,
@@ -1014,7 +1023,7 @@ async function myConfirmedAssignments(
     .withIndex("by_plan", (q) => q.eq("planId", planId))
     .filter((q) => q.eq(q.field("userId"), userId))
     .collect();
-  return rows.filter((a) => a.status === "confirmed");
+  return rows.filter((a) => a.status !== "declined");
 }
 
 /**
@@ -1042,17 +1051,18 @@ type SharedTeamTaskItem = {
 
 /**
  * The "Shared" surface: team-level tasks (roleId == null) for the CURRENT
- * user's confirmed team(s) on a plan. These are "the whole team is responsible,
- * no single assignee" — so completion is TEAM-WIDE (`sharedTaskCompletions`),
- * not per-user: any confirmed teammate marking it done marks it done for
- * everyone. Completion is single-source on `sharedTaskCompletions` — a task
+ * user's serving team(s) (non-declined) on a plan. These are "the whole team is
+ * responsible, no single assignee" — so completion is TEAM-WIDE
+ * (`sharedTaskCompletions`), not per-user: any serving teammate marking it done
+ * marks it done for everyone. Completion is single-source on
+ * `sharedTaskCompletions` — a task
  * shows done iff a shared row exists (pre-migration per-user completions were
  * snapshotted into `sharedTaskCompletions` by a one-time backfill).
  * Returned as a flat array (each item carries its `segment`), ordered
  * before → during → after, then by the task's sortOrder.
  *
  * Auth: an active member of the plan's group (community read gate). Only the
- * user's own confirmed teams are included.
+ * user's own serving teams (non-declined) are included.
  */
 export const getSharedTeamTasks = query({
   args: { token: v.string(), planId: v.id("eventPlans") },
@@ -1065,8 +1075,8 @@ export const getSharedTeamTasks = query({
     if (!plan) return [];
     await requireGroupMember(ctx, plan.groupId, userId);
 
-    const confirmed = await myConfirmedAssignments(ctx, args.planId, userId);
-    const myTeamIds = new Set(confirmed.map((a) => a.teamId as string));
+    const serving = await myServingAssignments(ctx, args.planId, userId);
+    const myTeamIds = new Set(serving.map((a) => a.teamId as string));
     if (myTeamIds.size === 0) return [];
 
     const [tasks, sharedRows] = await Promise.all([
@@ -1154,7 +1164,8 @@ export const getSharedTeamTasks = query({
  * marking done writes only the shared row, but un-checking also deletes any
  * legacy `eventTaskCompletions` rows for the task.
  *
- * Auth: the caller must be a confirmed member of the task's team on the plan.
+ * Auth: the caller must be serving (non-declined — unconfirmed or confirmed) on
+ * the task's team on the plan.
  */
 export const toggleSharedTeamTask = mutation({
   args: {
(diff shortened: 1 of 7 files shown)


apps/convex/functions/scheduling/serving.ts
```diff
@@ -210,20 +216,20 @@ export const getServingEligibility = query({
...
+      // The user's assignments on this plan (any status), from the set above.
+      const planAssignments = nonDeclined.filter(
+        (a) => (a.planId as string) === planIdStr,
+      );
+      // Auto-entry stays confirmed-only: only drop a *confirmed* volunteer
+      // straight into serving mode. Unconfirmed people are eligible but must
+      // tap in themselves.
+      const hasConfirmed = planAssignments.some((a) => a.status === "confirmed");
+      const autoEnter =
+        hasConfirmed && now >= startsAt - AUTO_ENTER_LEAD_MS && now <= endsAt;
```
lilseyi, line 237: **Blocking — confirmed auto-entry regresses in the multi-plan case (spec: "a confirmed volunteer's auto-entry behavior is unchanged").**

The *per-entry* `autoEnter` here is correctly gated on `hasConfirmed`. The problem is the *global* `autoEnter` at line 260:

```ts
const autoEnter = entries.length === 1 && entries[0].autoEnter;
```

`entries` is now built from the **non-declined** plan set (line 204), not the confirmed-only set it used before. So an unconfirmed-only plan now inflates `entries.length`.

Scenario (the exact "two services / two campuses same morning" case this function's own comments cite): a volunteer **confirmed on plan A** who *also* holds an **unconfirmed** assignment on plan B, same day, both in-window.
- **Before:** confirmed-only query → `planIds = [A]` → `entries.length === 1` → auto-entered into A. ✅
- **After:** `planIds = [A, B]` → `entries.length === 2` → global `autoEnter = false`. The confirmed volunteer is **no longer auto-entered.** ❌

An unrelated unconfirmed assignment on a different plan silently suppresses a confirmed volunteer's auto-entry, which the spec lists as a hard invariant.

Suggested fix: gate the global `autoEnter` on the count of *auto-eligible* entries (confirmed + inside the auto-enter window) being exactly 1 — e.g. `const autoEntries = entries.filter((e) => e.autoEnter); const autoEnter = autoEntries.length === 1;` — rather than on the total eligible-plan count. Please also add a multi-plan test pinning this (no eligibility test currently exercises >1 plan).

---

lilseyi (commented):
## Automated review — verdict: changes requested

Reviewed the diff against the approved spec with parallel correctness / security / spec-fidelity / test subagents, then adversarially verified each finding. The core predicate flip (confirmed → non-declined), the two "Unconfirmed" badges, per-(user,role) status, empty-state copy, and the `ChatInboxScreen` chip/auto-enter split are all correctly implemented, and **security is clean**. One blocking issue and a few minor ones remain.

### 🔴 Blocking
1. **Confirmed auto-entry regresses in the multi-plan case** — inline on `serving.ts:226-228`. The global `autoEnter = entries.length === 1 && entries[0].autoEnter` (line 260) now counts unconfirmed-only plans, so a volunteer *confirmed* on plan A who also has an *unconfirmed* assignment on a same-day plan B is no longer auto-entered into A. The spec lists confirmed auto-entry as unchanged, so this needs a fix + a multi-plan test.

### 🟡 Minor (worth addressing before merge)
2. **`eventTasks.ts:497` still confirmed-only — the "~497" spot the spec flagged, missed.** In `updateTask`'s orphaned-completion cleanup, `coveredUserIds` filters `a.status === "confirmed"`. Since unconfirmed volunteers can now hold `eventTaskCompletions` rows (`toggleTaskCompletion` gates only on group membership; `getMyServingTasks` now surfaces their role tasks), removing an unrelated role from a multi-role task wrongly deletes an unconfirmed-but-still-assigned user's completion — contradicting that branch's own contract. Fix: `a.status !== "declined"` to match the rest of the PR. (Couldn't anchor this inline — the line is outside the diff.)

3. **Run sheet includes unconfirmed but doesn't mark them.** `ServingRunsheetScreen` renders no crew list of its own (PR's claim is narrowly true), but it embeds `PlanRunSheet` / `NativeRunSheetView`, which lists role→volunteer names already filtered `status !== "declined"` — so unconfirmed volunteers appear there with **no signifier**. The spec's "Done when" names the run sheet as a surface to mark. There's real tension (that component is the shared *leader* Rostering view, and the spec forbids changing leader-facing UI; it's also a comma-joined name string, not per-person cards), so skipping is defensible — but it's a literal gap, so please confirm this is an intentional scoping decision rather than an oversight.

### 🟡 Test coverage gaps
- No test exercises **>1 eligible plan**, which is exactly the finding-1 regression case.
- **`toggleSharedTeamTask`** changed from a confirmed-only to a non-declined auth gate on a mutation, but no test pins that an *unconfirmed* teammate is now admitted and a *declined* member is still rejected (the existing reject test uses an *unassigned* user, which rejects regardless).
- **`getServingUpcomingChannels`** (confirmed → non-declined) has zero coverage.

### ℹ️ Spec call-out (as the spec requested)
- **Phone-number exposure now extends to unconfirmed same-event teammates** (`serving.ts:549`, for the day-of "Text" action). This matches the approved decision to treat unconfirmed as serving and is bounded to same-event roster members — flagged as intended, not blocking.

Backend tests pass locally (serving.test.ts + eventTasks.test.ts, 30/30).

---

apps/convex/functions/scheduling/serving.ts
```diff
@@ -445,14 +466,14 @@ export const getServingTeamRoster = query({
...
+    const myAssignments = await ctx.db
       .query("roleAssignments")
-      .withIndex("by_user_status", (q) =>
-        q.eq("userId", userId).eq("status", "confirmed"),
-      )
+      .withIndex("by_user", (q) => q.eq("userId", userId))
       .collect();
-    const planIds = [...new Set(confirmed.map((a) => a.planId as string))];
+    const nonDeclined = myAssignments.filter((a) => a.status !== "declined");
+    const planIds = [...new Set(nonDeclined.map((a) => a.planId as string))];
```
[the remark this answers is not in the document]
lilseyi: Fixed in 6bcd0c3. Confirmed the gap: `groupMembers.remove` only soft-deletes (sets `leftAt`) and leaves the `roleAssignments` row intact, so a former member with a never-answered (unconfirmed) assignment would pass the widened non-declined selection and reach the roster (and teammate phone numbers).

Added an `isActiveGroupMember(ctx, groupId, userId)` boolean helper in `permissions.ts` (mirrors `requireGroupMember`'s active-member + community-admin logic; `requireGroupMember` now delegates to it). Both `getServingEligibility` and `getServingTeamRoster` now skip any plan whose group the caller has left, before returning anything. Covered by a new test — a former group member with a stale unconfirmed assignment gets `eligible: false` and an empty roster.

---

apps/convex/functions/scheduling/serving.ts
lilseyi: Fixed in 6bcd0c3, exactly as suggested. The global `autoEnter` now gates on the count of *auto-eligible* entries rather than the total non-declined plan count:

```ts
const autoEntries = entries.filter((e) => e.autoEnter);
const autoEnter = autoEntries.length === 1;
```

So a volunteer confirmed on plan A who also holds an unconfirmed assignment on same-day plan B is auto-entered into A again (only A is auto-eligible → `autoEntries.length === 1`). Added two multi-plan tests: confirmed+unconfirmed → `eligible: true, autoEnter: true, plans: 2`, and two confirmed same-day plans → `autoEnter: false` (ambiguous), pinning that the fix didn't loosen the multi-confirmed case.

---

apps/convex/functions/scheduling/eventTasks.ts
```diff
@@ -494,7 +494,12 @@ export const updateTask = mutation({
           assignments
             .filter(
               (a) =>
-                a.status === "confirmed" && nextSet.has(a.roleId as string),
+                // Non-declined (not confirmed-only): unconfirmed volunteers can
+                // now hold completions on their assigned role's tasks, so a
+                // still-assigned unconfirmed user must count as covered — else
+                // removing an unrelated role would wrongly orphan their
+                // completion. Matches the rest of this PR's predicate.
+                a.status !== "declined" && nextSet.has(a.roleId as string),
```
lilseyi, line 502: Non-blocking (test coverage). This predicate change (`confirmed` → `!== "declined"`) decides which users keep an `eventTaskCompletion` when a role is removed from a multi-role task. It's now genuinely reachable — an unconfirmed volunteer can hold a completion via the widened `getMyServingTasks` — but no test distinguishes the new behavior from the old: the existing `updateTask` orphan-cleanup test (`eventTasks.test.ts`, "drops completions from users no longer covered when a role is removed", ~line 1001) uses only *confirmed* users, so it passes identically under both predicates. Reverting this line would go uncaught.

Suggested test: task with roles `[r1, r2]`; an **unconfirmed** user assigned to `r2` completes it; leader removes `r1`; assert their completion is **kept**. Not blocking — the change itself is correct (it deletes fewer completions, the conservative direction).

---

lilseyi (commented):
**Automated review — parallel reviewers (correctness, security, spec-fidelity/UX, test adequacy), each adversarially self-verified.**

Verdict: **approved** — no blocking findings.

- **Correctness** ✔ `by_user` index exists on `roleAssignments`; the new `autoEnter = autoEntries.length === 1` logic correctly preserves confirmed auto-entry (incl. a confirmed volunteer who also holds an unrelated same-day *unconfirmed* assignment) and never auto-enters an unconfirmed-only volunteer; the `isActiveGroupMember`/`requireGroupMember` refactor is behavior-identical.
- **Security** ✔ The critical risk — a soft-removed former member (`leftAt` set) with a stale *unconfirmed* assignment regaining serving access + teammate phone numbers — is blocked: `getServingEligibility` and `getServingTeamRoster` add the per-plan `isActiveGroupMember` gate, and every other widened surface (`getMyServingTasks`, `getCrewTasks`, `getSharedTeamTasks`, `toggleSharedTeamTask`, `getServingUpcomingChannels`) already gates on `requireGroupMember`. `toggleSharedTeamTask` still rejects declined/unassigned callers. Phone exposure widening is the spec-sanctioned same-event extension.
- **Spec-fidelity/UX** ✔ Web-parity claim holds (no `.web.tsx` variants of the serving screens exist — they're single cross-platform files); `ServingRunsheetScreen` is a launcher with no teammate list, so it correctly needs no per-row marking; badge uses `colors.warning` + clock, only on unconfirmed; both empty-state strings updated; `ChatInboxScreen` correctly unchanged (chip keys off `eligible`, auto-enter off `autoEnter`); leader-facing fill math (`roster.ts`/`events.ts`) untouched.
- **Tests** ✔ `serving.test.ts` + `eventTasks.test.ts` run green (36 passing); headline behaviors well covered with meaningful (non-tautological) assertions.

Two **non-blocking** notes:
1. Inline: the `updateTask` "covered users" predicate change has no test that distinguishes the new behavior from the old.
2. Minor doc nit: the header docstring in `apps/mobile/features/serving/components/ServingTeamScreen.tsx` (~lines 10–15) still describes "a card per **confirmed** volunteer" / "confirmed assignments" — now stale since the screen shows unconfirmed teammates too. Worth a one-line touch-up.

Neither blocks merge.

---

PR #633, opened 2026-07-20
Fix staging login crash on archived-community tokens + make "Send to developer" actually deliver

lilseyi:
## Summary

Fixes the two staging-blocking bugs reported today. Both pre-date the link-preview PR (#630) — root causes traced to PR #588's archival hardening and the original mailto-based error reporting.

### 1. `COMMUNITY_ARCHIVED` hard crash on login/boot
A JWT scoped to an archived community crashed the whole app: `NotificationProvider` mounts an always-on `unreadCount` query the moment a token exists, `requireAuth` throws `COMMUNITY_ARCHIVED`, and the only boundary above it was the app-root `ErrorBoundary` (outside `AuthProvider`) — fatal screen, no recovery path.

**Two-layer fix:**
- **Server**: `list`/`inboxSummary`/`unreadCount` (the queries mounted unconditionally at boot) now use a new `requireAuthWithArchivedStatus` helper and return benign empty data for archived-community tokens — notification counts for an inaccessible community are meaningless.
- **Client**: new `AuthErrorBoundary` (below `AuthProvider`, wrapping the query tree in `_layout.tsx`) catches `COMMUNITY_ARCHIVED` from **any** query, calls the existing `exitToCommunitySelection()` (re-mints a community-less token, wipes per-community caches), and routes to community selection. Every other error re-throws to the root boundary unchanged. This fills in the `AuthErrorBoundary` component that backend comments already referenced by name.

### 2. "Send to developer" never sent anything
`expo-mail-composer` on web just `window.open`s a `mailto:` and unconditionally reports success — nothing is transmitted, popup blockers eat it silently. Replaced with an unauthenticated Convex action `functions/support/sendErrorReport` sending via Resend (mirrors the `emailOtp` pattern: lazy client init, server-side recipient, subject/body length caps), and the ErrorBoundary now awaits the real result and shows a distinct failed state.

Also includes hand-applied `_generated/api.d.ts` entries for the new support module (matches deterministic codegen output; this environment has no Convex deployment to run codegen against).

## Test plan
- [x] Convex: full suite 146 files green, including 4 new archived-community query tests and 11 new `sendErrorReport` tests
- [x] Mobile: `tsc --noEmit` clean apart from 20 pre-existing errors in files untouched by this branch
- [x] `convex typecheck` clean
- [ ] On staging after deploy: log in with a token scoped to an archived community → should land on community selection, not the crash screen; trigger the error screen and "Send to developer" → email arrives at the support inbox

https://claude.ai/code/session_013VHVFzj5sj8uq91cP7389D

---
_Generated by [Claude Code](https://claude.ai/code/session_013VHVFzj5sj8uq91cP7389D)_

### DIFF
--- a/apps/convex/functions/support/sendErrorReport.ts
+++ b/apps/convex/functions/support/sendErrorReport.ts
@@ -4,8 +4,8 @@
  *
  * Delivers "Send to developer" crash reports from the app-root ErrorBoundary
  * (apps/mobile/components/ErrorBoundary.tsx) to the Togather support inbox
- * via Resend. Mirrors the lazy-Resend-client / RESEND_API_KEY pattern in
- * auth/emailOtp.ts.
+ * via Resend. Uses the shared lazy-Resend-client from `lib/resend.ts` (also
+ * used by auth/emailOtp.ts).
  *
  * Must be callable WITHOUT authentication: the ErrorBoundary renders above
  * AuthProvider, so a crash before/during auth setup still needs to be able
@@ -13,10 +13,10 @@
  */
 
 import { v } from "convex/values";
-import { Resend } from "resend";
 import { action } from "../../_generated/server";
 import { internal } from "../../_generated/api";
 import { DOMAIN_CONFIG } from "@togather/shared/config";
+import { getResendClient } from "../../lib/resend";
 
 /**
  * Where crash reports are delivered. Sourced here (server-side) rather than
@@ -29,34 +29,43 @@ const EMAIL_FROM = DOMAIN_CONFIG.emailFrom;
 /**
  * Abuse guards.
  *
- * This action is unauthenticated by design (see file header), so unlike
- * the OTP flows in auth/phoneOtp.ts / auth/emailOtp.ts there's no user
- * identity (phone/email) to key a per-caller rate limit on — the existing
- * `lib/rateLimit.ts` helper (checked before adding this) is generic and
- * reused here, just keyed on a single deployment-wide bucket instead of a
- * per-user one. The cap is intentionally generous: a real incident causing
- * many simultaneous crashes is exactly when reports matter most, so this
- * only needs to stop a runaway loop or scripted flood, not throttle normal
- * usage.
+ * This action is unauthenticated by design (see file header), so unlike the
+ * OTP flows in auth/phoneOtp.ts / auth/emailOtp.ts there's no user identity
+ * (phone/email) to key a rate limit on. Instead the client sends a
+ * `reportKey` — a random id it generates once and persists in AsyncStorage
+ * (see ErrorBoundary.tsx). This key is entirely client-supplied and
+ * therefore spoofable/forgeable by an attacker; it does NOT provide real
+ * authentication. Its purpose is narrower: it lets honest clients (the vast
+ * majority of callers) get their own rate bucket instead of all sharing one
+ * global bucket that a single scripted client could exhaust for everyone.
+ * An attacker who rotates the key on every request just falls back to the
+ * global backstop below, same as before this change.
  */
-const RATE_LIMIT_KEY = "error-report:global";
-const RATE_LIMIT_MAX_ATTEMPTS = 100;
-const RATE_LIMIT_WINDOW_MS = 60 * 60 * 1000; // 1 hour
+const PER_KEY_RATE_LIMIT_MAX_ATTEMPTS = 5;
+const PER_KEY_RATE_LIMIT_WINDOW_MS = 60 * 60 * 1000; // 1 hour
+
+const GLOBAL_RATE_LIMIT_KEY = "error-report:global";
+const GLOBAL_RATE_LIMIT_MAX_ATTEMPTS = 500;
+const GLOBAL_RATE_LIMIT_WINDOW_MS = 60 * 60 * 1000; // 1 hour
 
 /** Server-side length caps — never trust client-supplied string sizes. */
 const MAX_SUBJECT_LENGTH = 200;
 const MAX_BODY_LENGTH = 20000;
+const MAX_REPORT_KEY_LENGTH = 64;
 
-// Lazy-initialize Resend client to avoid throwing if API key is missing at module load
-let resendClient: Resend | null = null;
-function getResendClient(): Resend | null {
-  if (!process.env.RESEND_API_KEY) {
-    return null;
-  }
-  if (!resendClient) {
-    resendClient = new Resend(process.env.RESEND_API_KEY);
-  }
-  return resendClient;
+/** Fixed prefix applied to every outgoing subject, server-side. */
+const SUBJECT_PREFIX = "[Togather error report] ";
+
+/**
+ * Bucket key for a client-supplied `reportKey`. Sanitized to a conservative
+ * character set and capped in length so a malformed/oversized/adversarial
+ * key can't be used to smuggle data into the rate-limit key or blow up the
+ * `rateLimits` table with garbage keys.
+ */
+function sanitizeReportKey(rawKey: string | undefined): string {
+  const trimmed = (rawKey ?? "").trim().slice(0, MAX_REPORT_KEY_LENGTH);
+  const sanitized = trimmed.replace(/[^a-zA-Z0-9_-]/g, "");
+  return sanitized.length > 0 ? sanitized : "anonymous";
 }
 
 /**
@@ -74,8 +83,15 @@ export function capLength(value: string, maxLength: number): string {
 
 /**
  * Builds the Resend payload from raw client input, applying the server-side
- * length caps and fixed from/to addresses. Pulled out as a pure function so
- * the capping/payload logic is unit-testable without a live Resend client.
+ * length caps, a fixed subject prefix, and fixed from/to addresses. Pulled
+ * out as a pure function so the capping/payload logic is unit-testable
+ * without a live Resend client.
+ *
+ * The fixed `SUBJECT_PREFIX` is not optional/skippable by the client — it
+ * guarantees every email this action can ever send is unambiguously
+ * labeled as a self-serve crash report, so this unauthenticated endpoint
+ * can't be abused to send mail from our trusted sender with an arbitrary
+ * subject.
  */
 export function buildErrorReportEmail(args: { subject: string; body: string }): {
   from: string;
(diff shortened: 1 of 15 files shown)


apps/mobile/providers/AuthErrorBoundary.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  private recover = async () => {
+    try {
+      await this.props.onCommunityArchived();
+    } catch (recoveryError) {
+      console.error('[AuthErrorBoundary] Recovery failed:', recoveryError);
+    }
+    // The archived community is no longer selected (token re-minted,
+    // per-community caches wiped) — send the user back to pick a community.
+    router.replace('/(auth)/select-community');
```
lilseyi: **[CONFIRMED — critical] `router.replace()` on a torn-down navigator → permanent blank screen on every recovery.**

While `isRecovering` is true, `render()` returns `null`, unmounting the entire subtree — including `ThemedStack`, the app's only Navigator. `router.replace` then hits expo-router's `assertIsReady()` with `navigationRef.isReady() === false` and throws "Attempted to navigate before mounting the Root Layout component". That throw is *outside* the `try/catch` (only `onCommunityArchived()` is wrapped), so `setState({ isRecovering: false })` on the next line never runs — the boundary renders `null` forever. This reproduces on the *success* path, not just edge cases: strictly worse than the crash screen this PR replaces.

Fix direction: never call the imperative router while the boundary is showing its fallback. Complete the token re-mint first, then clear the boundary state to remount the tree (fresh community-less token → queries no longer throw) and hand off routing to a mechanism that runs after the navigator is ready (e.g. AuthGuard's existing redirect logic, or a deferred navigation once `navigationRef.isReady()`).

---

apps/mobile/providers/AuthErrorBoundary.tsx
```diff
@@ -0,0 +1,100 @@
...
+      '[AuthErrorBoundary] Recovering from archived-community session:',
+      error,
+    );
+    this.recover();
+  }
+
+  private recover = async () => {
+    try {
+      await this.props.onCommunityArchived();
+    } catch (recoveryError) {
```
lilseyi, line 181: **[CONFIRMED — correctness] Failed recovery is swallowed → unbounded catch/recover loop with the poisoned token.**

Two compounding issues: (1) a throw from `onCommunityArchived()` is only `console.error`'d and recovery proceeds as if it succeeded; (2) `exitToCommunitySelection` itself *never throws* — `refreshAuthTokens()` catches internally and returns `false` — so on a transient network failure the archived-scoped token remains active, the remounted `unreadCount` query throws `COMMUNITY_ARCHIVED` again, and the boundary re-enters `recover()` in a visible flicker loop until connectivity returns.

Fix direction: check the re-mint actually succeeded (return value / token no longer community-scoped) before clearing the boundary; on failure show an explicit retry state instead of remounting; cap attempts (e.g. 2) and re-throw to the root ErrorBoundary beyond that. Also render a minimal "Switching community…" view rather than `null` so a slow recovery isn't a blank app.

---

apps/convex/functions/support/sendErrorReport.ts
```diff
@@ -0,0 +1,153 @@
...
+ * the OTP flows in auth/phoneOtp.ts / auth/emailOtp.ts there's no user
+ * identity (phone/email) to key a per-caller rate limit on — the existing
+ * `lib/rateLimit.ts` helper (checked before adding this) is generic and
+ * reused here, just keyed on a single deployment-wide bucket instead of a
+ * per-user one. The cap is intentionally generous: a real incident causing
+ * many simultaneous crashes is exactly when reports matter most, so this
+ * only needs to stop a runaway loop or scripted flood, not throttle normal
+ * usage.
+ */
+const RATE_LIMIT_KEY = "error-report:global";
```
lilseyi: **[CONFIRMED — abuse/DoS] Single global rate bucket lets any client disable crash reporting deployment-wide.**

(Matches Codex's P1.) The action is unauthenticated and every call shares `error-report:global` (100/hour) — a scripted client, or simply a genuine mass incident affecting >100 users, exhausts it for everyone for the rest of the hour, exactly when reports matter most. Related: the free-text subject/body with the trusted `EMAIL_FROM` makes the inbox a spam vector.

Fix direction: per-caller bucket keyed on a client-supplied install/user id (spoofable, but forces key rotation and rate-limits honest clients correctly) at ~5/hour, plus a much higher global backstop (~500/hour); force a fixed server-side subject prefix (e.g. `[App error report]`) so fabricated content can't masquerade as arbitrary mail.

---

apps/mobile/components/ErrorBoundary.tsx
```diff
@@ -163,24 +159,14 @@ function DefaultErrorFallback({
...
-        });
-      } else {
-        // Fall back to the device's default mail handler.
-        const mailto = `mailto:${SUPPORT_EMAIL}?subject=${encodeURIComponent(
-          subject,
-        )}&body=${encodeURIComponent(body)}`;
-        await Linking.openURL(mailto);
-      }
-      setSendState('sent');
+      const result = await convexVanilla.action(
```
lilseyi: **[CONFIRMED — native regression] Replacing MailComposer everywhere loses a working native path.**

On native, `MailComposer.composeAsync` opened the OS mail app with the report pre-filled but *user-editable* (redaction/context) and *OS-queued* (delivers after connectivity returns). The web `mailto:` path was genuinely broken, but native wasn't — and the new server action requires a live round-trip at exactly the moment (post-crash, possibly offline) delivery is least reliable, with no queue.

Also: the single failure message ("check your connection") conflates rate-limiting and server misconfiguration (missing `RESEND_API_KEY`) with client connectivity — a Resend outage would send every user chasing their wifi.

Fix direction: platform-split — keep MailComposer on native, use the Convex action on web only; and have the action return a failure reason so the client can show an accurate message.

---

apps/mobile/services/api/convex.ts
```diff
@@ -435,6 +435,40 @@ export function useAuthenticatedPaginatedQuery<
...
+ * error-handling idiom in this codebase (see SongLibraryScreen.tsx,
+ * app/inbox/[groupId]/create.tsx) and avoids a hard dependency on exactly
+ * which `convex` build reconstructs the error.
+ */
+export function isCommunityArchivedError(error: unknown): boolean {
+  if (!error || typeof error !== 'object') return false;
+  const data = (error as { data?: unknown }).data;
+  if (data === COMMUNITY_ARCHIVED_CODE) return true;
+  const message = (error as { message?: unknown }).message;
+  return typeof message === 'string' && message.includes(COMMUNITY_ARCHIVED_CODE);
```
lilseyi: **[PLAUSIBLE — robustness] Loose substring fallback can misclassify unrelated errors as recoverable.**

`message.includes("COMMUNITY_ARCHIVED")` matches any error whose message merely *contains* the code (including future superstring codes) — misclassified errors get silently swallowed into the community-switch flow instead of surfacing to the root boundary/Sentry. Tighten to an exact `error.data === "COMMUNITY_ARCHIVED"` check with an anchored/word-boundary message fallback.

---

apps/convex/lib/auth.ts
```diff
@@ -559,6 +559,25 @@ export async function requireAuthAllowArchivedCommunity(
   return userId;
 }
 
+/**
+ * Like {@link requireAuthAllowArchivedCommunity} but also reports whether the
+ * token's community is archived, for the handful of queries mounted
+ * unconditionally at app boot (e.g. notification badge counts). Those queries
+ * can't reject with COMMUNITY_ARCHIVED — convex/react re-throws query errors
+ * during render, and these mount above any recovery UI — so they use this to
```
lilseyi: **[CONFIRMED — stale rationale] "these mount above any recovery UI" is false as of this same PR.**

`AuthErrorBoundary` (added in this PR) wraps all four call sites of these queries, so recovery UI *does* exist below them now. The two layers still compose correctly — the boundary is the general mechanism; the query short-circuits just prevent crash-recovery churn on boot — but as written this comment will teach the next engineer that server-side special-casing is the *required* pattern for new boot queries. Reword here (and at the three short-circuits in `notifications/queries.ts`) to present the boundary as the primary fix and this as intentional defense-in-depth that new queries don't need to copy. Also worth a line: `getResendClient` in `support/sendErrorReport.ts` is copy-pasted from `emailOtp.ts` — extract a shared `lib/resend.ts` helper.

---

lilseyi (commented):
## Claude review — 8 angles × 5 finder agents, cross-checked against Codex

**6 findings posted inline** (1 critical, 3 correctness/abuse, 1 robustness, 1 stale-rationale/cleanup). The critical one: the `AuthErrorBoundary` recovery flow calls `router.replace()` after its own fallback render has unmounted the app's only Navigator, which throws uncaught and leaves the boundary stuck rendering `null` — a permanent blank screen on the recovery *success* path. The recovery redesign, per-caller rate limiting (concurring with Codex's P1), the native MailComposer platform-split, and the smaller items are all being fixed on this branch now; threads will be resolved as fixes land.

Verified non-issues for the record: the archived-community short-circuits leak no data (early return precedes all db reads), and the auth core (`requireAuthResolved` — expiry/revocation/validation) is untouched.

---

apps/convex/functions/support/sendErrorReport.ts
```diff
@@ -0,0 +1,153 @@
...
+    body: v.string(),
+  },
+  handler: async (ctx, args): Promise<{ success: boolean }> => {
+    try {
+      await ctx.runMutation(
+        internal.functions.authInternal.checkRateLimitInternal,
+        {
+          key: RATE_LIMIT_KEY,
+          maxAttempts: RATE_LIMIT_MAX_ATTEMPTS,
+          windowMs: RATE_LIMIT_WINDOW_MS,
```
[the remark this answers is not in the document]
lilseyi: Fixed in 361a36c: per-caller bucket keyed on a client-persisted `reportKey` (sanitized server-side, 5/hour per key) plus the global bucket raised to 500/hour as a backstop. The key is client-supplied and spoofable by design — it exists to give honest clients independent budgets and force an attacker rotating keys into the global backstop. Also added a forced `[Togather error report]` subject prefix so the unauthenticated action can't send arbitrary-looking mail from the trusted sender. Covered by new tests (independent keys not cross-blocked; global backstop still trips with fresh per-key buckets).

---

PR #629, opened 2026-07-17
chore: migrate dev-assistant onto @supa-media/dev-assistant@2.0.0

lilseyi:
Migrates Togather's `devAssistant` module onto `@supa-media/dev-assistant@^2.0.0`.
The migration is **wired, verified, and behavior-neutral** — the upstream
package defect that previously blocked this PR (`v1.0.0`'s `createDevAssistant`
factory dropping every re-exported function from the strict-typechecked
consumer's generated `api`/`internal`) is fixed upstream in `v2.0.0`
(supa-framework PR #29: module-level config holder + module-level function
consts, factory removed).

## What changed

- **Install**: `@supa-media/dev-assistant@^2.0.0` via a filtered add (no root
  `pnpm install`; no React re-keying — see the dependency note below).
- **`_instance.ts` → `config.ts`**: `createDevAssistant({...})` →
  `setDevAssistantConfig({...})`. Every config seam (signatureHeader
  `x-togather-signature`, R2 resolvers, notifier, repo config incl.
  `confirm: "deploy"`, areas, auto-merge) is byte-identical.
- **`_reexportTypes.ts` deleted** — the cast shim is no longer needed. Every
  function in `bugs.ts`/`actions.ts`/`contributions.ts`/`maintainers.ts` now
  does `import "./config"` (side effect) then re-exports the package's
  module-level function consts directly from
  `@supa-media/dev-assistant/functions/<module>` — genuine builder outputs
  that survive Convex's generated `api`/`internal` with concrete visibility +
  args, **no cast**. Re-exports exactly match what the old wrappers surfaced;
  local functions (`createBug`, `processThreadMention`, maintainer role
  grant/revoke, the `@Togather` chat bot's DB ops) stay put and coexist on the
  api node.
- **`http.ts`**: `registerRoutes(http)` (was
  `devAssistant.registerRoutes(http)`).
- **`crons.ts`**: `registerDevAssistantCrons(crons)` reads `functionsPath`
  from the config holder (was `registerDevAssistantCrons(crons,
  devAssistant.config)`).
- **`schema.ts`**: unchanged — the `supaDevAssistantTables()` cast (documented
  upstream typing gap: the factory returns `Record<string, TableDefinition>`,
  erasing the concrete `devBugs`/`devBugMessages` keys) is independent of the
  mounting-contract fix and still needed.
- **New**: `__tests__/devAssistant-mount.test.ts` — the package's
  `assertMounted`/`validateMount` smoke test (README "Smoke test" recipe),
[...]

PR #646, opened 2026-07-29
fix(convex): pin react-dom to match react — verification emails were crashing in production

lilseyi:
## What

Verification emails (email OTP) were throwing in production:

```
Error: Incompatible React versions: The "react" and "react-dom" packages must have the exact same version. Instead got:
  - react:      19.1.0
  - react-dom:  19.2.4
```

This pins `react-dom@19.1.0` in `apps/convex/package.json` next to the existing load-bearing `react@19.1.0` pin, and adds guards so the skew can't silently recur.

## Root cause

PR #629's `react@19.1.0` devDependency pin re-keyed the react-email tree onto react 19.1.0 — but `react-dom` was declared nowhere, so pnpm auto-installed it as a transitive peer of `@react-email/render`/`resend` at the latest in-range version (19.2.4). The lockfile ended up with `/react-dom@19.2.4(react@19.1.0)`, and react-dom's server renderer hard-errors on exactly that mismatch (React >= 19.2, `ensureCorrectIsomorphicReactVersion`).

It shipped with CI green because the verification email is the **only** backend path that executes react-dom (`@react-email/render` in `functions/auth/emailOtp.ts`) — mention/notification emails are plain HTML strings — and nothing in CI ever rendered it. Neither existing guard covered this: the `devAssistant-mount` test guards the convex-type duplication, and `check-react-consistency@1.1.0` only guards the native module graph.

## Changes

- **Fix:** `react-dom: "19.1.0"` devDependency pin in `apps/convex`, added via scoped `pnpm add --filter convex-functions` per the CLAUDE.md rule. The skewed lockfile pair is gone (`@react-email/*` now keyed `(react-dom@19.1.0)(react@19.1.0)`); the native graph diff is zero lines.
- **Runtime backstop:** new `apps/convex/__tests__/email-render.test.ts` renders the real `VerificationCode` template through the real react-dom — the exact call that crashed — so any future skew fails CI instead of the login flow.
- **Docs:** CLAUDE.md's load-bearing-pin section now covers both pins, the failure mode, and the backstops.

A companion PR (Supa-Media/supa-framework#claude/react-version-mismatch-gqpxpg) adds a lockfile-wide react-dom/react exact-pairing gate to `check-react-consistency`; once 1.2.0 releases, `pnpm update "@supa-media/*"` here activates it as the static guard.

## Testing

- Full `apps/convex` suite: 149 files / 2451 tests green (including the new render test).
- `check-react-consistency` (with the new gate): all three gates pass post-fix; the new gate correctly failed on the pre-fix lockfile.
- Pre-push hook ran the full mobile Jest suite: 160 suites / 1478 tests green — confirms the scoped add didn't disturb the native React graph.

https://claude.ai/code/session_013ScbYySAW8nAg2bCopottA

---
_Generated by [Claude Code](https://claude.ai/code/session_013ScbYySAW8nAg2bCopottA)_


lilseyi (commented):
**Adversarially-verified review** (parallel Claude review agents: correctness / lockfile safety / test validity / spec-fidelity):

**Verified clean:**
- **Lockfile surgery:** only the `@react-email/*` / `resend` trees re-keyed; zero `expo*` / `react-native*` / `expo-modules-core` key changes; no dangling refs; remaining react-dom entries are exactly paired (`19.1.0(react@19.1.0)`, `19.2.4(react@19.2.4)` — the latter is apps/web's, untouched). `check-react-consistency` passes as run in mobile CI.
- **Test validity:** the new render test resolves `react-dom` through the same peer-keyed pnpm instance the Convex bundler uses (verified via `createRequire` realpath probe), and `ensureCorrectIsomorphicReactVersion` exists in every server build of react-dom — so the test provably throws under the old skew; not a vacuous pass. `test-convex`'s change filter includes `pnpm-lock.yaml`, so lockfile-only re-keys still trigger it.
- **Scope claim:** confirmed by grep that `functions/auth/emailOtp.ts` is the only backend path executing react-dom.

**Findings (2 low, both fixed in 8b7cd56):**
1. CLAUDE.md described the native-safety 1.2.0 lockfile gate as an active backstop, but this repo is on 1.1.0 — reworded to make explicit the render test is the only live guard until the bump.
2. "React >= 19.2" version bound was wrong — react-dom 19.1.0's server builds also throw on skew; corrected in CLAUDE.md and the test docblock.

No high/medium findings.

---

PR #654, opened 2026-07-29
Add event check-in: take attendance from the RSVP list

lilseyi:
## What this does

Gives event organizers a fast, tap-to-check-in way to take attendance at an event by going down the list of people who RSVPed **Going**. Requested first for **Marketplace Mandate**, but it works for every event — nothing is hard-coded to one event.

Think of it as a door list: the organizer opens the event, taps **Check in**, and sees everyone who said they're coming. Tapping a person marks them arrived (with a timestamp); a running **"N / M checked in"** count with a progress bar sits at the top. Walk-ins who don't have a Togather account can be added by name so they're counted too.

Most of the plumbing already existed, so this is one focused piece rather than a new system: a check-in records the same **Present** attendance the app already uses, so member-health / attendance numbers stay correct — nothing is tracked twice.

## Before / after

The current **Guest List** stays as the view-only roster; **Check in** is the new interactive attendance mode reached from the same leader event view.

[Before/after: Guest List vs Check-in screen](https://images.togather.nyc/dev-assistant/30e64bb2-26d8-4fd4-9922-233d54e97001-checkin-before-after.png)

> ⚠️ **Rendered mock, not a device capture.** This run executes on a headless Linux runner with no iOS simulator, so the image above is a faithful HTML render built from the real screen's styles/labels. Please spot-check on a device/staging OTA before merge (no native-media surfaces are touched, so this is low native risk).

## How it behaves

- **Tap to check in / undo.** Tapping a Going person marks them Present; tapping again undoes it. Each change saves immediately and the top count updates live. Two leaders checking in at once see each other's taps (shared live data); re-tapping is safe (upsert of the one Present record, never a duplicate).
- **Walk-ins.** **Add walk-in** asks for a first name (last name / phone optional) and adds them already checked in, reusing the existing guest record so no account is required. Walk-ins can be removed if added by mistake.
- **Who's on the list.** Everyone who RSVPed Going, plus any walk-ins. Maybe / Can't Go aren't pre-listed; if one shows up, the organizer adds them as a walk-in.
- **Edge cases.** RSVP disabled / zero Going → empty state with just Add walk-in (headcount still works). Undo down to 0 / M is fine.

## Permissions

Only event managers — the event's host(s), the group's leaders, or a community admin — can open Check-in or use its actions. This is the exact `canEditMeeting` rule the app already uses for editing an event and marking attendance. Enforced **server-side**, not just hidden in the UI.

## What changed

**Backend** (`apps/convex/functions/meetings/attendance.ts`, reuse-first):
- Gated `addGuest` with `canEditMeeting` so walk-in adds are manager-only — it previously had no permission check, unlike `removeGuest`/`updateGuest`. Its only caller is the leader attendance flow, so no behavior regresses.
- Added a `canManageAttendance` query so the screen/entry point reflect the same rule the mutations enforce.
- Reuses existing `markAttendance` (Present = status `1`, undo = `0`, upsert), `listAttendance`, `listGuests`, `removeGuest`, and the RSVP roster from `meetingRsvps.list`.

**Frontend:**
- New `checkin.tsx` screen (mirrors the existing `guests.tsx` patterns) with the summary card, tappable Going roster, walk-ins section, and an Add-walk-in modal.
- Extracted pure `computeCheckInSummary` / `indexAttendanceByUser` helpers so the live count is unit-testable.
- Added a **Check in** entry point on the leader event view next to Guest List.

**Docs:** Events onboarding guide (`apps/web/src/pages/guides/Events.tsx`) gains a Check-in section + mock.

## Tests

- **Convex** (`__tests__/event-checkin.test.ts`): permission enforcement (non-manager can't check in others or add a walk-in), check-in/undo via `markAttendance` (no duplicate record), walk-in add/remove, and `canManageAttendance`. ✅ 6/6
- **Mobile** (`features/leader-tools/utils/__tests__/checkIn.test.ts`): live-count cases — Going present, walk-ins counted in N and M, empty roster, full bar. ✅ 7/7

Out of scope (called out in the spec as fast-follows): offline check-in and QR / self check-in.

Closes #651
Dev-dashboard item: `x976b03n6f0cn8w5btj0hhnbmn8bf7mm`

---
_Generated by [Claude Code](https://claude.ai/code/session_015sRmm9QVCJRh68CrfdThA6)_


apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx
```diff
@@ -0,0 +1,841 @@
...
+    api.functions.meetings.attendance.canManageAttendance,
+    meetingId && token
+      ? { meetingId: meetingId as Id<"meetings">, token }
+      : "skip"
+  );
+  const isLoadingPermission = canManage === undefined && !!meetingId && !!token;
+
+  // Going roster.
+  const rsvpData = useQuery(
+    api.functions.meetingRsvps.list,
```
Comment (lilseyi) [inline apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx]: **Blocking — the Going roster (and the "N / M" count) is silently truncated to 10 for most managers.**

`meetingRsvps.list` returns a **limited preview**, not the full roster, whenever the calling user has not themselves RSVPed to this meeting. From `meetingRsvps.ts`:

```ts
// If user is NOT authenticated OR has NOT RSVPed, return limited access response
if (!userId || !userRsvp) {
  ...
  users: optionRsvps.slice(0, 10)  // First 10 users per option for preview
```

The check-in screen is opened by an event **manager** (host / group leader / community admin). A manager very often has **not** RSVPed "Going" to the event they're running — a community admin checking in another group's event essentially never has. In that case `userRsvp` is null and this query returns at most **10** Going users (drawn from only the first 50 RSVPs).

Consequences, both of which break "Done when" items in the spec:
- **Person #11+ can never be checked in** — they don't appear in the roster at all. For any event with more than 10 "Going", check-in is unusable past the tenth person. (Spec: "The screen lists everyone who RSVPed **Going**.")
- **The live count is wrong** — `computeCheckInSummary` derives `total` from the truncated `goingUsers`, so "N / M checked in" under-reports M (and the progress bar is wrong). (Spec: "A live **N / M checked in** summary reflects the current count.")

This limitation is inherited from reusing `meetingRsvps.list` (the existing view-only Guest List has the same cap), but for a *view-only preview* a 10-person cap is cosmetic — for an *interactive door list you must act on every person*, it's a functional blocker.

Suggested fix: give the check-in screen a full, manager-gated roster source — e.g. a `canEditMeeting`-gated `listGoingRoster(meetingId, token)` query that returns all Going users, or extend `meetingRsvps.list` to return the full list when the caller passes `canEditMeeting`. Please also add a test with >10 Going RSVPs where the manager has not RSVPed, asserting the full roster and correct M.

---

apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx
```diff
@@ -0,0 +1,841 @@
...
+  // Who is currently marked Present.
+  const attendance = useQuery(
+    api.functions.meetings.attendance.listAttendance,
+    meetingId ? { meetingId: meetingId as Id<"meetings"> } : "skip"
+  );
+  const isLoadingAttendance = attendance === undefined && !!meetingId;
+
+  // Walk-ins (guest records without an account).
+  const guests = useQuery(
+    api.functions.meetings.attendance.listGuests,
```
Comment (lilseyi) [inline apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx:111]: **Non-blocking (privacy note, pre-existing).** Both `listGuests` and `listAttendance` (in `meetings/attendance.ts`) take **no token and enforce no permission** — any authenticated client can read a meeting's full attendance and guest records for any `meetingId`. The mutations are correctly gated (`markAttendance`/`addGuest` now check `canEditMeeting` — nice catch adding that to `addGuest`), and the *screen* is gated by `canManageAttendance`, but the underlying read queries are not.

This predates the PR so it isn't a merge blocker for this change, and the spec did say to reuse these. Flagging because this feature now actively encourages capturing **walk-in phone numbers**, which then live in the unauthenticated `listGuests` response — so the exposure matters more than it did before. Worth a follow-up to gate these two read queries behind `canManageAttendance` (or at least stop returning phone numbers to non-managers).

---

lilseyi (commented):
## Review: Add event check-in (changes requested)

Reviewed against the approved spec across correctness, security, spec-fidelity/UX, and tests. The shape of the implementation is good — reuse-first, `addGuest` correctly gained the `canEditMeeting` gate it was missing, the pure count helpers are well-tested, the entry point and route identifier match the existing Guest List, and the restricted-access + empty-state handling is thoughtful.

**One blocking issue** (inline): the Going roster is sourced from `meetingRsvps.list`, which returns only a **10-person preview** to any caller who hasn't personally RSVPed to the event — which is the normal state for a host/leader/admin running check-in. So for events with more than 10 "Going", attendees past the tenth never appear and can't be checked in, and the live "N / M" count under-reports M. That defeats two of the spec's "Done when" items (lists everyone Going; live count is correct). Needs a manager-gated full-roster source + a >10-RSVP test.

**One non-blocking note** (inline): `listGuests` / `listAttendance` are unauthenticated reads; the new walk-in phone capture now flows into that unguarded response. Pre-existing, reuse was per-spec, so not a blocker — flagged for a follow-up.

Everything else checks out. Once the roster truncation is addressed, this looks close.

---

apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx
Comment (lilseyi) [reply apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx]: Fixed in a1a3067. Added a manager-gated `meetingRsvps.goingRoster` query (gated by `canEditMeeting` — the same rule the attendance mutations enforce) that returns the **complete** "Going" roster, and switched the check-in screen to it instead of `meetingRsvps.list`. The shared `list` keeps its 10-per-option preview cap for the view-only guest list.

Added the requested test: `event-checkin.test.ts` now seeds 15 "Going" RSVPs with a manager who has **not** RSVPed, asserts `goingRoster` returns all 15 (and that `list` still truncates to 10 for that caller), plus a non-manager is rejected.

---

apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx
```diff
@@ -0,0 +1,841 @@
...
+      : "skip"
+  );
+  const isLoadingPermission = canManage === undefined && !!meetingId && !!token;
+
+  // Going roster.
+  const rsvpData = useQuery(
+    api.functions.meetingRsvps.list,
+    meetingId
+      ? { meetingId: meetingId as Id<"meetings">, token: token ?? undefined }
+      : "skip"
```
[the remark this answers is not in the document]
Comment (lilseyi) [reply apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx]: Fixed in a1a3067. The check-in screen now reads from a new manager-authorized `meetingRsvps.goingRoster` query (gated by `canEditMeeting`) that returns the full Going roster rather than the 10-per-option preview from `meetingRsvps.list`. A test with >10 Going RSVPs and a manager who hasn't RSVPed asserts the complete roster and correct count.

---

apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx
Comment (lilseyi) [reply apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx:111]: Agreed this is worth doing, and thanks for the nuance that walk-in phone numbers raise the stakes. Leaving it as a scoped follow-up rather than folding it into this PR, for two reasons:

1. **It's pre-existing and shared.** `listAttendance` / `listGuests` are unauthenticated today and are also consumed by the existing leader attendance report (`useAttendanceReport` → `AttendanceScreen` / `AttendanceEditScreen` / `AttendanceDetails`). Gating them means adding a required `token` + `canEditMeeting` check and threading auth through those call sites too — a change to shared query contracts that belongs in its own PR with its own review/verification, not bundled into the check-in feature (the spec explicitly scoped this work to *reuse* these queries).
2. **No new exposure surface in this PR's own UI.** The check-in screen itself is already gated by `canManageAttendance`, so within this feature the reads only render for managers.

Filed as a follow-up to gate both reads behind `canManageAttendance` (or at minimum stop returning `phoneNumber` to non-managers). Happy to pick it up next if you'd like it prioritized.

---

apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx
```diff
@@ -0,0 +1,841 @@
...
+    meetingId ? { meetingId: meetingId as Id<"meetings"> } : "skip"
+  );
+  const isLoadingAttendance = attendance === undefined && !!meetingId;
+
+  // Walk-ins (guest records without an account).
+  const guests = useQuery(
+    api.functions.meetings.attendance.listGuests,
+    meetingId ? { meetingId: meetingId as Id<"meetings"> } : "skip"
+  );
+  const isLoadingGuests = guests === undefined && !!meetingId;
```
Comment (lilseyi) [inline apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx]: **Blocking — the check-in screen's data is not gated server-side (spec acceptance criterion: "Non-managers cannot see … its data … permission enforced on the server, not just hidden in the UI").**

This screen reads its attendance state from `listAttendance` and its walk-in list from `listGuests` — and both of those queries in `apps/convex/functions/meetings/attendance.ts` take **only `{ meetingId }`**, with no `token` and no `canEditMeeting` check:

- `listAttendance` returns each record joined to the **full `Doc<"users">`** (including `email` and `phone`).
- `listGuests` returns the raw `meetingGuests` rows (walk-in `phoneNumber`, `notes`).

So a non-manager — or an entirely unauthenticated caller — can pass any `meetingId` and read who is checked in plus attendee/walk-in PII. The PR correctly gated the *new* read (`goingRoster`) and the mutations (`markAttendance`, `addGuest`), which makes the gap here inconsistent: the roster is manager-only but the attendance/walk-in data the same screen displays is world-readable.

I know these two queries pre-date this PR (they're also used by `useAttendanceReport`), so the fix touches shared code — but this feature's own acceptance criterion explicitly requires the check-in data to be server-gated, and it isn't. Please mirror `goingRoster`: add a `token` arg + `canEditMeeting` gate to `listAttendance`/`listGuests` (threading the token through the existing `useAttendanceReport` caller, which already runs in a manager context), or route this screen through gated equivalents.

---

apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx
```diff
@@ -0,0 +1,841 @@
...
+  // which caps non-RSVPed callers at a 10-person preview) so every "Going"
+  // attendee can be checked in and the N/M count is complete — a manager
+  // running an event often hasn't RSVPed to it themselves.
+  const goingRoster = useQuery(
+    api.functions.meetingRsvps.goingRoster,
+    meetingId && token
+      ? { meetingId: meetingId as Id<"meetings">, token }
+      : "skip"
+  );
+  const isLoadingRsvp = goingRoster === undefined && !!meetingId && !!token;
```
Comment (lilseyi) [inline apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx]: **Non-blocking (robustness) — `goingRoster` is issued before permission is confirmed, which defeats the Restricted-Access modal and can crash the screen for a non-manager.**

This query fires as soon as `meetingId && token` are present — it is **not** gated on `canManage`. Server-side, `goingRoster` *throws* for anyone who fails `canEditMeeting`, and the re-exported `useQuery` here is the stock `convex/react` hook, which surfaces a query error by **throwing during render**. So when a non-manager actually reaches this route (a direct Expo-Router deep link — which is the reason the "Restricted Access" modal exists — or a client/server permission mismatch, e.g. the entry point's client-side `isCommunityAdmin` flag vs. the server's community-scoped check), the component throws instead of showing the intended modal. The `canManage === false` → `showRestrictedModal` path becomes effectively dead.

Security is intact (the server denies), but the graceful-degradation UI the screen was built around never runs. Suggest gating this query on `canManage === true` (pass `"skip"` until permission resolves), or having `goingRoster` return `[]` for non-managers instead of throwing.

---

apps/convex/__tests__/event-checkin.test.ts
```diff
@@ -0,0 +1,388 @@
...
+    );
+
+    let guests = await t.query(
+      api.functions.meetings.attendance.listGuests,
+      { meetingId }
+    );
+    expect(guests).toHaveLength(1);
+    expect(guests[0].firstName).toBe("Walk");
+
+    await t.mutation(api.functions.meetings.attendance.removeGuest, {
```
lilseyi, line 411: **Non-blocking (test coverage) — the `removeGuest` permission path is never exercised.** `removeGuest` here is only ever called with `leaderToken`. The spec requires server-side permission enforcement for the walk-in actions, and `addGuest` has a symmetric non-manager rejection test (good) — but there's no test that a non-manager is rejected by `removeGuest`. The backend does gate it (`canEditMeeting`), so this is a coverage gap rather than a bug; adding a `.rejects.toThrow()` case for a non-manager calling `removeGuest` would fully pin the "permission enforced server-side" guarantee.

---

lilseyi (commented):
Automated review (dev-assistant, high-risk item) — reviewed the diff against the approved spec with parallel correctness / security / spec-fidelity / test-adequacy passes, then adversarially verified each finding.

**Verdict: changes requested.** The feature is well-built and faithful to the spec in most respects — the roster query (`goingRoster`) and the mutations (`markAttendance`, `addGuest`) are correctly gated with `canEditMeeting`, the count math and toggle/undo logic are sound, tests pass (convex 8/8, mobile 7/7), no web-only UI libs or `runtimeVersion` bump. One finding blocks:

- **Blocking:** the screen's attendance state and walk-in list come from `listAttendance` / `listGuests`, which are **ungated** (no token, no `canEditMeeting`) and expose attendee/walk-in PII to any caller with a `meetingId` — so the spec's "non-managers cannot see the check-in data, enforced server-side" criterion isn't fully met even though the roster and mutations are gated. Details inline.
- **Non-blocking:** `goingRoster` fires before permission resolves, so a non-manager reaching the route throws/crashes instead of seeing the built-in Restricted-Access modal; and the `removeGuest` permission path is untested. Both inline.

(Routine sessions can't submit a formal approve/request-changes review, so this is posted as review comments; the changes-requested verdict is recorded on the dashboard.)

---

apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx
Comment (lilseyi) [reply apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx]: Fixed in e9422f1. `listAttendance` and `listGuests` now require a `token` and gate on `canEditMeeting` — mirroring `goingRoster` and the attendance mutations — so a non-manager (or unauthenticated caller) can no longer read attendee PII or walk-in phone numbers by `meetingId`. Threaded the token through the two production callers: the check-in screen (passes `token`) and `useAttendanceReport` (fetches it via `useAuth`; it only runs in leader/manager contexts). Added non-manager `.rejects.toThrow()` tests for both reads in `event-checkin.test.ts`, and updated the existing callers/tests to pass a token.

---

apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx
Comment (lilseyi) [reply apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx]: Fixed in e9422f1. The three data queries (`goingRoster`, `listAttendance`, `listGuests`) now skip until `canManage === true` instead of firing as soon as `meetingId && token` are present. A non-manager reaching the screen (deep link or client/server permission mismatch) no longer triggers the throwing query during render — `canManage` resolves to `false`, the queries stay skipped, and the `showRestrictedModal` path runs as intended. The loading flags were updated to key off the same `canFetch` gate so a non-manager lands on the Restricted Access modal rather than an infinite spinner. Server remains the source of truth.

---

apps/convex/__tests__/event-checkin.test.ts
lilseyi: Fixed in e9422f1. Added a "a non-manager cannot remove a walk-in" test to `event-checkin.test.ts`: a manager adds a walk-in, then a plain member's `removeGuest` call is asserted to `.rejects.toThrow()`. This pins the server-side permission guarantee symmetrically with the existing `addGuest` rejection test.

---

apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx
Comment (lilseyi) [reply apps/mobile/app/(user)/leader-tools/[group_id]/events/[event_id]/checkin.tsx:111]: Update: rather than defer this, I folded the read-gating into this PR (e9422f1), since the follow-up review escalated it to a blocker. `listAttendance` and `listGuests` now take a `token` and enforce `canEditMeeting`, so walk-in phone numbers and attendee PII are no longer world-readable. The token is threaded through the shared `useAttendanceReport` caller (it runs only in manager contexts), so the existing leader attendance report keeps working.

---

apps/convex/functions/meetings/attendance.ts
```diff
@@ -157,6 +194,21 @@ export const addGuest = mutation({
...
+    // Adding a guest is a host action (ADR-022): event creator, group
+    // leaders, and community admins can add to the list. Mirrors the check
+    // already guarding removeGuest / updateGuest.
+    if (!(await canEditMeeting(ctx, recordedById, meeting))) {
+      throw new Error(
+        "Only the event creator, group leaders, or community admins can add guests"
+      );
+    }
+
     return await ctx.db.insert("meetingGuests", {
```
lilseyi, line 212: Non-blocking (minor): the spec makes first name required for a walk-in, but `addGuest`'s validator keeps `firstName: v.optional(v.string())`, so the "required" rule is enforced only in the UI (`AddWalkInModal`'s `canSubmit`). A direct client call could insert a nameless guest, which would render as a blank row in the check-in list. The user-facing flow is correct today, so this is a server-side robustness hardening for a follow-up, not a merge blocker.

---

apps/convex/__tests__/event-checkin.test.ts
```diff
@@ -0,0 +1,433 @@
...
+      api.functions.meetings.attendance.listAttendance,
+      { token: leaderToken, meetingId }
+    );
+    expect(attendance).toHaveLength(1); // still one record, flipped
+    expect(attendance.filter((a) => a.status === 1)).toHaveLength(0);
+  });
+});
+
+describe("event check-in — Going roster", () => {
+  test("goingRoster returns EVERY Going attendee to a manager who has not RSVPed (>10)", async () => {
```
lilseyi, line 310: Non-blocking (minor test gap): `goingRoster` correctly filters to `GOING_RSVP_OPTION_ID`, excluding Maybe/Can't Go (spec item: "Maybe/Can't Go RSVPers are not pre-listed"). But this fixture only seeds Going RSVPers, so if the `.filter(r => r.rsvpOptionId === GOING_RSVP_OPTION_ID)` were ever removed, every test here would still pass. Consider seeding one Maybe and one Can't Go RSVPer and asserting they do NOT appear in the roster, so the exclusion is verified rather than true only by construction.

---

lilseyi (commented):
## Automated review — Event Check-In (verdict: approved)

Reviewed the diff against the approved spec with four parallel lenses (correctness, security, spec-fidelity/UX, tests) at high scrutiny (high-risk change), adversarially verifying each candidate finding. **No blocking findings survived** — the feature implements the spec faithfully and the security-sensitive parts hold up.

**What's solid**
- **Server-side permission enforcement** — `goingRoster`, `listAttendance`, `listGuests`, `markAttendance`, `addGuest`, `removeGuest` all gate on `canEditMeeting` (host / group leaders / community admin). Newly gating `listAttendance`/`listGuests` also closes a pre-existing PII leak (attendee name/email/phone, walk-in phone) — good catch to fix in the same change.
- **No caller regressions** from the `listAttendance`/`listGuests` signature change: the only production caller (`useAttendanceReport`) and all tests were updated to pass `token`.
- **Undo semantics** are correct: setting status 0 (vs deleting) is behaviorally identical to "no record" for every stats consumer (all read `status === 1`, denominators are meeting counts), so attendance/health numbers stay accurate.
- **Roster is Going-only**, walk-in count math is correct (no double-count), and the event_id encoding round-trips identically to the sibling Guest List screen.
- **Tests pass** (11/11 convex, 7/7 util) and cover permission enforcement, check-in/undo, walk-in add/remove, and the live count.

**Non-blocking items** (2 inline comments): server-side `firstName` is UI-enforced only; the Going-only exclusion is asserted by fixture construction rather than a negative test.

**Pre-existing, out of scope (heads-up only):** `markAttendance` skips the permission check when a caller marks *their own* attendance (`args.userId === recordedById`) and, unlike `selfReportAttendance`, does not verify group/community membership — so a non-member could self-insert a `Present` record for any meeting. This predates this PR and the check-in screen doesn't expose it (score recompute is skipped for non-members). Worth a follow-up to align the self path with `selfReportAttendance`'s membership check.

---

PR #662, opened 2026-07-30
Show the full message in previews & notifications

lilseyi:
Closes #661

## What was wrong

When someone messaged you, every preview surface chopped the message off — often mid-word. The message body was sliced to **100 characters with no "…"**, and that single sliced value fed the notification **emails**, the **push** body, and (separately) the **inbox rows**. The reported chat-request email ended on *"…Please keep me in pr"* — the word "prayers" cut in half, with nothing to signal it was shortened. It read like the email was broken.

## What changed

The one 100-char cut is now three surface-appropriate caps, all **emoji-safe** (they cut on grapheme boundaries via `Intl.Segmenter`, falling back to code points, so a cut can never leave a broken glyph like half a 🙏):

- **Emails** now show the **whole message**. The only limit is a generous ~1000-char safety cap (so a pasted essay can't create a runaway email), and only then does it end in "…". The quote box already escapes the body with `escapeHtml()`, and the full body stays escaped.
- **Push** re-truncates to ~200 chars and **always ends in "…"** when shortened, so the cut reads as an intentional preview rather than a mid-word glitch. iOS/Android still clip the lock-screen banner further — tapping opens the full thread.
- **Inbox / group-list rows** now show **up to 2 lines** of the message-body preview (was 1), and the stored `lastMessagePreview` grew from 100 to ~200 chars so a 2-line row has content to show. Rows stay a normal height.
- **Non-text messages** (photo, files, shared event, prayer, availability request) keep their fixed labels — "Sent a photo", "Shared an event", etc. — unchanged on every surface.

The same source value feeds the group-message, 1:1 DM-request, leader-DM, and @mention paths, so all of them read fully.

## Before / after

Rendered mock built from the real email `messageBox` and inbox-row styles — **not a device capture** (this run has no iOS simulator).

[before/after](https://images.togather.nyc/dev-assistant/a92985af-dfd5-465c-8d9b-17a91711e7aa-before-after.png)

## Where the change lives

- `apps/convex/lib/text.ts` — **new** `truncatePreview()` + the per-surface caps (`EMAIL_PREVIEW_MAX` 1000, `PUSH_PREVIEW_MAX` / `LIST_PREVIEW_MAX` 200).
- `apps/convex/functions/messaging/events.ts` — email gets the full body (email-safety-capped); push re-truncates to the push cap.
- `apps/convex/lib/notifications/definitions.ts` — group/mention push body re-truncates; the mention email keeps the full body.
- `apps/convex/functions/messaging/messages.ts` (`generateMessagePreview`), plus `meetings/index.ts` and `scheduledJobs.ts` — stored inbox preview raised to ~200, emoji-safe.
- `apps/mobile/features/chat/components/ChatInboxScreen.tsx` & `GroupedInboxItem.tsx` — message-body preview lines `numberOfLines={1}` → `{2}` (name/title clamps untouched).

## Testing

- `apps/convex` unit tests: `truncatePreview` (emoji boundary, ellipsis counting, per-surface caps) and `generateMessagePreview` (raised cap, canned labels preserved). Existing "truncate long message preview" test updated for the new list cap.
- Convex `tsc --noEmit` clean; the full mobile Jest suite (198 suites / 1880 tests) passes via the pre-push hook.

## Notes for the reviewer

- **No backfill:** existing conversations keep their stored 100-char preview until their next message lengthens it — an accepted trade-off, no migration needed.
- The two legacy `bottomRow` inbox layouts render the timestamp inline beside the preview; with a 2-line preview the timestamp centers vertically. Left the styles untouched per the spec (change was scoped to the `numberOfLines` clamp).

---
_Generated by [Claude Code](https://claude.ai/code/session_01JJrL6jTeFJJU6FsRQej4xS)_


apps/convex/lib/text.ts
```diff
@@ -0,0 +1,64 @@
...
+ * own `numberOfLines` renders the visual ellipsis).
+ *
+ * Returns the original string unchanged when it's already within `max`.
+ */
+export function truncatePreview(
+  text: string,
+  max: number,
+  ellipsis = true,
+): string {
+  const graphemes = toGraphemes(text);
```
lilseyi: **Non-blocking (efficiency).** `toGraphemes(text)` segments and materializes an array over the **entire** message body before the length check, regardless of how small `max` is. This now runs on the send hot path — once in `generateMessagePreview` (via `sendMessage`) and again in `onMessageSent` — and message `content` for accepted channels has no server-side length cap (only the ~1 MiB Convex document ceiling). So a very large message runs `Intl.Segmenter` over the whole thing twice per send where the old code did an O(1) `slice(0, 100)`.

Cheap guard: slice `text` to a safe upper bound before segmenting, e.g.
```ts
const graphemes = toGraphemes(text.length > max * 8 ? text.slice(0, max * 8) : text);
```
A grapheme is ≥1 UTF-16 code unit, so `max * 8` code units is always ≥ `max` graphemes with generous headroom for ZWJ sequences — the visible result is identical, but you never segment megabytes to keep 200 graphemes.

---

lilseyi (commented):
## Automated review — Show the full message in previews & notifications

Reviewed against the approved spec with four parallel lenses (correctness, security, spec-fidelity/UX, tests), each finding adversarially verified. **Risk level: high**, so this was a line-by-line read.

**What's solid:**
- **HTML-injection safe.** The now-longer (up-to-1000-char) body stays escaped at every render site: chat-request (`emailTemplates.ts:199`), leader-DM (`:242`), and mention (`definitions.ts:416`) all wrap the exact `messagePreview` field that now carries the fuller body in `escapeHtml()`. No path reaches email HTML unescaped.
- **Emails / push / inbox all satisfy their primary "Done when."** Emails get the full body capped at 1000+"…" only if the cap is hit; both push paths re-truncate to 200 and always end in "…"; primary inbox rows now render 2 lines. `truncatePreview`'s ellipsis math is correct (result stays within `max`) and the grapheme-boundary cut is genuinely emoji-safe.
- **Tests pass** (52 tests across the 3 touched files, vitest) and cover the emoji-boundary + ellipsis-counting edge cases well.
- **No authorization leak** (every notification recipient is already an authorized reader) and **no ReDoS**.

**Findings to address (see also the inline comments):**

1. **Missed 2-line preview line — `GroupedInboxItem.tsx:1106`.** The legacy (non-WA-shell) `subChannelPreview` renders a real message body (`getMessagePreview(channel)`) but is still `numberOfLines={1}`, while its WA-shell twin at line 876 uses `{2}` and this PR raised the sibling parent rows (743, 1046). The spec's "Where the change lives" names `GroupedInboxItem.tsx` message-body preview lines specifically. Change this to `numberOfLines={2}` to match. *(Couldn't anchor an inline comment here — the line is outside the diff hunks.)*

2. **Missed emoji-safe migration — `apps/convex/functions/eventBlasts.ts:427`** (`lastMessagePreview: args.message.slice(0, 100)`). This is a stored inbox-preview write that the PR didn't migrate. Two consequences: it still caps at 100 (vs. the new `LIST_PREVIEW_MAX` 200 the 2-line rows now expect) and, being a raw UTF-16 `.slice`, it can split an emoji straddling char 100 into mojibake — the exact failure `truncatePreview` was written to prevent. Swap to `truncatePreview(args.message, LIST_PREVIEW_MAX, false)` for consistency with the other stored-preview sites. *(File not in this PR's diff, so no inline anchor.)*

**Non-blocking notes:**
- **Non-text messages on email/push:** the push/email excerpt is fed raw `message.content`, so an attachment-only message yields an empty quote box / excerpt rather than "Sent a photo". This is **pre-existing** (old code sliced the same empty `content`) and not introduced here — the stored inbox preview still shows the label — but the spec's "labels on all three surfaces" isn't fully met on email/push. Worth a follow-up if in scope.
- **Test coverage gap:** `formatChatPushBody` (`definitions.ts`) — the one place proving the mention *email* keeps the full body while the mention *push* re-truncates — has no direct test. Behavior is correct by inspection; a small test would lock it in.

The two numbered items are small, concrete, and squarely within the spec's stated goals (2-line previews everywhere; emoji-safe cuts on every preview write), so I'm flagging changes before merge.

---

apps/convex/lib/text.ts
```diff
@@ -0,0 +1,69 @@
...
+  max: number,
+  ellipsis = true,
+): string {
+  // A grapheme is ≥1 UTF-16 code unit, so `max * 8` code units always yields
+  // ≥ `max` graphemes (generous headroom for ZWJ sequences). Slicing to that
+  // bound first means we never segment a multi-megabyte body just to keep a
+  // couple hundred graphemes — the visible result is identical.
+  const bound = max * 8;
+  const graphemes = toGraphemes(text.length > bound ? text.slice(0, bound) : text);
+  if (graphemes.length <= max) return text;
```
lilseyi: **Blocking — the `max * 8` bound silently defeats the length cap for graphemes wider than 8 UTF‑16 code units, returning the *entire* untruncated message.**

The comment claims "`max * 8` code units always yields ≥ `max` graphemes," but that only holds when every grapheme is ≤ 8 code units. It isn't for:
- **ZWJ emoji** — 👨‍👩‍👧‍👦 is 11 code units / 1 grapheme.
- **Combining‑mark / "Zalgo" text** — a base char + N combining marks is 1 grapheme of essentially unbounded length.

When the leading `max*8` code units segment to ≤ `max` graphemes, `graphemes.length <= max` is true and line 66 does `return text` — the full, unsliced original.

Reproduced against this exact helper:
```
truncatePreview("👨‍👩‍👧‍👦".repeat(300), 200)
// → 300 graphemes / 3300 code units, no "…": the whole body comes back
```

This is reachable in production: `sendMessage` enforces a 1000‑char cap **only** for still‑pending DM requests — accepted DMs and group channels have no send‑time length limit, so a member can post a multi‑KB emoji/Zalgo message. Consequences on all three surfaces at once:
- **Email safety cap defeated** — the runaway email the ~1000 cap exists to prevent (spec: *"a pasted essay can't create a runaway email"*).
- **Push** — `formatChatPushBody` / `sendAdHocMessageNotifications` re‑truncate to 200 but get the full body back; the serialized Expo payload can blow past the ~4 KB APNs/FCM limit → the push is rejected and **dropped for every recipient of that message** (send failures are caught, so no crash — just silent non‑delivery).
- **Stored `lastMessagePreview`** exceeds `LIST_PREVIEW_MAX`.

This also means the spec's "very long messages degrade gracefully" / emoji‑boundary done‑when items aren't met for this input class.

**Fix:** don't early‑return the original. Segment the full string (or, if keeping the pre‑slice for perf, when `text.length > bound` and the sliced window is within `max`, fall through to slicing from the grapheme array rather than returning `text`). A simple correct form:
```ts
const graphemes = toGraphemes(text);
if (graphemes.length <= max) return text;
const keep = ellipsis ? Math.max(0, max - 1) : max;
return graphemes.slice(0, keep).join("") + (ellipsis ? "…" : "");
```
If the multi‑megabyte‑body perf concern is real, cap by a much larger byte bound *and* still slice from `graphemes` on the truncation branch.

---

apps/convex/__tests__/truncatePreview.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    const text = "Please keep me in prayers🙏";
+    const graphemeLen = [...text].length; // code-point count
+    // Cut exactly where only half the emoji would fit if we sliced by code unit.
+    const result = truncatePreview(text, graphemeLen, false);
+    expect(result).toBe(text); // fits exactly, unchanged
+    // One shorter: the emoji is dropped entirely, not split.
+    const shorter = truncatePreview(text, graphemeLen - 1, false);
+    expect(shorter).toBe("Please keep me in prayers");
+  });
+
```
lilseyi, line 71: **Test gap that hides the bug above.** Both emoji tests use only 🙏 (a single astral code point = 2 UTF‑16 units, avg < 8), so the `max*8` pre‑slice path is never stressed and the suite asserts emoji‑cut safety that doesn't generalize. A case with a wide grapheme would fail today and pin the fix, e.g.:
```ts
test("caps content made of wide (ZWJ / combining) graphemes", () => {
  const body = "👨‍👩‍👧‍👦".repeat(300); // 11 code units each
  const out = truncatePreview(body, PUSH_PREVIEW_MAX);
  expect([...out].length).toBeLessThanOrEqual(PUSH_PREVIEW_MAX);
  expect(out.endsWith("…")).toBe(true);
});
```
Worth adding a combining‑mark ("Zalgo") variant too, since that class triggers the same path.

---

lilseyi (commented):
## Automated review — changes requested

Reviewed against the approved spec with parallel subagents (correctness, security, spec‑fidelity/UX, tests); every finding was adversarially re‑checked before posting.

**What's solid.** Spec fidelity is clean: all three emails (chat‑request, leader‑DM, mention) now receive the full body via `truncatePreview(content, EMAIL_PREVIEW_MAX)`, escaped with `escapeHtml()` on every template — no XSS/injection regression. Push re‑truncates to 200 with a trailing "…" on all five push paths. The mobile change is exactly right — all message‑body preview lines flipped to `numberOfLines={2}` in both the WhatsApp‑shell and legacy row variants, with no name/title/label line touched. Non‑text canned labels ("Sent a photo", "Shared an event", …) are preserved. The 52 convex unit tests pass locally.

**Blocking finding (1).** The shared `truncatePreview` helper defeats its own length cap for graphemes wider than 8 UTF‑16 code units (ZWJ emoji, combining‑mark/"Zalgo" text) and returns the *entire* untruncated message — defeating the ~1000‑char email safety cap the spec explicitly requires and risking dropped push notifications (oversized Expo payload). It's reachable because accepted/group channels have no send‑time length limit. Details + a fix on `apps/convex/lib/text.ts`. The tied test gap (emoji tests only exercise 2‑code‑unit 🙏) is noted on `truncatePreview.test.ts`.

**Non‑blocking observations (no change required to approve):**
- **Other `lastMessagePreview` writers still use `slice(0, 100)`** and were not unified: `functions/messaging/polls.ts` (~315/633/748), `functions/messaging/availabilityRequests.ts:167`, `functions/cli/messaging.ts:291`. Those inbox rows will render a shorter preview (and can split an emoji at the 100‑code‑unit boundary) than the new 2‑line rows expect. Not a regression, but the same inconsistency the PR set out to remove.
- **No surface‑level test** asserts the email path shows the full body up to 1000 with "…" only at the cap, nor that the (now longer) body stays HTML‑escaped. Escaping is correct by inspection; a regression test would be cheap insurance.

Fixing finding (1) — a small, local change to `truncatePreview` plus the added test — clears the block.

---

apps/convex/lib/text.ts
lilseyi: Fixed in `bddd367`. You're right — `max * 8` only bounds the grapheme count when every grapheme is ≤ 8 UTF‑16 code units, which ZWJ emoji (11) and stacked combining marks violate, so the bounded prefix segmented to ≤ `max` and the old code early‑returned the full body.

The fix keeps the fast path but no longer trusts it as a terminal answer: when the bounded prefix segments to ≤ `max` graphemes **and** we actually pre‑sliced (`text.length > bound`), it now falls back to segmenting the full body to get the true grapheme count before deciding. Wide‑grapheme messages are truncated correctly; normal messages still never segment more than the bounded prefix. Verified with `truncatePreview("👨‍👩‍👧‍👦".repeat(300), 200)` → 199 graphemes + "…".

---

apps/convex/__tests__/truncatePreview.test.ts
lilseyi: Added in `bddd367`. There are now two wide-grapheme tests that exercise the pre-slice escape hatch and fail against the old code:

- **ZWJ emoji** — `"👨‍👩‍👧‍👦".repeat(300)` (11 code units / grapheme), asserting the *grapheme* count (`Intl.Segmenter`, not `[...out]`, which counts code points) is ≤ `PUSH_PREVIEW_MAX` and the result ends in "…".
- **Combining-mark ("Zalgo")** — `("a" + "́".repeat(20)).repeat(300)`, asserting the body is truncated and no lone surrogate survives the cut.

---

apps/convex/schema.ts
```diff
@@ -1715,7 +1715,7 @@ export default defineSchema({
     disabledByUserId: v.optional(v.id("users")),
     // Denormalized for performance
     lastMessageAt: v.optional(v.number()), // Unix timestamp ms
-    lastMessagePreview: v.optional(v.string()), // First 100 chars
+    lastMessagePreview: v.optional(v.string()), // First ~200 chars (see LIST_PREVIEW_MAX)
```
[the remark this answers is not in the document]
lilseyi: Already addressed — `recordBlast` was updated in `e296263` (before this fix round) and now writes `lastMessagePreview: truncatePreview(args.message, LIST_PREVIEW_MAX, false)` at `apps/convex/functions/eventBlasts.ts:428`, so event-chat mirror rows share the same emoji-safe ~200-grapheme cap as every other inbox writer.

---

PR #689, opened 2026-07-31
Allocate payouts on Stripe NET, and never strand a failed item

lilseyi:
## Why

Money could never reach a group fund.

`planAllocations` matched **gross** donation totals (`amountCents + feeCoverCents`) against a Stripe payout, but a payout is **net** of processing fees. Gross always exceeds net, so in the common one-donation-per-payout case the first donation never fit its own payout — and the loop `break`ed rather than `continue`d, so the queue never advanced again. Group Increase Accounts stayed at zero while `funds.balanceCents` reported the money was there.

This was invisible to every alarm we have: the nightly invariant computed pending allocations on the same gross basis, so gross-vs-gross stayed perfectly self-consistent. `ARCHITECTURE.md` described it as "a tail of donations pending". It is not a tail; it is a deterministic total stall. ADR-032 named net matching a hard go-live prerequisite (Phase 2 item 6).

Two further defects in the same job:

- `runAllocation` claimed the payout in `processedStripePayouts` **before** transferring, then looped `createAccountTransfer` with no per-item error handling. One transient Increase failure stranded every remaining donation permanently — the redelivered webhook was ignored as "already processed", and `retryStaleAllocations` is alert-only.
- Nothing ever wrote the `"fee"` ledger kind, so processing fees were absent from the books entirely.

## What changed

**Net matching.** New `listPayoutChargeNets` (`lib/finance/stripeConnect.ts`) reads `GET /v1/balance_transactions?payout=…` on the connected account and returns each charge's `net` in integer cents. Stripe tells us exactly which donations a payout carried, so membership is read rather than inferred from a running total.

**Never stall.** The residual-budget check `continue`s. Ordering stays deterministic.

**Exactly-once under partial failure.** Replay protection moved from the payout to the **(payout, donation)** pair: `planAllocations` stamps `allocationPayoutId` + `payoutNetCents` in the same transaction that selects a donation. A redelivered `payout.paid` re-derives the identical set, drops what is already allocated, and finishes only the tail. `claimPayout` is gone. Each transfer is individually wrapped; a failure audits `allocation.transfer_failed`, leaves the donation pending-with-stamp, and the loop continues. Two independent locks hold the guarantee: `recordAllocation` no-ops an already-allocated donation, and the `alloc:{donationId}` Increase idempotency key collapses the transfer-landed-then-crashed case. Stripe is queried before any write, so an unreachable Stripe leaves the payout untouched.

**An honest invariant.** `computePendingAllocationCents` sums through `allocationAmountCents` — gross while a donation is in the pipeline, net once a payout binds it. On its own that would make every healthy fund drift by its fees forever, so `recordAllocation` posts the realised `gross − net` as a `fee` ledger debit when the transfer lands:

| state | ledger | bank | pending | result |
|---|---|---|---|---|
| in pipeline | gross | 0 | gross | ok |
| bound, transfer stalled | gross | 0 | net | **drift = fee** |
| allocated | net | net | 0 | ok |

**Reviewer note — this changes `funds.balanceCents` semantics** from gross to net after allocation. Net is what the bank actually holds, so the mirror is now truthful, but it is a behavioural change worth a second opinion. It also means a `Fee` row appears alongside each `Donation` row in the member-visible activity feed.

## Tests

`finance-allocation.test.ts` (16) drives the `runAllocation` action itself with the provider boundary mocked and **no suite-wide fake timers**, so an awaited action cannot silently no-op. Covers the exact stall (gross 10000 vs net 9680 now allocates), skip-don't-stall, per-donation replay protection, a 3-item partial failure whose redelivery finishes only the missing item without re-transferring the first, no-op replay, Stripe-down leaving zero writes, cron resume, and the drift-then-close reconcile sequence. `finance-payout-nets.test.ts` (7) covers net extraction, PaymentIntent mapping, paging, and malformed rows.

Full convex suite: **159 files, 2720 tests passing.**

Schema adds `donations.allocationPayoutId?` and `donations.payoutNetCents?`, both optional and backfill-safe.

### DIFF
--- a/apps/convex/__tests__/finance-allocation.test.ts
+++ b/apps/convex/__tests__/finance-allocation.test.ts
@@ -44,7 +44,7 @@ const stubs = vi.hoisted(() => ({
     | {
         charges: unknown[];
         reversedPaymentIntentIds?: string[];
-        unattributedReversalCents?: number;
+        unattributedNetCents?: number;
       }
     | { error: string }
   >(),
@@ -71,6 +71,10 @@ vi.mock("../lib/finance/stripeConnect", async (importOriginal) => {
   const actual = await importOriginal<typeof import("../lib/finance/stripeConnect")>();
   return {
     ...actual,
+    // The real one takes `payoutCents` too and refuses a composition whose
+    // nets don't add up to it — that guard is exercised against raw Stripe
+    // rows in finance-payout-nets.test.ts. Here the composition is the
+    // fixture, so the third argument is irrelevant.
     getPayoutComposition: vi.fn(async (_accountId: string, payoutId: string) => {
       const stub = stubs.payouts.get(payoutId);
       if (!stub) {
@@ -82,7 +86,7 @@ vi.mock("../lib/finance/stripeConnect", async (importOriginal) => {
       return {
         charges: stub.charges as PayoutChargeNet[],
         reversedPaymentIntentIds: stub.reversedPaymentIntentIds ?? [],
-        unattributedReversalCents: stub.unattributedReversalCents ?? 0,
+        unattributedNetCents: stub.unattributedNetCents ?? 0,
       };
     }),
   };
@@ -1285,6 +1289,106 @@ describe("refunds and allocation", () => {
     ).toEqual({ ok: true, driftCents: 0 });
   });
 
+  test("P1-7: a partial refund settling in a LATER payout transfers only what's left of the gift", async () => {
+    const t = convexTest(schema, modules);
+    const { communityId, fundAId } = await seedAllocFixture(t);
+
+    // $100 gift, $30 refunded before the payout settles. Stripe does NOT
+    // guarantee the refund's balance transaction shares the charge's payout,
+    // so this payout still carries the charge at its FULL net (9680) — the
+    // netting in `getPayoutComposition` cannot help here, and the gift stays
+    // legitimately "pending" so the full-refund belt-and-braces doesn't reach
+    // it either. Taking Stripe's net verbatim moved 9680 while the ledger
+    // said 7000, leaving 2680 of returned donor money spendable on the
+    // group's card.
+    const donationId = await seedDonation(t, {
+      fundId: fundAId,
+      paymentIntentId: "pi_late_refund",
+      amountCents: 10_000,
+      createdAt: 1_000,
+      creditLedger: true,
+    });
+    await refundDonation(t, {
+      paymentIntentId: "pi_late_refund",
+      chargeId: "ch_late_refund",
+      amountRefundedCents: 3_000,
+    });
+
+    stubs.payouts.set("po_late_refund", {
+      charges: [{ paymentIntentId: "pi_late_refund", netCents: 9_680 }],
+    });
+
+    await t.action(internal.functions.finance.jobs.runAllocation, {
+      communityId,
+      stripePayoutId: "po_late_refund",
+      payoutCents: 9_680,
+    });
+
+    // What the bank was actually ASKED to move is the assertion that matters.
+    expect(stubs.transferCalls).toEqual([
+      {
+        toAccountId: "account_fund_a",
+        amountCents: 7_000, // 10000 gross − 3000 refunded, NOT Stripe's 9680
+        idempotencyKey: `alloc:${donationId}`,
+      },
+    ]);
+
+    const donation = await getDonation(t, donationId);
+    expect(donation?.allocationStatus).toBe("allocated");
+    expect(donation?.payoutNetCents).toBe(7_000);
+
+    // Ledger and bank agree exactly: credit 10000, refund 3000, no fee left
+    // to realise on the part the fund kept.
+    const entries = await t.run((ctx) =>
+      ctx.db
+        .query("ledgerEntries")
+        .withIndex("by_fund", (q) => q.eq("fundId", fundAId))
+        .collect(),
+    );
+    expect(entries.map((e) => `${e.kind}:${e.direction}:${e.amountCents}`)).toEqual([
+      "donation:credit:10000",
+      "refund:debit:3000",
+    ]);
+    expect((await t.run((ctx) => ctx.db.get(fundAId)))?.balanceCents).toBe(7_000);
+  });
+
+  test("P1-7: the retry cron resumes a bound gift at the refunded-down amount, not its stale stamped net", async () => {
+    const t = convexTest(schema, modules);
+    const { communityId, fundAId } = await seedAllocFixture(t);
+
+    const donationId = await seedDonation(t, {
+      fundId: fundAId,
+      paymentIntentId: "pi_bound_then_partial",
+      amountCents: 10_000,
+      createdAt: 1_000,
+      creditLedger: true,
+    });
+    // Bound at the full net by a pass whose transfer never landed — it
+    // couldn't have known about a refund that hadn't happened yet.
+    await t.run(async (ctx) => {
+      await ctx.db.patch(donationId, {
+        allocationPayoutId: "po_bound_partial_retry",
+        payoutNetCents: 9_680,
+      });
+    });
+    await refundDonation(t, {
+      paymentIntentId: "pi_bound_then_partial",
+      chargeId: "ch_bound_partial_retry",
+      amountRefundedCents: 3_000,
+    });
+
+    await t.action(internal.functions.finance.jobs.retryStaleAllocations, {});
+
+    expect(stubs.transferCalls).toEqual([
+      {
+        toAccountId: "account_fund_a",
+        amountCents: 7_000,
+        idempotencyKey: `alloc:${donationId}`,
+      },
+    ]);
+    expect((await t.run((ctx) => ctx.db.get(fundAId)))?.balanceCents).toBe(7_000);
+  });
+
   test("a gift refunded while bound to a payout is dropped by the retry cron, not transferred", async () => {
     const t = convexTest(schema, modules);
     const { communityId, fundAId } = await seedAllocFixture(t);
@@ -1517,6 +1621,176 @@ describe("unmatched payouts are observable", () => {
     });
 
     expect(await auditActions(t, communityId)).toContain("allocation.payout_unmatched");
+  });
+
+  test("P1-8: replaying a HEALTHY payout raises no unmatched alarm — that's the success condition", async () => {
+    const t = convexTest(schema, modules);
+    const { communityId, fundAId } = await seedAllocFixture(t);
+
+    await seedDonation(t, {
+      fundId: fundAId,
+      paymentIntentId: "pi_healthy",
+      amountCents: 10_000,
+      createdAt: 1_000,
+      creditLedger: true,
+    });
+    stubs.payouts.set("po_healthy", {
+      charges: [{ paymentIntentId: "pi_healthy", netCents: 9_680 }],
+    });
+
+    const first = await t.action(internal.functions.finance.jobs.runAllocation, {
+      communityId,
+      stripePayoutId: "po_healthy",
+      payoutCents: 9_680,
+    });
+    expect(first).toEqual({ allocated: 1, failed: 0 });
+
+    // webhooks.ts routes `payout.reconciliation_completed` to this same
+    // handler, and Stripe emits it after EVERY payout — so this second pass
+    // is the normal course of business, not an edge case. The gift is
+    // already allocated, so the plan is empty; treating an empty plan as
+    // "matched nothing" wrote a full-payout `payout_unmatched` row (with
+    // leftoverCents equal to the entire payout) on every healthy payout,
+    // forever, drowning the one signal that has no automatic recovery.
+    const second = await t.action(internal.functions.finance.jobs.runAllocation, {
+      communityId,
+      stripePayoutId: "po_healthy",
+      payoutCents: 9_680,
+    });
+    expect(second).toEqual({ allocated: 0, failed: 0 });
+
+    expect(await auditActions(t, communityId)).toEqual(["donation.allocated"]);
+  });
+
+  test("P1-8: a payout that finished some items but genuinely lost the rest still alarms", async () => {
+    const t = convexTest(schema, modules);
+    const { communityId, fundAId } = await seedAllocFixture(t);
+
+    await seedDonation(t, {
+      fundId: fundAId,
+      paymentIntentId: "pi_known",
+      amountCents: 1_000,
+      createdAt: 1_000,
+      creditLedger: true,
+    });
+    // 941 of a 50000 payout maps to a gift we know. The alarm must survive
+    // the P1-8 guard: "something was already allocated" is not a licence to
+    // stop looking at the 49059 we cannot account for.
+    stubs.payouts.set("po_partial_known", {
+      charges: [{ paymentIntentId: "pi_known", netCents: 941 }],
+    });
+
+    await t.action(internal.functions.finance.jobs.runAllocation, {
+      communityId,
+      stripePayoutId: "po_partial_known",
+      payoutCents: 50_000,
+    });
+    await t.action(internal.functions.finance.jobs.runAllocation, {
+      communityId,
+      stripePayoutId: "po_partial_known",
+      payoutCents: 50_000,
+    });
+
+    const unmatched = (await auditActions(t, communityId)).filter(
+      (a) => a === "allocation.payout_unmatched",
+    );
+    expect(unmatched).toHaveLength(2);
+  });
+
+  test("P1-6: a payout short by money we can't attribute is REFUSED, not allocated around", async () => {
+    const t = convexTest(schema, modules);
+    const { communityId, fundAId } = await seedAllocFixture(t);
+
+    const donationId = await seedDonation(t, {
+      fundId: fundAId,
+      paymentIntentId: "pi_healthy",
+      amountCents: 50_000,
+      createdAt: 1_000,
+      creditLedger: true,
+    });
+    // A reversal Stripe attributed to this payout that we could not tie to
+    // any PaymentIntent. The charge rows now sum to MORE than the payout
+    // delivered, so funding them all would move money that never arrived —
+    // and the residual-budget check would quietly strand whichever gift
+    // sorted last instead, with leftoverCents at 0.
+    stubs.payouts.set("po_short", {
+      charges: [{ paymentIntentId: "pi_healthy", netCents: 48_500 }],
+      unattributedNetCents: -48_500,
+    });
+
+    const result = await t.action(internal.functions.finance.jobs.runAllocation, {
+      communityId,
+      stripePayoutId: "po_short",
+      payoutCents: 0,
+    });
+
+    expect(result).toEqual({ allocated: 0, failed: 0 });
+    expect(stubs.transferCalls).toEqual([]);
+    const donation = await getDonation(t, donationId);
+    expect(donation?.allocationStatus).toBe("pending");
+    expect(donation?.allocationPayoutId).toBeUndefined();
+
+    const rows = await t.run((ctx) =>
+      ctx.db
+        .query("financeAuditEvents")
+        .withIndex("by_community", (q) => q.eq("communityId", communityId))
+        .collect(),
+    );
+    const unmatched = rows.find((r) => r.action === "allocation.payout_unmatched");
+    expect(JSON.parse(unmatched!.detailsJson!)).toMatchObject({
+      stripePayoutId: "po_short",
+      refused: true,
+      unattributedNetCents: -48_500,
+    });
+  });
+
+  test("P1-6: a gift skipped for overrunning the payout alarms instead of vanishing", async () => {
+    const t = convexTest(schema, modules);
+    const { communityId, fundAId } = await seedAllocFixture(t);
+
+    await seedDonation(t, {
+      fundId: fundAId,
+      paymentIntentId: "pi_a",
+      amountCents: 50_000,
+      createdAt: 1_000,
+      creditLedger: true,
+    });
+    await seedDonation(t, {
+      fundId: fundAId,
+      paymentIntentId: "pi_b",
+      amountCents: 50_000,
+      createdAt: 2_000,
+      creditLedger: true,
+    });
+    // Both gifts are in the payout at 49600, but the payout only carried one
+    // of them. The second is skipped — correctly, it can't be funded — and
+    // leftoverCents then comes out at 0, which used to look perfectly
+    // healthy. A stranded gift is not a healthy pass.
+    stubs.payouts.set("po_overrun", {
+      charges: [
+        { paymentIntentId: "pi_a", netCents: 49_600 },
+        { paymentIntentId: "pi_b", netCents: 49_600 },
+      ],
+    });
+
+    await t.action(internal.functions.finance.jobs.runAllocation, {
+      communityId,
+      stripePayoutId: "po_overrun",
+      payoutCents: 49_600,
+    });
+
+    const rows = await t.run((ctx) =>
+      ctx.db
+        .query("financeAuditEvents")
+        .withIndex("by_community", (q) => q.eq("communityId", communityId))
+        .collect(),
+    );
+    const unmatched = rows.find((r) => r.action === "allocation.payout_unmatched");
+    expect(unmatched).toBeDefined();
+    expect(JSON.parse(unmatched!.detailsJson!)).toMatchObject({
+      leftoverCents: 0, // the number that made this invisible
+      overrunSkippedCount: 1,
+    });
   });
 
   test("a fully matched payout does not alarm", async () => {
--- a/apps/convex/__tests__/finance-allocation.test.ts
+++ b/apps/convex/__tests__/finance-allocation.test.ts
@@ -1,0 +1,898 @@
+/**
+ * Payout allocation: NET matching, never-stall selection, per-item failure
+ * recovery, and honest reconcile (ADR-032 §3 + Phase-2 requirement 6).
+ *
+ * Unlike the other finance suites, this one drives `runAllocation` — the
+ * internalAction itself — end to end, because the bugs it covers live in the
+ * action's control flow (what it claims before it transfers, what it does
+ * when transfer N of M throws, what a redelivered webhook re-does). Both
+ * providers the action lazy-imports are mocked at the module boundary:
+ * `listPayoutChargeNets` (Stripe balance transactions) and
+ * `createAccountTransfer` (Increase).
+ *
+ * Deliberately NO suite-wide `vi.useFakeTimers()` (see finance-cards.test.ts
+ * for why the other suites need it): nothing here goes through
+ * `ctx.scheduler`, so freezing timers would only risk making an awaited
+ * action look like it ran when it didn't.
+ *
+ * Run with: cd apps/convex && pnpm test __tests__/finance-allocation.test.ts
+ */
+
+import { convexTest } from "convex-test";
+import { expect, test, describe, vi, beforeEach } from "vitest";
+import schema from "../schema";
+import { internal } from "../_generated/api";
+import { modules } from "../test.setup";
+import type { Id } from "../_generated/dataModel";
+import type { PayoutChargeNet } from "../lib/finance/stripeConnect";
+
+process.env.JWT_SECRET = "test-jwt-secret-for-unit-tests-minimum-32-chars";
+
+// ============================================================================
+// Provider mocks
+//
+// `vi.hoisted` because `vi.mock` factories are hoisted above the module body —
+// the shared state they close over has to be hoisted with them.
+// ============================================================================
+
+const stubs = vi.hoisted(() => ({
+  /** payoutId -> charges Stripe reports for it, or an error to throw. */
+  payouts: new Map<string, { charges: unknown[] } | { error: string }>(),
+  /** Increase idempotency keys that should fail this run. */
+  failTransferKeys: new Set<string>(),
+  /** Every createAccountTransfer call, in order. */
+  transferCalls: [] as Array<{
+    toAccountId: string;
+    amountCents: number;
+    idempotencyKey: string;
+  }>,
+}));
+
+vi.mock("../lib/finance/stripeConnect", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../lib/finance/stripeConnect")>();
+  return {
+    ...actual,
+    listPayoutChargeNets: vi.fn(async (_accountId: string, payoutId: string) => {
+      const stub = stubs.payouts.get(payoutId);
+      if (!stub) {
+        throw new Error(`test stub missing for payout ${payoutId}`);
+      }
+      if ("error" in stub) {
+        throw new Error(stub.error);
+      }
+      return stub.charges as PayoutChargeNet[];
+    }),
+  };
+});
+
+vi.mock("../lib/finance/increase", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../lib/finance/increase")>();
+  return {
+    ...actual,
+    createAccountTransfer: vi.fn(
+      async (input: {
+        toAccountId: string;
+        amountCents: number;
+        idempotencyKey: string;
+      }) => {
+        stubs.transferCalls.push({
+          toAccountId: input.toAccountId,
+          amountCents: input.amountCents,
+          idempotencyKey: input.idempotencyKey,
+        });
+        if (stubs.failTransferKeys.has(input.idempotencyKey)) {
+          throw new Error(
+            "Increase API POST /account_transfers failed (503): service unavailable",
+          );
+        }
+        return { id: `at_${input.idempotencyKey}`, status: "pending" };
+      },
+    ),
+  };
+});
+
+beforeEach(() => {
+  stubs.payouts.clear();
+  stubs.failTransferKeys.clear();
+  stubs.transferCalls.length = 0;
+});
+
+// ============================================================================
+// Fixture
+// ============================================================================
+
+interface AllocFixture {
+  communityId: Id<"communities">;
+  fundAId: Id<"funds">;
+  fundBId: Id<"funds">;
+  generalFundId: Id<"funds">;
+}
+
+const RECEIVING_ACCOUNT = "account_receiving";
+
+async function seedAllocFixture(
+  t: ReturnType<typeof convexTest>,
+  options: { receivingAccount?: string | null } = {},
+): Promise<AllocFixture> {
+  return await t.run(async (ctx) => {
+    const timestamp = Date.now();
+    const communityId = await ctx.db.insert("communities", {
+      name: "Allocation Church",
+      slug: `allocation-church-${Math.floor(Math.random() * 1_000_000)}`,
+      createdAt: timestamp,
+      updatedAt: timestamp,
+    });
+
+    const mkFund = async (
+      name: string,
+      type: "group" | "general",
+      increaseAccountId: string,
+    ) =>
+      ctx.db.insert("funds", {
+        communityId,
+        name,
+        type,
+        increaseAccountId,
+        status: "active",
+        balanceCents: 0,
+        createdAt: timestamp,
+        updatedAt: timestamp,
+      });
+
+    const fundAId = await mkFund("Fund A", "group", "account_fund_a");
+    const fundBId = await mkFund("Fund B", "group", "account_fund_b");
+    const generalFundId = await mkFund("General", "general", "account_general");
+
+    const receiving =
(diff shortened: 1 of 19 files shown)


apps/convex/lib/finance/stripeConnect.ts
```diff
@@ -136,6 +136,125 @@ export async function createAccountOnboardingLink(
...
+        ...(startingAfter ? { starting_after: startingAfter } : {}),
+      },
+      { stripeAccount: connectedAccountId },
+    );
+
+    for (const txn of result.data) {
+      // "charge" is a card payment; "payment" is the ACH-debit equivalent
+      // (ADR-032 §3 offers ACH for large gifts). Everything else is fee or
+      // reversal bookkeeping, not donor money arriving.
+      if (txn.type !== "charge" && txn.type !== "payment") continue;
```
lilseyi: **P0** — Dropping `refund` (and `adjustment`) rows is not safe. A donation refunded *before* its payout is still `allocationStatus: "pending"` (nothing in `recordDonationRefund` changes it), so `planAllocations` will match it on the surviving `charge` row and transfer the full net into the group's Increase Account — money that already went back to the donor.

The refund balance transaction's expanded `source` is a Refund object, which carries `payment_intent` — exactly the field `paymentIntentIdFromSource` below already reads. So net them: accumulate refund/adjustment rows as negative contributions under the same PaymentIntent and emit only PaymentIntents whose total is > 0.

apps/convex/lib/finance/stripeConnect.ts
```diff
@@ -136,6 +136,125 @@ export async function createAccountOnboardingLink(
...
+        console.error(
+          `[finance] listPayoutChargeNets: balance transaction ${txn.id} has a non-positive/non-integer net (${txn.net}) — skipping`,
+        );
+        continue;
+      }
+
+      nets.push({ paymentIntentId, netCents: txn.net });
+    }
+
+    if (!result.has_more) break;
```
lilseyi: **P1** — When the loop exits because `page` hit `BALANCE_TXN_MAX_PAGES` rather than because `has_more` went false, that is silent truncation on the money path: the un-read charges' donations never get bound and nothing alarms. Log (ideally audit) on cap exhaustion.

Also worth documenting here: `?payout=` is filtered to **automatic** payouts only. A community that switches to manual payouts in the Express dashboard gets an empty list forever, which lands in the silent `plan.length === 0` path in jobs.ts.

apps/convex/functions/finance/jobs.ts
```diff
@@ -109,25 +207,68 @@ export const planAllocations = internalMutation({
...
+      } else {
+        const matched = netByPaymentIntent.get(donation.stripePaymentIntentId);
+        if (matched === undefined) {
+          continue; // Not one of this payout's charges — a later payout owns it.
+        }
+        // Defensive only: the nets came out of this payout, so they sum to
+        // it by construction. `continue` (never `break`) so one oversized
+        // outlier can't wedge the queue behind it forever, which is exactly
+        // how the old gross matcher stalled.
+        if (usedCents + matched > args.payoutCents) {
```
lilseyi: **P0 context** — This residual-budget check is the only thing standing between a refunded donation and a real transfer, and it cannot do the job: it compares one charge's net against the *whole* payout, not against that PaymentIntent's own net-of-refunds. A payout carrying `charge +9680`, `refund -10000` and 50000 of other charges leaves `payoutCents` at 49680, so `9680 <= 49680` passes and the refunded gift is stamped and transferred.

Separately: when this branch does fire on a legitimately over-budget item, the donation is left unbound and pending forever — `retryStaleAllocations` only resumes *bound* donations, so the only recovery is the 3-day alert. Worth an audit event, not just a `console.error`.

lilseyi (commented):
## Independent review — allocation on Stripe NET (PR #689)

I did not write this code and I went looking for ways it loses or double-moves money. The core fix is real: gross-vs-net matching was a deterministic stall, `continue` instead of `break` is right, and reading membership off Stripe's balance transactions rather than inferring it from a running total is the correct shape. The per-item isolation is a genuine improvement over the old all-or-nothing claim.

But three of the PR's stronger claims do not survive contact, and one of them moves real money into the wrong place.

**Verdict: DO NOT MERGE** until P0-1 is resolved. P1-1 through P1-4 should be fixed or explicitly accepted in writing before this runs against production money.

---

### P0

**P0-1 — A donation refunded before its payout is still allocated at full net. Real money moves into a group fund for a gift that was already returned to the donor.**

`listPayoutChargeNets` drops `type === "refund"` rows (`lib/finance/stripeConnect.ts:212`), `recordDonationRefund` never touches `allocationStatus` (`functions/finance/giving.ts:747-819`), and `planAllocations` selects any donation that is still `"pending"` and whose PaymentIntent appears in the payout (`functions/finance/jobs.ts:204`, `237`).

Sequence:
1. Donor gives $100. `donations` row `pending`, ledger credit 10000, `funds.balanceCents` 10000.
2. Donor is refunded in full the next day, before the T+2 payout. `charge.refunded` posts a 10000 `refund` debit → ledger 0. The donation is **still `allocationStatus: "pending"`** — nothing anywhere sets it otherwise.
3. Payout `po_X` settles. Its balance transactions include `charge` (net +9680) **and** `refund` (net −10000), plus other donors' charges totalling say 50000. `payout.amount` = 49680.
4. `listPayoutChargeNets` returns `{pi_x: 9680}` — the −10000 refund row is discarded.
5. `planAllocations`: `9680 <= 49680`, so the residual-budget check at `jobs.ts:245` passes. The donation is stamped and planned.
6. Increase moves **9680 real cents** from the receiving Account into the group's Account. `recordAllocation` posts a 320 `fee` debit → `funds.balanceCents` = **−320**, while the group's bank Account now holds 9680 that belongs to the donor and is spendable on the group's Increase card.

The budget check cannot save you here because it compares against the *whole* payout, not per-PaymentIntent. The information needed to net this out is right there and is thrown away: a refund balance transaction's expanded `source` is a Refund object, which carries `payment_intent` — the very field `paymentIntentIdFromSource` already reads.

Worse, `__tests__/finance-payout-nets.test.ts:84` codifies the discard as correct: *"skips fee/payout/refund rows — they are the payout's leftover, not donor money."* A refund is not leftover. It is donor money going back out.

Minimum fix: accumulate `refund` / `adjustment` rows as **negative** contributions keyed by the same PaymentIntent and net them against the charge, and skip (or zero) any donation whose net comes out ≤ 0. Belt and braces: have `planAllocations` skip a donation that already carries a `refund` ledger debit for its charge.

---

### P1

**P1-1 — Removing `claimPayout` opened a concurrency window that the per-donation stamp does not close.** (`jobs.ts:151-273`, `jobs.ts:432-462`)

The stamp is a *marker*, not a lease. Two passes over the same payout can both come out of `planAllocations` holding the same item: pass A because it stamped it, pass B because of the resume branch at `jobs.ts:225-235`, which re-selects a donation already bound to this payout. Both then call `createAccountTransfer` with `alloc:{donationId}` (`jobs.ts:446`) — concurrently. `recordAllocation`'s already-allocated no-op (`jobs.ts:300`) is no help: at that moment neither pass has recorded anything.

Two realistic triggers:
- Stripe redelivers `payout.paid` (at-least-once, and the handler is `await`ed inline at `webhooks.ts:691`) while the first delivery is still transferring.
- The `:45` `retryStaleAllocations` cron fires while a `runAllocation` pass is mid-loop. This one is **new**: before this PR the retry cron was alert-only, so it could never race the webhook.

The old `claimPayout` insert-if-absent genuinely closed this under Convex OCC. What is left is entirely Increase's behaviour on two *in-flight* requests sharing a key — and their idempotency docs say only "at most one object is created per key", with nothing about concurrency or retention. I would not ship a money guarantee that rests on an undocumented property of a third party.

Suggest: make the stamp a lease (an `allocationTransferStartedAt` / `"transferring"` state written before the provider call and cleared on failure), or reinstate a pass-level claim.

**P1-2 — `executeAllocationItems` cannot distinguish "transfer never happened" from "transfer landed, recording failed", and audits both as `allocation.transfer_failed`.** (`jobs.ts:432-489`)

The `try` spans both `createAccountTransfer` (`:446`) and `recordAllocation` (`:454`). If the transfer succeeds and `recordAllocation` throws — `postLedgerEntry` rejects a `closed` fund at `lib/finance/ledger.ts:119`, OCC retry exhaustion, any future validation — then the money *has* moved, the audit row says it has not, and the donation stays pending-with-stamp. The hourly cron then re-issues that transfer, forever, and every replay is safe only for as long as Increase still honours the key.

No test covers this. The Increase mock (`__tests__/finance-allocation.test.ts:72-91`) does not model idempotency at all: it always records a call and always mints a fresh `at_…` id. The partial-failure test injects its failure at the *transfer*, which is the easy half. So the PR's headline "two independent locks" claim is asserted in the description and in `ARCHITECTURE.md` but exercised by nothing — the only lock the suite actually tests is `recordAllocation`'s no-op.

**P1-3 — A payout that matches zero donations is a silent, unrecoverable stall — the same failure class this PR set out to fix.** (`jobs.ts:545-547`)

`if (plan.length === 0) return { allocated: 0, failed: 0 }` — no `console.error`, no audit event, and the `processedStripePayouts` row was already written at `jobs.ts:169`. The webhook then returns 200 (`http.ts:509`), so Stripe never redelivers. `retryStaleAllocations` only ever resumes donations that already carry `allocationPayoutId` (`jobs.ts:843-861`, `jobs.ts:874-913`); an unbound donation is alert-only after three days (`jobs.ts:978-1007`). Nothing re-plans that payout, ever.

Three ways to land there:
- **Reconciliation timing.** Stripe attributes balance transactions to a payout asynchronously and emits `payout.reconciliation_completed` for precisely this reason; their reconciliation guidance is to fetch on `payout.paid` **or** `payout.reconciliation_completed`. A query at `payout.paid` can return a partial or empty set. Nothing here handles the reconciliation event.
- **Manual payouts.** `GET /v1/balance_transactions?payout=` is documented as filtering automatic payouts only. A church that flips to manual payouts in the Express dashboard — already listed in `ARCHITECTURE.md` as an unhandled payout-destination-drift risk — gets an empty list permanently.
- **The page cap.** `BALANCE_TXN_MAX_PAGES = 100` (`stripeConnect.ts:165`, loop at `:197`) truncates at 10k charges with no log on exhaustion.

And `leftoverCents` — the one number that would reveal "this payout carried money we could not map" — is computed at `jobs.ts:269` and then discarded by its only caller (`jobs.ts:530`). At minimum: alarm when `plan.length === 0` or when `leftoverCents` exceeds a plausible fee fraction of the payout, and handle `payout.reconciliation_completed`.

**P1-4 — The member-facing "spent" figure now includes Stripe's processing fees.**

`summarizePeriod` (`functions/finance/giving.ts:176-191`) buckets *every* debit into `spentCents`. With the new `fee` debit, a group that has spent nothing now reports 2.9% + 30¢ of every gift as spend: `apps/mobile/features/finance/member/FundScreenView.tsx:119` and `:132` ("… spent"), and `apps/mobile/features/finance/leader/GivingHubView.tsx:194` ("SPENT THIS MONTH"). On a transparency screen whose whole purpose is telling members where their money went, that is a wrong number, not a cosmetic one. Exclude `fee` (and arguably `refund`) from `spentCents`, or give fees their own line.

**P1-5 — The invariant table has a fourth state it does not name, and `balanceCents` can now be deterministically negative.**

A refunded-but-still-`pending` donation contributes its full **gross** to `computePendingAllocationCents` (`jobs.ts:623-626` via `allocationAmountCents`, `jobs.ts:101-105`) while the ledger has already been debited by the refund: ledger = gross − refunded, bank = 0, pending = gross → permanent drift of exactly the refunded amount, alarming nightly.

Post-allocation full refund is worse: 10000 credit − 10000 refund − 320 fee = **−320**. `funds.balanceCents` goes negative by exactly the fee on every full refund of an allocated gift, and `FundScreenView.tsx:93` renders it to members. That is arguably economically true (Stripe keeps the fee on a refund), but nothing debits it from anywhere else, nothing reconciles it, and the fund shows a negative balance forever. Disputes are the same shape (`recordDonationDisputed` is audit-only). Either cover these in the table or state explicitly that refund/dispute states are known-drifting.

---

### P2

- **P2-1** `retryStaleAllocations` reports `stalePendingCents` as the community's *entire* pending sum (`jobs.ts:984-991`), not the >3-day portion. Pre-existing, but the number is now on the net basis and lands in the audit row labelled as the stale amount.
- **P2-2** `getCommunitiesWithStrandedAllocations` and `getCommunitiesWithStalePendingDonations` (`jobs.ts:812-861`) each `.collect()` every pending donation across every community, hourly, then `db.get` each fund. Full scans on the money path.
- **P2-3** `payout.failed` is unhandled. Stripe documents that `payout.paid` can be followed by `payout.failed`. Donations stamped by the paid event stay stamped, and the new unconditional hourly resume will keep trying to move money out of a receiving Account that never received it — a live loop rather than the old one-shot.
- **P2-4** `__tests__/finance-giving.test.ts:1111` still asserts `balanceAfter === balanceBefore` with the comment "no ledger entry posted — see jobs.ts's design-decision comment". That is now only true because the fixture donation has no `payoutNetCents`. The test title and comment describe behaviour this PR removed; it passes for the wrong reason.
- **P2-5** `planAllocations` writes `processedStripePayouts` with the *first* pass's `payoutCents` and never revisits it (`jobs.ts:164-175`), though the schema comment now calls it "the bookkeeping record of a pass" (plural passes implied).

---

### On the specific claims

- **"No suite-wide fake timers, so an awaited action cannot silently no-op"** — checks out. `test.setup.ts` does no timer manipulation, `finance-allocation.test.ts` calls no `useFakeTimers`, and `stubs.transferCalls` proves the action bodies really execute. Good.
- **"Stripe is queried before any write, so an unreachable Stripe leaves the payout untouched"** — checks out, and the 500 at `http.ts:513` does get Stripe to redeliver. Verified by the test at `finance-allocation.test.ts:652`.
- **"Two independent locks"** — one of them (`recordAllocation`'s no-op) is tested; the other (the Increase key) is not tested and not documented by Increase. See P1-1/P1-2.
- **"An honest invariant"** — honest for the three states listed, silent about refunds and disputes. See P1-5.
- **`funds.balanceCents` gross → net** — I checked every read: `giving.ts:249` (transparency), `FundScreenView.tsx:93`, `GivingHubScreen.tsx:94` → `GivingHubView.tsx:173`, `CreateCardScreen.tsx:105` (client-side spend-limit hint only), and `jobs.ts:757` (the reconcile input). No backend gate — reimbursement payout and card issuance do **not** consult the balance — so nothing *decides* wrongly, and net is the more truthful display. The two problems are the "spent" bucket (P1-4) and the negative balance (P1-5).

### The one thing I would most want changed

Net refunds out of the per-PaymentIntent totals in `listPayoutChargeNets` — and drop the test that asserts refund rows are "leftover". Everything else here is recoverable bookkeeping; that one puts a donor's returned money into a group's spendable bank account.

apps/convex/functions/finance/jobs.ts
```diff
@@ -109,25 +207,68 @@ export const planAllocations = internalMutation({
...
-        break;
+      if (
+        donation.allocationPayoutId &&
+        donation.allocationPayoutId !== args.stripePayoutId
+      ) {
+        continue; // Belongs to another payout's pass.
+      }
+
+      let netCents: number;
+      if (donation.allocationPayoutId === args.stripePayoutId) {
```
lilseyi: **P1** — This resume branch is what makes the stamp a marker rather than a lease. Two passes over the same payout can both hold the same item at the same time: pass A because it just stamped it, pass B because it lands here. Both then issue `createAccountTransfer` with `alloc:{donationId}` concurrently, and `recordAllocation`'s already-allocated no-op cannot help because neither has recorded yet.

Realistic trigger: the `:45` `retryStaleAllocations` cron firing while a `runAllocation` pass is mid-loop. That race is **new** — before this PR the retry cron was alert-only, and `claimPayout` closed the webhook-vs-webhook case under Convex OCC.

That leaves the guarantee resting entirely on Increase collapsing two *in-flight* requests on one key. Their idempotency docs say only "at most one object is created per key" — nothing about concurrency or retention. Consider writing a `"transferring"` state / `allocationTransferStartedAt` before the provider call so a second pass skips the item.

apps/convex/functions/finance/jobs.ts
```diff
@@ -306,9 +461,103 @@ export const runAllocation = internalAction({
         });
       }
       allocated += 1;
+    } catch (error) {
```
lilseyi, line 770: **P1** — This `catch` spans both `createAccountTransfer` and `recordAllocation`, so it cannot distinguish "the transfer never happened" from "the transfer landed and the *recording* failed". In the second case money has already moved, the audit row says `allocation.transfer_failed` (a lie), the donation stays pending-with-stamp, and the hourly cron re-issues the transfer indefinitely — safe only for as long as Increase honours the key, which is a window they do not document.

`postLedgerEntry` throwing on a `closed` fund (lib/finance/ledger.ts:119) is one concrete way in; OCC exhaustion is another.

No test covers this path, and the Increase mock in finance-allocation.test.ts models no idempotency at all (it always records a call and always mints a fresh id) — so the PR's second "independent lock" is exercised by nothing. Suggest splitting the transfer and the record into separate try blocks with distinct audit actions, and adding a test where the transfer succeeds and `recordAllocation` throws.

apps/convex/functions/finance/jobs.ts
```diff
@@ -306,9 +461,103 @@ export const runAllocation = internalAction({
...
+        charges,
+      },
+    );
+
+    if (alreadyClaimed) {
+      console.log(
+        `[finance] runAllocation: payout ${args.stripePayoutId} was already planned — resuming ${plan.length} unfinished item(s)`,
+      );
+    }
+    if (plan.length === 0) {
```
lilseyi, line 955: **P1** — An empty plan returns silently: no `console.error`, no audit event. But the `processedStripePayouts` row was already written at line 169, the webhook returns 200 (http.ts:509) so Stripe never redelivers, and `retryStaleAllocations` only resumes donations that already carry `allocationPayoutId`. An unbound donation is alert-only after three days. So a payout that matches nothing is a permanent, silent stall — the exact failure class this PR set out to eliminate.

This is reachable: Stripe attributes balance transactions to a payout asynchronously (that is what `payout.reconciliation_completed` is for; their reconciliation guidance is to fetch on `payout.paid` *or* that event), `?payout=` filters automatic payouts only, and the 100-page cap truncates silently.

Alarm here, and consider handling `payout.reconciliation_completed`.

apps/convex/functions/finance/jobs.ts
```diff
@@ -306,9 +461,103 @@ export const runAllocation = internalAction({
...
+      return { allocated: 0, failed: 0 };
+    }
+
+    const { listPayoutChargeNets } = await import("../../lib/finance/stripeConnect");
+    const charges = await listPayoutChargeNets(
+      communityFinance.stripeConnectedAccountId,
+      args.stripePayoutId,
+    );
+
+    const { plan, alreadyClaimed } = await ctx.runMutation(
```
lilseyi: **P2** — `leftoverCents` is the one signal that would say "this payout carried money we could not map to a donation", and it is computed at line 269 and then dropped on the floor here. Destructure it and alarm when it exceeds a plausible fee fraction of `payoutCents`.

apps/convex/functions/finance/jobs.ts
```diff
@@ -163,133 +308,143 @@ export const recordAllocation = internalMutation({
...
+    // than inventing a number.
+    const feeCents = netCents === undefined ? 0 : grossCents - netCents;
+    if (feeCents < 0) {
+      // Net above gross is impossible from Stripe; refusing to post rather
+      // than crediting a fund for a number we don't understand.
+      console.error(
+        `[finance] recordAllocation: donation ${args.donationId} has net ${netCents} above gross ${grossCents} — not posting a fee entry`,
+      );
+    } else if (feeCents > 0) {
+      await postLedgerEntry(ctx, {
```
lilseyi, line 555: **P1** — The fee debit itself is idempotent and correctly signed, but two consequences are unhandled:

1. `summarizePeriod` in giving.ts:176-191 buckets *every* debit into `spentCents`, so this entry now inflates the member-facing "… spent" (FundScreenView.tsx:119/:132) and "SPENT THIS MONTH" (GivingHubView.tsx:194) by Stripe's processing fees. A group that has spent nothing will report spending 2.9% + 30¢ of every gift.
2. Combined with a later full refund of the same donation, `funds.balanceCents` goes deterministically negative: 10000 credit − 10000 refund − 320 fee = −320, rendered as-is at FundScreenView.tsx:93. Nothing corrects it and nothing reconciles it.

apps/convex/functions/finance/jobs.ts
```diff
@@ -362,7 +621,7 @@ export const computePendingAllocationCents = internalMutation({
         .filter((q) => q.eq(q.field("allocationStatus"), "pending"))
         .collect();
       const pendingCents = pendingDonations.reduce(
```
lilseyi, line 1178: **P1** — The three-state table in the file header does not cover a refunded donation. A `pending` donation that has been (partially or fully) refunded still contributes its full gross here, while the ledger has already taken the refund debit: ledger = gross − refunded, bank = 0, pending = gross → permanent drift of exactly the refunded amount, alarming nightly and never closing. Disputes are the same shape (`recordDonationDisputed` is audit-only).

Either exclude refunded amounts from the pending sum, or say explicitly in the header that refund/dispute states are known-drifting.

apps/convex/__tests__/finance-payout-nets.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    expect(stubs.calls[0].params).toMatchObject({
+      payout: "po_abc",
+      expand: ["data.source"],
+    });
+    // Payouts and their charges live on the CONNECTED account, not the
+    // platform account.
+    expect(stubs.calls[0].options).toEqual({ stripeAccount: "acct_connected" });
+  });
+
+  test("skips fee/payout/refund rows — they are the payout's leftover, not donor money", async () => {
```
lilseyi: **P0** — This test asserts the P0 bug is correct behaviour. A `refund` row is not "the payout's leftover" — it is donor money going back out, and the donation it belongs to is still `allocationStatus: "pending"`, so dropping the row lets `planAllocations` transfer the full net into the group's Account for a gift that was refunded.

The test should instead assert that a `charge +9680` / `refund -10000` pair on the same PaymentIntent yields **no** entry for that PaymentIntent (or a netted negative that is then filtered), while `stripe_fee` and `payout` rows are still skipped.

apps/convex/__tests__/finance-allocation.test.ts
```diff
@@ -0,0 +1,898 @@
...
+      return stub.charges as PayoutChargeNet[];
+    }),
+  };
+});
+
+vi.mock("../lib/finance/increase", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../lib/finance/increase")>();
+  return {
+    ...actual,
+    createAccountTransfer: vi.fn(
```
lilseyi, line 99: **P1** — This mock models no idempotency: it unconditionally pushes to `transferCalls` and mints a fresh `at_${key}` every call. That means the `alloc:{donationId}` Increase key — one of the two locks the PR's exactly-once claim rests on — is never actually exercised. Every "no second transfer" assertion in this suite is really testing `recordAllocation`'s already-allocated no-op, which is the other lock.

To test the claim: have the mock return the *same* transfer id for a repeated key and record a `deduped: true`, then add a case where `createAccountTransfer` succeeds and `recordAllocation` throws — the transfer-landed-then-crashed case the PR names explicitly — and assert the retry produces one transfer, one fee entry, one `donation.allocated` audit row.

apps/convex/__tests__/finance-allocation.test.ts
```diff
@@ -0,0 +1,898 @@
...
+        .query("ledgerEntries")
+        .withIndex("by_fund", (q) => q.eq("fundId", fundAId))
+        .collect(),
+    );
+    const fee = feeEntries.find((e) => e.kind === "fee");
+    expect(fee).toMatchObject({ direction: "debit", amountCents: 320 });
+    expect(fee?.stripeObjectId).toBe("po_happy");
+  });
+
+  test("PARTIAL FAILURE: item 2 of 3 fails, 1 and 3 still land, and a redelivery finishes 2 without re-transferring 1", async () => {
```
lilseyi, line 587: **P1 (test gap)** — This is the suite's strongest test, but its failure is injected at the *transfer*, which is the recoverable half. Missing cases the PR's claims depend on:

- transfer succeeds, `recordAllocation` throws (money moved, nothing recorded)
- two passes over the same payout overlapping (webhook redelivery, or `retryStaleAllocations` racing an in-flight `runAllocation`)
- a payout containing a refund for a still-`pending` donation
- `plan.length === 0` (Stripe returned nothing for the payout)

lilseyi (commented):
## Re-review — verifying the P0 fix (commit `c6c4350b`)

I re-ran my original failure sequence against the new code, then went looking for variants, wrong sign assumptions, and anything the fix broke. **The P0 I reported is closed.** Three of the eight things I flagged are now genuinely fixed rather than papered over. But I found three new issues, two of which I reproduced against this branch, and one of them is a regression introduced by the fix for P1-3.

**Verdict: MERGE AFTER FIXES** (P1-6, P1-7, P1-8 below). This is much closer than last time — none of these are architectural, all three are a few lines.

---

### The P0 is closed

Original sequence, re-walked:

1. `getPayoutComposition` (`lib/finance/stripeConnect.ts:254`) sums the `refund` row against its charge's PaymentIntent instead of discarding it, and only emits strictly-positive totals (`:349`). The refunded gift lands in `reversedPaymentIntentIds` and never reaches the planner.
2. Independently, `recordDonationRefund` (`functions/finance/giving.ts:863-868`) flips a fully-refunded pending gift to the terminal `"refunded"` status, and every selection query filters `allocationStatus === "pending"` (`jobs.ts:318`, `:1310`, `:1029`).

Those two are genuinely independent, which matters: #1 alone would miss a refund whose balance transaction settles in a *different* payout than the charge (Stripe does not guarantee they share an `available_on` bucket), and #2 catches exactly that case. `__tests__/finance-allocation.test.ts:985` asserts both — the status flip *and* zero transfer calls — and `finance-payout-nets.test.ts:114` replaced the test that encoded the bug. The tests would fail against the old code; I checked they assert on `stubs.transferCalls`, on the ledger entries, and on the fund balance, not just on return values.

**The signed-`net` assumption is correct.** Stripe: *"A positive value represents incrementing a Stripe balance, and a negative value decrementing a Stripe balance."* Summing `net` as-is is right for every type in the list, including `refund_failure` (positive — funds returned to balance) and `adjustment` in both its dispute and dispute-reversal directions.

**Arithmetic.** I walked all six cases; the invariant holds or drifts exactly as documented:

| case | ledger | bank | pending | invariant |
|---|---|---|---|---|
| no refund | gross − fee | net | 0 | holds |
| partial refund, same payout | 10000 − 3000 − **320** | 6680 | 0 | holds (the fee is 320, not 3320 — the double-charge is real and really avoided) |
| partial refund before payout | gross − refund | 0 | gross − refund | holds |
| full refund before payout | 0 | 0 | 0 | holds (terminal `"refunded"`) |
| full refund after allocation | 0 (fee reversed) | net | 0 | drifts by net — documented |
| two sequential partials | 0 | net | 0 | drifts — documented |

The fee-reversal boundary is exactly right, which I did not expect: refund **==** net (9680) leaves the fund at 0 and does *not* reverse; refund 9681 reverses and leaves 319. It cannot post twice (`alloc-fee-reversal:{donationId}` is keyed per donation, not per refund step, and `postLedgerEntry` dedupes at `ledger.ts:99`), and it cannot post with no fee debit (`allocationFeeCents > 0`, `giving.ts:883`). A refund arriving *between* transfer and recording is also handled — `recordAllocation:436` sees `"refunded"` and declines to post the fee, so the -320 can't be created in the first place.

---

## New findings

### P1-6 — The reversal-type list is an allowlist, it is missing a real Stripe type, and a miss is silent

`REVERSAL_TXN_TYPES` (`lib/finance/stripeConnect.ts:214-221`) contains `payment_refund_failure`, **which is not a balance-transaction type Stripe emits** — it is not in the API enum. And it omits **`payment_reversal`, which is**: *"Created when a debit/failure related to a payment is detected from a banking partner. This balance transaction takes funds that were previously credited to the merchant for a payment out of the merchant balance."* That is precisely a post-credit reversal of a `payment` — the ACH type this file deliberately treats as inbound (`:199`, and ACH is live: `automatic_payment_methods: {enabled: true}` at `giving.ts:478` plus the `us_bank_account_ach_payments` capability at `stripeConnect.ts:110`).

The reason this is P1 and not P2 is what happens on a miss. I ran it: two $500 ACH gifts, one reversed, payout 49600.

```
[finance] planAllocations: donation … (net 49600) would overrun payout po_r (49600, 49600 used) — skipping it, not the rest
TRANSFERS:        [{ amountCents: 49600, idempotencyKey: 'alloc:<reversedGift>' }]
reversed status:  allocated
healthy  status:  pending
ALARMED?          false
```

The **reversed** gift is funded and the **healthy** gift is silently stranded — the residual-budget check at `jobs.ts:361` absorbs the shortfall by skipping whichever donation happens to be last in `createdAt` order, so `leftoverCents` comes out 0 and `unmatchedPayoutIsAlarming` never fires. No audit row. The only trace is one `console.error` about an overrun.

An allowlist is the wrong shape here because the failure is one-directional: every omission over-allocates. Suggestions, cheapest first:
- add `payment_reversal` (and drop the non-existent `payment_refund_failure`);
- better, make the *inbound* set the allowlist and treat every other `payment_intent`-bearing row as a signed contribution;
- best, have `getPayoutComposition` also return the sum of **every** row's `net` and have `runAllocation` assert it equals `payout.amount`. That single check catches every future type Stripe adds, which an enumerated list cannot.

### P1-7 — A partial refund that settles in a different payout still over-transfers (reproduced)

`planAllocations` takes Stripe's charge net verbatim (`jobs.ts:367`) and the resume branch takes the stamped `payoutNetCents` verbatim (`jobs.ts:345`). Neither consults `donations.refundedCents`. `allocationAmountCents` (`jobs.ts:144-163`) already computes the right number — `min(payoutNetCents, gross − refunded)` — but it is only ever called by `computePendingAllocationCents`. So the planner and the invariant disagree about what the gift is worth.

$100 gift, $30 refunded before the payout, refund's balance transaction settling in a different payout (so the composition shows the charge at its full net):

```
TRANSFERS:       [{ amountCents: 9680, … }]
LEDGER BALANCE:  7000    BANK (transferred): 9680
ENTRIES:         [ 'donation:credit:10000', 'refund:debit:3000' ]
UNMATCHED ALARM? false
[finance] recordAllocation: donation … has net 9680 above its unrefunded gross
          (10000 gross − 3000 refunded) — not posting a fee entry
```

2680 cents of money the donor already has back is now sitting in the group's spendable Increase Account. It is the P0's shape at reduced magnitude, and the full-refund belt-and-braces doesn't reach it because the gift legitimately stays `"pending"`. It does surface as nightly drift — but only after the money moved.

Fix is one line in each selection path: cap at `donationTotalCents(donation) − (donation.refundedCents ?? 0)`, i.e. just call the helper you already wrote. `finance-allocation.test.ts:1242` covers this shape for the *pending sum* only; it never runs the transfer, which is why it passes.

### P1-8 — The alarm added for P1-3 now fires on every healthy payout, and this commit guarantees it (reproduced)

`jobs.ts:821` is `if (plan.length === 0 || unmatchedIsAlarming)` with no `alreadyClaimed` guard. A redelivery of a payout that was fully allocated produces an empty plan — that is the *success* condition — and it writes `allocation.payout_unmatched` with `leftoverCents` equal to the entire payout, plus a `console.error`:

```
AUDITS AFTER HEALTHY REPLAY: [ 'donation.allocated', 'allocation.payout_unmatched' ]
UNMATCHED ROW: { stripePayoutId: 'po_1', payoutCents: 9680, leftoverCents: 9680,
                 matchedCount: 0, chargeCount: 1, reversedCount: 0 }
```

This isn't a rare edge: the same commit routes `payout.reconciliation_completed` to the same handler (`webhooks.ts:684-685`), so in normal operation `payout.paid` allocates and `payout.reconciliation_completed` immediately raises a full-payout "unmatched" alarm. Every payout, forever. That turns the one signal added for the no-automatic-recovery case into pure noise, which is worse than not having added it. `finance-allocation.test.ts:674` replays a successful payout but only asserts on `donation.allocated`, so it doesn't catch this.

Guard on `alreadyClaimed && plan.length === 0 && everything-bound-is-allocated`, or simply skip the empty-plan alarm when `alreadyClaimed` is true and the payout has at least one allocated donation.

---

### P2

- **P2-6** `runAllocation` does **not** error on `unattributedReversalCents` — `jobs.ts:791-795` logs and proceeds to plan and transfer. Defensible (the budget check absorbs the shortfall by skipping donations), but the PR narrative says it errors. Worth aligning one or the other. Note the shortfall is absorbed by skipping an arbitrary donation, per P1-6.
- **P2-7** The dispute claim in `ARCHITECTURE.md` — *"a disputed gift is never transferred out of the receiving Account"* — holds only when the chargeback `adjustment` shares a payout with the charge. Disputes typically arrive weeks after the payout, when the gift is already allocated; and unlike refunds, `recordDonationDisputed` (`giving.ts:921`) sets no `refundedCents` and no status, so a dispute filed *before* the payout but settling in a different one funds the gift in full. Say "safe when the adjustment settles in the same payout", not "never".
- **P2-8** `unbindFailedPayout` (`jobs.ts:929-960`) clears `payoutNetCents` and the lease without checking whether a transfer is in flight. A `payout.failed` racing an in-flight pass leaves `recordAllocation` flipping the gift to `"allocated"` with no net and therefore no fee entry — drift of exactly the fee. Skip donations whose lease is live.
- **P2-9** `FundPeriodTotals.refundedCents` is computed (`giving.ts:210`) and typed (`member/types.ts`) but rendered nowhere. The stated rationale for showing fees — *"hiding them would leave 'given' minus 'spent' not reconciling"* — applies identically to refunds, and today it doesn't reconcile when one exists.
- **P2-10** If `charge.refunded` is processed *after* `recordAllocation`, the refund is booked as a `fee` (`gross − 0 − net`), the later refund debit stacks on top, and the reversal condition can't fire. Narrow ordering window; worth a comment at `jobs.ts:480` if not a fix.
- **P2-11** The lease is stamped once for the whole batch (`jobs.ts:381`), not refreshed per item, so its safety rests on `ALLOCATION_LEASE_TTL_MS` (15m) exceeding Convex's action time limit (10m). That's true today and the comment gestures at it, but nothing asserts it and a batch that grows past the limit fails differently. A one-line note naming the 10-minute limit would make the reasoning auditable.

---

### Previously-flagged items — status

- **P1-1 (lease)** — Genuinely fixed, not papered over. The claim happens *inside* `planAllocations`/`listResumableAllocations` in the same transaction that returns the item (`jobs.ts:374-382`, `:1318-1319`), so Convex's OCC serializes it: a loser re-reads and sees the stamp. Failure releases immediately (`recordAllocationFailure:554`), and expiry is reclaimable (test at `:454`). The race test at `:417` is sequential rather than concurrent — unavoidable in `convex-test` — but it exercises the right code path. ✅
- **P1-2 (transfer vs record)** — Fixed properly. Separate try blocks (`jobs.ts:667`/`:700`), distinct audit actions, and — the part I specifically doubted — **the Increase mock now models key dedupe** (`finance-allocation.test.ts:113-120`), records the attempt *before* deduping so `transferCalls` stays honest, and the test asserts `stubs.dedupedKeys` and `transfersByKey.size === 1` (`:1398`). That lock is now genuinely exercised. ✅
- **P1-3 (silent unmatched payout)** — Mostly fixed: alarm added, `leftoverCents` now used instead of discarded, page-cap exhaustion throws instead of allocating on a prefix (`stripeConnect.ts:336`), `payout.reconciliation_completed` and `payout.failed` both routed. Undermined by P1-8. Manual payouts still yield a permanent empty composition — now loud, still unrecoverable, which is the honest state. ⚠️
- **P1-4 (fees as spend)** — Fixed. `summarizePeriod` buckets `fee`/`refund` separately and signs them (`giving.ts:205-210`), both surfaces show fees on their own line, and the tests assert the exact strings and the omit-when-zero case. ✅
- **P1-5 (negative balance / missing invariant states)** — Fixed, with the boundary arithmetic correct (above). The refund and dispute rows are now in the table and the deliberately-drifting case says so. ✅
- **P2-1** stale sum now scoped via `createdBeforeMs`. ✅ **P2-3** `payout.failed` handled. ✅ **P2-4** test comment now honest about *why* it passes. ✅ **P2-5** schema comment corrected. ✅ **P2-2** full scans remain, now documented with the condition under which they become wrong — acceptable.

The one thing I'd fix first is **P1-8**: it's the smallest change and it's currently degrading the alarm that the rest of the observability work depends on.

PR #690, opened 2026-07-31
The General Fund never had a bank account, and a leader could hand themselves the keys

lilseyi:
> **Stacked on #689.** Base is `finance/allocation-net-matching`; review that first. The diff shown here is only this branch's four commits.

Four verified holes in the finance surface, found by auditing ADR-032's promises against the code.

## 1. The community General Account was never built

ADR-032 §1 specifies a receiving Account **and** a General Account under each community's Entity. Only two `createAccount` call sites existed — receiving and per-group — and `recordProvisioned` inserted the "General Fund" row with **no `increaseAccountId`**.

Consequences: general-fund donations were marked `"allocated"` with no transfer at all, on the premise that the General Account is where Stripe already landed the money (it lands in the *receiving* Account), and `getFundsWithIncreaseAccount` filters on `increaseAccountId`, so the General Fund was excluded from nightly reconcile entirely. Its ledger balance was not bank-backed and nothing alarmed.

`provisionProviders` now mints the General Account alongside the receiving one. `runAllocation` lost its general-fund special case — every fund gets a real receiving→own-Account transfer. A fund with no Account now **fails** its item (audited, retry-eligible, donation stays `pending` with its payout stamp) rather than being recorded as allocated.

**Backfill, three layers:** a re-runnable `backfillGeneralFundAccounts` internal action; opportunistic self-heal when an admin touches `enableGroupGiving`; and a patch path if provisioning is re-run. Communities that never finished onboarding are skipped — they get theirs from `provisionProviders`.

## 2. A group leader could self-escalate and issue themselves a card

`requireFundRoleOrGroupLeader` returned early for any active group leader **regardless of `minRole`**, and `grantFundRole` calls it with `"finance_admin"`. So: leader → grant self `finance_admin` → `createFundCard(self)` → spend the fund.

ADR-032 §4 does intend leaders to bootstrap finance roles, so the carve-out stays. The helpers now report **which** path authorised the call (`community_admin | fund_role | group_leader`), and a self-targeted grant is refused only when the answer is `group_leader`. The strong paths resolve first, so a community admin who also leads the group is unaffected. A test asserts the whole chain is closed: blocked self-grant → `createFundCard` still rejects.

**Residual risk, deliberately left open and re-documented in COMPLIANCE.md:** a leader can still grant `manager` to a second account they control and approve with it. No permission check distinguishes that from two real people; detection heuristics are the honest answer, not a gate.

## 3. `enableGroupGiving` didn't check the community was live

It required community admin but never checked `onboardingStatus === "live"`. If it wasn't, provisioning silently no-op'd with a `console.error` **after** the admin already saw a success toast — leaving a fund with no bank account and no error anywhere. Now refused up front with three distinct human messages naming where to go.

## 4. Feature-flag gaps, and the ones left open on purpose

Added `requireGroupGivingEnabled` to `listExpenses`, `listMyExpenses`, `listFundRoles`, `getMyFundRole`.

**Deliberately NOT gated: `denyExpense` and `revokeFundRole`.** Both are pure de-escalation, the switch is most likely flipped *during* an incident, and gating them would trap pending expenses and block revoking a compromised finance_admin. Comments and tests record the intent so nobody "fixes" it later.

## 5. Receipt provenance

`submitExpense` accepted any `r2:` string, so a submitter could attach someone else's receipt image. No ownership mechanism existed. Both presign actions in `functions/uploads.ts` now record an `uploadGrants` row (storage path → uploading user) at the one moment we know both, and `submitExpense` rejects a receipt whose grant is missing or belongs to someone else.

**Reviewer note — this one is app-wide, not finance-scoped.** Every chat/profile/group upload now writes a row. I judged that better than duplicating the presign path inside finance, but it has a wider blast radius than the other three fixes and **no retention policy** — the table grows unbounded. A cleanup cron is a fair follow-up ask.

Duplicate receipts stay possible (re-attaching your *own* receipt to several expenses) — different threat, flagged as an open item rather than silently expanded into.

## Tests

**161 files, 2744 tests passing** with #689's suites together. New: `finance-general-account.test.ts` (8), `finance-roles.test.ts` (7); extended onboarding, expenses, flag, ledger. Coverage verified by stashing the fixes and re-running — **15 of the new tests fail without them**.

`finance-general-account.test.ts` deliberately breaks the house pattern: suite-wide `vi.useFakeTimers()` would leave `provisionProviders`/`runAllocation`/`runNightlyReconcile` queued and unexecuted, so it mocks the provider modules and invokes the actions directly, asserting on what was asked of the bank.

## Ops

`backfillGeneralFundAccounts` must be run against staging after deploy, and against production before giving goes live there.


lilseyi (commented):
Independent review. I did not write this code; I went looking for what breaks. Read ADR-032 §1/§4 and `functions/finance/ARCHITECTURE.md` first, then traced every caller of the changed helpers and pulled Increase's live OpenAPI spec where the diff makes provider claims.

**P0 — none.** I could not find an authorisation bypass. The self-escalation fix genuinely holds (details under §2 below), and the allocation change genuinely leaves the donation recoverable.

---

## P1

### 1. The three backfill paths do NOT share an Increase idempotency key — concurrent runs mint two General Accounts

`functions/finance/onboarding.ts:347` mints the General Account under
`finance:general-account:${communityId}:${fp}`, but `provisionFundAccount` (`onboarding.ts:935`) — the *other two* backfill layers, the migration and the `enableGroupGiving` self-heal — mints the same logical account under `finance:group-account:${fundId}`.

Increase dedupes on the key. Two different keys = two different Accounts. The comment at `onboarding.ts:931` explicitly reasons about not changing the `group-account` prefix "because it would ask Increase for a SECOND account" — and then the new code introduces exactly that second key from the other side.

Interleaving (all three are non-transactional actions):

```
backfill:            read fund -> no account
provisionProviders:  getGeneralFundAccountId -> null
provisionProviders:  createAccount(general-account:cid:fp) -> B
backfill:            createAccount(group-account:fundId)   -> A
provisionProviders:  recordProvisioned -> fund.increaseAccountId = B
backfill:            recordFundAccount -> guarded no-op
```

`recordFundAccount` and `recordProvisioned` both guard on `!fund.increaseAccountId`, so the *fund* stays single-valued — no data corruption, and no money can land in the loser. But an orphan bank Account exists under the church's Increase Entity, unreferenced, unaudited (the losing write no-ops silently with no audit row), and outside reconcile. That is a compliance artefact on a KYB'd entity, and the PR body + `ARCHITECTURE.md` both assert it cannot happen.

Fix: use one key for the General Account across all three paths — `finance:general-account:${fundId}` (fund-scoped, fingerprint-free) in both `provisionProviders` and `provisionFundAccount`. Also worth logging when a `recordProvisioned` / `recordFundAccount` guard drops a *different* account id on the floor, so ops learns an orphan exists.

Not covered by tests: `finance-general-account.test.ts` asserts `createAccount.mock.calls.map(c => c[1])` (the **name**), never `c[2]` (the key). Assert the key.

### 2. After the backfill, the nightly reconcile alarms forever on historically "allocated" general-fund donations, and this PR ships no sweep

This is the one that will actually bite, and the Ops section of the PR does not mention it.

Pre-fix, general-fund donations were flipped to `"allocated"` with no transfer, and `recordAllocation` posted the `fee` debit. So for any such donation:

- ledger credit (gross) − `fee` debit = **net** on the General fund
- `computePendingAllocationCents` (`jobs.ts:615`) counts only `allocationStatus === "pending"` → **0**
- the money is physically in the **receiving** Account → General Account balance **0**

Today that is invisible because `getFundsWithIncreaseAccount` filters the fund out. The moment `backfillGeneralFundAccounts` mints the Account, the fund enters reconcile scope and `runNightlyReconcile` reports `drift = the entire historical general-fund net balance`, every night, permanently. Nothing in this PR moves that cash.

Reconcile drift is the single control that would catch real theft. Making it fire unconditionally on every community with pre-existing general giving destroys it by alarm fatigue.

Before merge, please either:
- confirm (and state in the PR) that zero general-fund donations are in `allocationStatus: "allocated"` in staging **and** production, or
- add the receiving → General catch-up AccountTransfer to the backfill (sum the already-allocated general nets, one keyed transfer, paired sweep entries), or
- at minimum add a go/no-go step to the runbook: query that count first, and do not run the backfill until the sweep exists.

### 3. `uploadGrants` turns a finance check into an app-wide upload dependency, with a new hard-failure mode

`functions/uploads.ts:381` and `:491` now `await ctx.runMutation(...)` **before** returning the presigned URL, on every upload in the product — chat images, profile photos, group covers, event posters. If that mutation fails for any reason (write conflict, validator drift, table-level issue), the presign action throws and **nobody in the app can upload anything**. A finance-only provenance check should not be able to take down chat image sends.

Suggestion: make the grant write non-fatal (try/catch + log) and let `submitExpense` be the thing that fails when a grant is missing — the security property is enforced at the *read*, not the write, so a lost grant row costs one user one re-upload rather than costing everyone all uploads.

Two more gaps in the claim that "`functions/uploads.ts` records a row for every key this backend hands out":

- `lib/r2.ts`'s `putR2Object` is a **second** `r2:` producer (used by `functions/devAssistant/config.ts`, `functions/pcoServices/songImport.ts`). Not a receipt path today, but `schema.ts`'s and `ARCHITECTURE.md`'s wording implies total coverage. Please scope the claim.
- Legitimate false reject at the deploy boundary: a user who picked and uploaded a receipt photo *before* the deploy and submits *after* it gets `"That receipt wasn't uploaded from this account"` with no way to understand why. `ARCHITECTURE.md:127` calls this "harmless for expenses already stored" — it is, but it is not harmless for an in-flight submission. Worth a friendlier message ("re-attach the photo") and a note in the deploy runbook.

### 4. `backfillGeneralFundAccounts` has no per-fund error isolation, and `provisioned` reports intent, not outcome

`migrations/backfillGeneralFundAccounts.ts:68` — `await ctx.runAction(...)` in a bare loop. One community whose Increase Entity is invalid throws and aborts the whole run mid-way; ops sees a stack trace and no idea how far it got. And `return { provisioned: fundIds.length }` counts the funds it *intended* to do, not the ones where `recordFundAccount` actually wrote — `provisionFundAccount` returns `void` and fails soft on a missing Entity, so a run can report `provisioned: 12` having provisioned zero.

Wrap each fund in try/catch, count actual successes, and return `{ provisioned, skipped, failed }`. `finance-general-account.test.ts:570` asserts `result.provisioned` — that assertion currently passes on a no-op.

---

## P2

- **`enableGroupGiving`'s new live-gate has a foot-gun via `startOnboarding`.** `startOnboarding` unconditionally patches `onboardingStatus: "collecting"` (`onboarding.ts:~175`) even on a `"live"` community. Pre-PR that was cosmetic; now an admin who corrects their address on a live community silently loses the ability to enable giving on any group until Stripe re-verifies, with copy ("isn't finished yet") that will read as a bug. Either refuse the intake-form edit once live, or don't regress the status on resubmission.
- **Naming inconsistency at the bank.** `provisionProviders` names the account `"General Account"`; the backfill/self-heal names it `fund.name` = `"General Fund"`. Same logical account, two names in the Increase dashboard depending on which path created it.
- **Allocation retry noise for unhealed communities.** With the new per-item throw (`jobs.ts:454`), every payout containing a general-fund donation now fails that item, and `retryStaleAllocations` re-attempts hourly, writing an `allocation.transfer_failed` audit row each time, indefinitely until the backfill runs. The failure semantics are right; consider backing off or de-duplicating the audit row.
- **`uploadGrants` has no retention and no deletion cascade.** The PR flags growth; it also isn't registered in any account-deletion path. Worth a row in the data-deletion checklist even before the cleanup cron.
- **`resolveFundAccess` now runs `isCommunityAdmin` before the leader check** in `requireFundRoleOrGroupLeader` (`lib/helpers.ts:373`). Correct — the ordering is what makes the admin-who-leads case work — but it adds an admin lookup to hot read paths (`listExpenses`, `listFundCards`, `getCardDetail`) that previously short-circuited. Noting, not asking for a change.

---

## What I checked and found sound

**§2, the self-escalation fix — I tried to get around it and could not.** Every caller of both helpers:

| call site | helper | min | via `group_leader` reachable? | consequence |
|---|---|---|---|---|
| `roles.ts:71` `listFundRoles` | OrGroupLeader | manager | yes | read only |
| `roles.ts:182` `grantFundRole` | OrGroupLeader | finance_admin | yes | **guarded at `roles.ts:190`** |
| `roles.ts:270` `revokeFundRole` | OrGroupLeader | finance_admin | yes | de-escalation only; last-finance_admin guard at `:282` correctly excludes non-admins |
| `expenses.ts:249` `approveExpense` | `requireFundRole` | manager | **no** | leader cannot approve without a real grant |
| `expenses.ts:321` `denyExpense` | `requireFundRole` | manager | **no** | |
| `expenses.ts:579` `listExpenses` | OrGroupLeader | manager | yes | read only |
| `cards.ts:159` `createFundCard` | `requireFundRole` | finance_admin | **no** | the ladder's second rung is closed independently |
| `cards.ts:412` `cancelCard` | `requireFundRole` | finance_admin | **no** | |
| `cards.ts:119`/`:503` list/detail | OrGroupLeader | cardholder | yes | read only |
| `cards.ts:84` `isFundFinanceAdmin` | `requireFundRole` | finance_admin | **no** | |

Attempts that fail correctly: leader also holding `cardholder` or `manager` still resolves to `group_leader` (`hasFundRole` fails the `finance_admin` rank, so `resolveFundAccess` returns `null`) — the guard fires. Leader who is a community admin resolves to `community_admin` because `isCommunityAdmin` is checked first in `resolveFundAccess` — correct per the ADR. General fund has no `groupId`, so the carve-out is unreachable there at all. Revoke-then-regrant doesn't help: the guard is on the *target*, not on prior state. `grantFundRole:206` additionally requires the target to be an active group member, which closes granting to an out-of-community sock puppet. The `requireFundRole` signature change from `void` to `FundAccessVia` has no caller that misuses the new return value.

The residual you documented (leader grants `manager` to a second account they control) is correctly characterised and correctly left to detection rather than a gate.

**§4, allocation.** Confirmed recoverable: the throw at `jobs.ts:454` is inside the per-item `try`, so it hits `recordAllocationFailure` (audit only), `recordAllocation` is never called, no `fee` debit is posted, and the donation keeps `allocationStatus: "pending"` with its `allocationPayoutId` / `payoutNetCents` stamp, which is exactly what `listResumableAllocations` selects on. Nothing marks it allocated.

**Tests.** `finance-general-account.test.ts` is the right call and the reasoning in its header is correct — it mocks the providers and invokes `provisionProviders` / `provisionFundAccount` / `runAllocation` / `runNightlyReconcile` directly, so the assertions are about what was asked of the bank rather than about a scheduler that never ran. `finance-roles.test.ts`'s suite-wide `vi.useFakeTimers()` is safe here because every assertion is on a rejection or on a row, never on a scheduled action. Two gaps: the Increase idempotency keys are never asserted (see P1.1), and `result.provisioned` is asserted against a counter that can't fail (see P1.4).

---

The single most important thing to change: **decide the historical general-fund drift before anyone runs `backfillGeneralFundAccounts` in production** (P1.2). The backfill as written switches on a nightly alarm that can never be cleared, on the one control that would catch real theft. Either prove the count is zero or ship the catch-up sweep with it.

PR #692, opened 2026-07-31
Card spend limits were never sent to the bank — and the UI said they were

lilseyi:
> **Stacked on #690** (which is stacked on #689). Review those first.

## Why

`createCard` posted `{ account_id, description }` and nothing else (`lib/finance/increase.ts`). `spendLimitCents` and `limitPeriod` were stored locally, audited, and displayed — and read by nothing that made a decision. A cardholder could spend the fund's entire Increase Account balance in one swipe.

The UI did not present this as advisory. It stated three controls as fact, on one screen:

- "Resets every Monday" / "Resets on the 1st of the month" / "Applies to each individual charge" — nothing reset, nothing applied.
- "Require receipts — Charges flag for approval until a receipt is attached: **On**" — a hardcoded value in a settings cell, with no implementation anywhere in the repo.
- "Charges over $200 need a second approver" — false for cards. A swipe settles via `recordCardSettlement` into an **already-spent** `pending` `card_charge` expense. There is no authorization path and no approval gate for a card charge.

Advertising unenforced financial controls to a church treasurer is a trust and liability problem, not a copy nit.

## The premise everyone was working from was wrong

`cards.ts` documented limits as "ADVISORY ONLY", and ADR-032's Phase-2 plan deferred enforcement to Increase real-time-authorization webhooks. **Neither was necessary.** Verified two ways:

- Increase's SDK types: `CardCreateParams.authorization_controls.usage.multi_use.spending_limits: [{ interval, settlement_amount, merchant_category_codes? }]`, intervals `all_time | per_transaction | per_day | per_week | per_month`. Identical shape on `CardUpdateParams`.
- [increase.com/documentation/launch-a-card-program](https://increase.com/documentation/launch-a-card-program): "Set spending limits with `authorization_controls.usage.multi_use.spending_limits`… **Increase enforces these controls at authorization time without a round trip to your server.**"

Real-time decisioning is the *other*, optional mechanism, for logic Increase can't express declaratively. We were leaving a free, bank-enforced control on the table.

## What changed

**Limits are real.** New `lib/finance/cardPolicy.ts` (period↔interval map, `validateCardLimit`, UTC window math, mirroring the `expensePolicy.ts` precedent). `createCard` takes the limit and always emits `authorization_controls`; new `updateCardSpendingLimit`. An empty `spending_limits: []` is sent explicitly to *clear* a limit — omitting the field on PATCH would silently keep the old cap. New `setCardLimit` → `applyCardLimit` → `recordCardLimit`, with the local mirror written only after Increase confirms.

**Copy tells the truth.** Hints now state Increase's real reset semantics (UTC midnight — Mondays for weekly, the 1st for monthly). The fake "Require receipts" cell is **deleted**, not reworded, on both the create form and card detail. The false second-approver claim is gone. An uncapped card now warns that it can spend the whole fund.

**The kill switch no longer blocks incident response.** `setCardFrozen` and `cancelCard` no longer call `requireGroupGivingEnabled`. Freezing is de-escalation; a card is a bank object that keeps working when the flag is off, so gating the freeze protected nothing and removed the only remedy. Role checks are untouched (holder can freeze, only finance_admin unfreezes, finance_admin only for cancel). `lib/finance/flag.ts`'s doc-comment corrected — it claimed the flag covers "every money-initiating or state-granting mutation".

**Validation and provisioning.** `spendLimitCents` was `v.optional(v.number())` and accepted negative, fractional, and absurd values — now validated on both entry points. `provisionCard` re-checks fund status, Increase Account, and onboarding immediately before the provider call and audits `card.provision_refused` rather than issuing a card on a fund frozen since creation.

**One addition beyond the brief:** `auditOverLimitCardCharge` in `recordCardSettlement` flags `card.limit_exceeded` when a settlement breaches the mirrored limit. Audit-only, runs last. Cards already provisioned carry limits the bank never received, and this is the only signal anyone gets for them.

## Follow-up this creates

**Cards provisioned before this PR must be re-provisioned or cancelled before real money.** Increase will not apply a limit retroactively. Tracked as a go/no-go row in the runbook (#691). No backfill here — documented as a Known Seam.

## Tests

`finance-cards.test.ts` **28 → 66**. The internalActions `provisionCard`, `applyCardStatus`, and `applyCardLimit` previously had **zero** references in the entire suite; they are now invoked directly against a mocked `lib/finance/increase`, so the payload sent to the provider — the limit and its interval mapping — is asserted rather than assumed. Covers 12 validation-rejection cases, provisioning refusal on a fund frozen after creation, freeze/unfreeze/cancel working with the flag OFF while every other entry point refuses, role checks surviving the flag exemption, and the over-limit audit.

Mobile finance **91 → 98**, including negative assertions that the receipts control and the second-approver claim are gone.

Convex **2737 passing**; mobile finance 98 passing; ESLint clean.

## Explicitly out of scope

No mobile UI for `setCardLimit` (none existed). No merchant-category or country controls, no real-time decisioning — now enumerated in `ARCHITECTURE.md` as "what card controls still don't do", with the rule that no surface may claim a control absent from that list.


lilseyi (commented):
Independent review. I did not write this code. I pulled Increase's live OpenAPI spec (`https://increase.com/openapi.json`) rather than trusting the PR body, and traced the settlement path and the copy claims through to what the backend actually enforces.

**P0 — none.**

---

## First, the thing the PR asks to be checked: the payload is correct

I verified `buildAuthorizationControls` (`lib/finance/increase.ts:379`) against `components.schemas.create_a_card_parameters` and `update_a_card_parameters` in Increase's published spec:

- `authorization_controls.usage` — `category` is **required**, enum `single_use | multi_use`. The code sends `"multi_use"`. Correct.
- `usage.multi_use` — "Required if and only if `category` is `multi_use`". Sent. Correct.
- `multi_use.spending_limits[]` — required fields exactly `["interval", "settlement_amount"]`. Correct, and `merchant_category_codes` is genuinely optional.
- `interval` enum: `all_time | per_transaction | per_day | per_week | per_month`. The map at `cardPolicy.ts:479` uses `per_week` / `per_month` / `per_transaction`. Correct.
- **Units**: `settlement_amount` is `type: integer, minimum: 0`. The spec's own card-domain descriptions confirm minor units — `card_authorization.amount` / `card_settlement.amount`: *"in the minor unit of the transaction's currency. For dollars, for example, this is cents."* So cents, positive, no ×100 error and no sign error. Correct.
- **Reset semantics**: the enum descriptions read *"Resets weekly on Mondays at midnight UTC"* and *"Resets on the first of the month, midnight UTC"*. `LIMIT_HINTS` in `CreateCardView.tsx` now says exactly that. Correct, and this is the rare case of UI copy matching a provider spec word for word — good.
- `update_a_card_parameters` carries the identical `authorization_controls` shape. Correct.

So the core premise of the PR holds. Everything below is about the code around it.

---

## P1

### 1. The wire payload has zero test coverage — the one thing that must not be wrong is the one thing nothing exercises

`finance-cards.test.ts:60` mocks the whole `lib/finance/increase` module. Every limit assertion (`:553`, `:1061`, `:1084`) checks the arguments to the **wrapper** — `{ interval, settlementAmountCents }` — which means `buildAuthorizationControls` never runs in any test in the repo. Grep confirms it: `authorization_controls`, `settlement_amount` and `spending_limits` appear only in `increase.ts` and in comments. There is no `__tests__` file for `lib/finance/increase.ts` at all.

Rename `spending_limits` to `spendingLimits`, drop `usage.category`, or nest one level wrong, and: CI is green, the tests still assert `per_week / 25000`, `createCard` succeeds (Increase ignores unknown-but-well-formed structures or 400s into `card.provision_failed`), and the card ships **uncapped** while `CardDetailView` tells a church treasurer "The bank declines anything over this limit."

That is precisely the failure this PR exists to eliminate, reproduced one layer down. Please export `buildAuthorizationControls` (or stub `fetch`/`increaseRequest`) and assert the literal body:

```json
{"usage":{"category":"multi_use","multi_use":{"spending_limits":[{"interval":"per_week","settlement_amount":25000}]}}}
```

…and the empty-array form for the clear path. Two assertions, and the PR's headline claim becomes load-bearing instead of aspirational.

### 2. `auditOverLimitCardCharge` is not "audit-only" — it can roll back the ledger entry, and it does an unbounded read

`functions/finance/webhooks.ts:217` calls it from **inside** `recordCardSettlement`'s `internalMutation`. The comment at `:216` says:

> Runs last, and only ever writes an audit row, so a bug here can never affect the expense or the ledger entry above.

That is not how Convex mutations work. The mutation is one transaction. If `auditOverLimitCardCharge` throws, the `card_charge` expense insert and the `card_capture` ledger debit above it are **both rolled back**, the webhook returns non-200, and Increase retries into the same failure. A card swipe would silently stop being recorded.

And there is a concrete way to make it throw: `webhooks.ts:259` does

```ts
.withIndex("by_card", q => q.eq("cardId", card._id)).collect()
```

— *every* `card_charge` expense the card has ever had, unbounded, then filters `createdAt >= windowStart` in JS. A card in weekly use for a couple of years will eventually hit Convex's document/bytes read limits inside a money-path mutation. The index only covers `cardId`, so there's no range to narrow on.

Fix either by adding `createdAt` to the `by_card` index and using a bounded range, or by moving the whole check to a scheduled action after the settlement commits. If it stays inline, wrap it in try/catch so the comment becomes true.

Related, smaller: it sums `card_charge` rows by `expenses.createdAt` (our insert time), not the bank's authorization time, so the window can disagree with Increase's around a boundary — and it counts `denied` rows.

### 3. Increase-succeeds / Convex-fails leaves a card the app is wrong about, in the dangerous direction, with nothing to reconcile it

`applyCardLimit` (`cards.ts:665`): `updateCardSpendingLimit(...)` then `recordCardLimit(...)`.

If the provider call throws, both sides keep the old value — consistent, fine. But if Increase **succeeds** and the mutation then fails (or the action is killed in between):

- **clear**: bank uncapped, mirror still shows a cap. `CardDetailView` renders "The bank declines anything over this limit." It doesn't.
- **raise**: bank at the new higher cap, mirror at the old lower one. The finance admin believes the card is tighter than it is.

Nothing reconciles this. `getCardDetail` reads the mirror. There is no `GET /cards/{id}` read-back and no cron. The only signal is `auditOverLimitCardCharge`, which by construction fires only after money has already moved past the cap.

Also, `applyCardLimit` has **no try/catch and no failure record at all** — unlike `provisionCard` (→ `recordCardProvisionFailed`) and `applyCardStatus`. A failed limit change is completely silent.

And the audit row is written on **intent**, not fact: `setCardLimit` logs `card.limit_updated` at `cards.ts:617`, before scheduling anything. If the action never runs or throws, `financeAuditEvents` permanently asserts a limit change that never reached the bank. In a system whose audit trail is the compliance artefact, that's the wrong ordering — the PR gets it right for the mirror and wrong for the audit.

Minimum: add `card.limit_update_failed` on the catch, and move the success audit into `recordCardLimit` (keep a `card.limit_update_requested` at the mutation if you want the actor recorded).

### 4. `setCardLimit` skips the fund-status and card-status gates every sibling mutation applies

`cards.ts:585`. It checks the flag, the role, and `card.increaseCardId` — but never `fund.status` or `card.status`.

- `createFundCard` refuses a non-`active` fund; `provisionCard` re-checks it immediately before the provider call (nice addition, this PR). But `setCardLimit` will happily **raise** a cap on a card whose fund was frozen because its group got archived. Raising spending power on a frozen fund contradicts every other gate in the module.
- A `canceled` card passes too (it has an `increaseCardId`). The PATCH will presumably error at Increase, and per finding 3 that error goes nowhere.

Add `if (fund.status !== "active") throw` and a card-status check, matching `createFundCard`.

### 5. The false second-approver claim was not deleted — it was reworded, moved into a shared constant, and now renders on *two* screens

The PR body says: *"The false second-approver claim is gone."*

`apps/mobile/features/finance/leader/types.ts:157`:

```
CARD_CHARGE_REVIEW_NOTE =
  "Charges settle straight away, then land in the fund's activity for sign-off. Anything over $200 needs a second approver."
```

Rendered at `CreateCardView.tsx:242` **and — newly added by this PR — `CardDetailView.tsx:165`**. Before this PR the claim appeared on the create sheet and the hub; now it appears on the create sheet and the card detail. Net surface unchanged, on a card-issuance screen where a reader takes it as a spending control.

Meanwhile `functions/finance/ARCHITECTURE.md` (unchanged here) still says: *"Approval/payout for `card_charge` kind still isn't wired… a `card_charge` expense today just sits `pending`."* I checked: `approveExpense` (`expenses.ts:249`) does not reject `card_charge`, so the threshold *would* apply if anything called it — but nothing in the mobile app surfaces card charges for approval, and `canPay` (`expensePolicy.ts:213`) correctly requires `kind === "reimbursement"`, so the "sign-off" the copy promises has no UI behind it either. The sentence asserts two controls, and neither is reachable by a user today.

By this PR's own stated rule ("no surface may claim a control absent from that list"), both halves should go — or the sign-off flow should ship.

### 6. The negative test for that copy is a string-literal test that passes while the claim survives

`GivingHubView.test.tsx:150`:

```ts
expect(screen.queryByText(/Charges over \$200 need a second approver/i)).toBeNull();
```

The new copy says *"Anything over $200 needs a second approver."* Different literal, same claim, and the test is green. `CardDetailView.test.tsx:52` and `CreateCardView.test.tsx:83` have the same shape (`queryByText("Receipts required")` / `"Require receipts"`) — those two are fine because the feature really was deleted, but this one is the case the task warns about: a test asserting on a phrasing rather than on the property. Assert `queryByText(/second approver/i)` and `/\$200/`, and it will fail today — correctly.

---

## P2

- **PATCH replace-vs-merge is asserted in a comment, not verified.** `increase.ts:445` PATCHes `authorization_controls` with only `usage`. The spec documents the shape but says nothing about merge semantics. If Increase replaces the whole object (as the comment claims), a `setCardLimit` silently wipes any `merchant_category_code` / `merchant_country` / `merchant_acceptor_identifier` control set out-of-band in the Increase dashboard. Togather sets none today, so this is fine now — but please confirm against the sandbox with a read-back before ops start using the dashboard, and record the answer. The clear-a-limit direction is the safe one either way (a merge would leave the old, tighter limit in place — "the most restrictive limit applies if multiple limits match"), so this is not urgent, just unverified.
- **Existing cards deserve a migration, not a runbook row.** The PR correctly identifies that already-provisioned cards carry limits Increase never received, and defers to "re-provision or cancel before real money." You already have `applyCardLimit` — a one-shot internal action looping active cards with a stored limit would close it in ten lines and remove a manual go/no-go item from #691. Strictly safer than trusting the runbook.
- **`MIN_CARD_LIMIT_CENTS = 1`** (`cardPolicy.ts:495`) rejects `0`, which Increase permits (`minimum: 0`) and which is the legitimate "block all spend on this card" state. Not wrong, but you're declining a useful control; worth a sentence saying it's deliberate.
- **`setCardLimit`'s return value** reports the *requested* limit (`cards.ts:~640`) as if applied. No mobile UI consumes it yet — but when one does, it will render an optimistic value the bank may not have taken. Return the mirror, or return `{ status: "pending" }`.
- **`cardLimitWindowStart` week math** (`cardPolicy.ts:552`) is correct — `(getUTCDay() + 6) % 7` handles Sunday properly, and `Date.UTC` avoids local-time drift. Verified, no action.

---

## What I checked and found sound

- **Flag exemption (item 4) is clean.** `setCardFrozen` gates on `isFundFinanceAdmin` (which wraps `requireFundRole`, community-admin override included) plus a `card.holderUserId === userId` identity check; `cancelCard` gates on `requireFundRole(finance_admin)`. Neither ever relied on `requireGroupGivingEnabled` for authorisation — it was purely a feature gate sitting in front of them, so removing it does not create a hole. The tests at `finance-cards.test.ts:1135-1235` cover it properly and *negatively*: flag off + plain member still refused, flag off + holder still can't self-unfreeze, and every other card entry point (including the new `setCardLimit`) still throws. This is the right shape.
- **`provisionCard`'s re-check of fund status / Increase Account / onboarding immediately before the provider call**, with `card.provision_refused` distinguished from `card.provision_failed`, is a genuine improvement — it closes a real gap where a fund frozen between mutation and action would still get a live card.
- **`validateCardLimit`** correctly rejects the pair-of-one case, non-integers, ≤0 and absurd values, on both entry points.
- **Suite-wide `vi.useFakeTimers()` at `finance-cards.test.ts:37` is not hiding anything**, because the new tests invoke `provisionCard` / `applyCardStatus` / `applyCardLimit` directly via `t.action` rather than relying on the scheduler, and the mutation-side tests explicitly assert the mirror still holds the *old* value before the action runs (`:1036`) — which is a good assertion, not an accident. Getting these three internalActions from zero references to directly-driven is the strongest part of this PR's test work.

---

The single most important thing to change: **assert the actual HTTP body sent to Increase** (P1.1). Every promise this PR makes to a treasurer — and every line of new UI copy — rests on `buildAuthorizationControls` being byte-correct, and it is currently the only code in the change that no test executes. I verified it by hand against the spec today; a test is what keeps it true tomorrow.

PR #693, opened 2026-07-31
Members had no way to give, and "Approved" meant nothing was coming

lilseyi:
> **Stacked on #692** (→ #690 → #689). Review those first.

## 1. A plain member could not reach giving at all (P0)

`app/groups/[group_id]/index.tsx:10` renders `whatsappShell ? <GroupInfoScreen /> : <GroupDetailScreen />`. The member-facing Giving tile existed **only** in `GroupDetailScreen` — the flag-OFF legacy screen. In `GroupInfoScreen` the sole Giving entry point sat inside the `{(isLeader || isAdmin)}` Leader-tools block and routed to the leader hub.

**`whatsapp-shell-on` is enabled in production.** This was never a staging quirk: the day `group-giving` turns on, no ordinary member can give, and the fully-built transparency screen has no door. The donor path — the entire point of the feature — was leader-only.

How it survived: ADR-032 §6 says giving "slots into `GroupDetailScreen`" and never mentions the WhatsApp-shell screen. The ADR is updated here with the placement and an explicit note that **both** screens must carry the member door.

Added a member `Giving` group to `GroupInfoScreen`, outside any leader gate, using the same lightweight `getGivingContext` query (no new backend surface), gated on the flag **and** fund existence. Placed between Channels and Leader tools — the screen runs member band → leader tools → metadata → destructive, and giving is a member capability per ADR-032 §4. Titled "Group fund" rather than "Giving" so it doesn't collide with the Leader-tools "Giving" row a leader sees directly below; the two now read as distinct destinations.

## 2. "Approved" read as "paid", and nothing was coming

`getPayoutDestination` returns `null` unconditionally, so `payReimbursement` always short-circuits to an audit breadcrumb. A member saw a blue **"Approved"** badge and nothing else — which any normal person reads as settled and on its way.

Badge is now **"Awaiting payout"** (the badge itself had to change; a skim-reader only sees the badge), with a sub-line:

> Approved — a treasurer still needs to send you the money. It'll say Paid here once they have.

No date, no "processing", no "on its way". `paid` is untouched and still reads as green "Paid". `labels.ts` carries a comment explaining *why*, so nobody reverts it.

**Known gap this exposes, not fixed here:** `recordReimbursementPaid` is an `internalMutation` with no caller and no UI, so there is currently no action available to *anyone* in the product that moves a reimbursement to Paid. The copy above is honest about the state but the terminal state is unreachable until a "Mark as paid" action exists — and that carries an accounting decision (whether a payout made from outside the fund's own account may be recorded at all). Flagged for a follow-up, deliberately not guessed at here.

~~Also inconsistent after this change: `features/finance/leader/utils.ts` keeps its own copy of the status labels, so a leader still sees "Approved".~~ **Correction — this claim was wrong.** `leader/utils.ts` contains only `financeDisplayName`, EIN helpers, and `formatLedgerKind` (ledger *entry* kinds, a different enum); it has no expense-status labels at all. The only leader-facing expense list is `GivingHubView`'s "Needs your approval", fed by `listExpenses` pre-filtered to `status: "pending"`, and its "All activity" link routes to the same member `FundScreenView` this PR changes. The one "Approved" string on the leader side is a post-action toast (`GivingHubScreen.tsx:129-130`), not a status badge. So there is no leader/member inconsistency to decide on.

## 3. The receipt you just attached was invisible — and it wasn't a finance bug

`utils/media.ts`'s `getMediaUrl` recognised only `http(s)://`, `file://` and `r2:`. On web, `expo-image-picker` returns a `blob:` object URL (or a `data:` URI), so it returned `undefined`, `AppImage` had no URL, and it rendered its gray placeholder — exactly the reported symptom (no `` in the DOM, but Change/Remove still present, since those live outside `AppImage`). Native was fine via the `file://` branch.

Accepting `blob:` and `data:` is one line and repairs **every** just-picked image preview on web — profile photo, group photo, not just receipts.

## Tests

Written first, confirmed failing (11 red). New `ImagePicker.test.tsx`; extended `media.test.ts`, `FundScreenView.test.tsx`, `SubmitReimbursementSheetView.test.tsx`, `GroupInfoScreen.test.tsx` — a plain member sees the member row and it pushes `/groups/1/fund` and never a `leader-tools` path; absent with the flag off; absent with no fund; a leader gets both rows routing to their own surfaces; approved renders the honest copy and matches no ETA pattern.

Full mobile suite: **212 suites, 2058 passed, 9 skipped, 0 failed** (baseline 211 / 2042 / 9). Typecheck: 20 pre-existing errors before and after, identical, none in touched files.

No dependency added — RN primitives and existing components only.


lilseyi (commented):
Independent review. Verified against the PR head worktree and the live `convex` client source, not just the diff.

---

**P0 — the new member Giving row crashes the screen for the exact users it's supposed to serve**

`apps/mobile/features/groups/components/GroupInfoScreen.tsx:223-227` fires `getGivingContext` unconditionally on `group?._id`, *before* the `isMember` check at line 478:

```ts
const givingContext = useAuthenticatedQuery(
  api.functions.finance.giving.getGivingContext,
  group?._id ? { groupId: group._id as Id<"groups"> } : "skip",
);
```

`getGivingContext` (`apps/convex/functions/finance/giving.ts:262-297`) calls `assertCanViewFund` (line 98-120), which **throws** a plain `Error("You don't have access to this fund")` for anyone who isn't an active `groupMembers` row or a community admin. Convex's `useQuery` re-throws server errors synchronously during render — confirmed directly in `node_modules/convex/dist/react.bundle.js:4466-4470`:

```js
const result = results["query"];
if (result instanceof Error) { throw result; }
```

The file's own comment right above the `isMember` branch (line 474-477) says this exact route is reached "from a share link or Explore" by non-members, and renders `GroupNonMemberView` (join/request CTA) for them. So: any group that has `group-giving` on and a fund created will show the app's root `ErrorBoundary` ("sad gif" full-screen error, `apps/mobile/components/ErrorBoundary.tsx`) instead of the join flow, to every non-member who opens it — a core discovery/growth path, not an edge case. This also fires for a member whose local `isMember` cache is stale (e.g. just left/removed).

Compare the two sibling queries added to this same component in earlier work, which fail **open** for exactly this reason: `countGroupJoinRequests` returns `0` for a non-reviewer (`apps/convex/functions/groupMembers.ts:1298-1300`), and `getGroupNotifications` returns a default object regardless of membership (`apps/convex/functions/notifications/preferences.ts:63-76`). `getGivingContext` doesn't follow that convention — it throws.

Fix: either gate the query on `isMember` (`group?._id && isMember ? {...} : "skip"`), or make `getGivingContext` fail-open (return `null`) for a non-member the way `getFundOverview`/`getGivingContext` already do for "flag off"/"no fund", consistent with `functions/finance/ARCHITECTURE.md:131`'s stated rule that member-facing reads return null instead of throwing.

**No other P0s** — the row's actual authorization (`assertCanViewFund`) is correctly scoped when it doesn't throw; I found no case where the giving entry point renders for someone who shouldn't see it, and no donor-identifying data leak.

---

**P1 — the new GroupInfoScreen tests can't catch the P0 above; they mock past it**

`apps/mobile/features/groups/components/__tests__/GroupInfoScreen.test.tsx`'s new `describe("member Giving entry point (ADR-032 §4)")` block mocks `useAuthenticatedQuery`'s return value directly (`setUpGiving`) and every scenario calls `renderAs()` with `isGroupMember` stubbed to `true`. There is no test that renders this component as a genuine non-member with a real (or realistically-thrown) `getGivingContext` result — the four assertions all validate the gating *logic* (flag off, no fund, leader vs member routing) against a friendly mocked value, not the integration behavior that actually breaks. This is exactly the "assert on mocks" pattern to watch for: this suite would stay green with the P0 bug present.

---

**P1 — `getMediaUrl` accepting `data:` is scoped to "local picker output" in the PR's reasoning, but the function is also fed attacker-controlled server data**

`apps/mobile/utils/media.ts`'s fix is correct for its stated purpose (web `expo-image-picker` blob/data URIs), but `getMediaUrl` is a shared, general-purpose resolver called from `MessageItem`, `ImageAttachmentsGrid`, `AudioPlayer`, `VideoPlayer`, `FileAttachment`, etc. — i.e. on `chatMessages.attachments[].url`, which is `v.string()` with **no format or size validation** in the schema (`apps/convex/schema.ts:1897-1910`) or in `sendMessage` (`apps/convex/functions/messaging/messages.ts:1251-1269`). Any authenticated group member can call `sendMessage` with `attachments: [{ type: "image", url: "data:image/png;base64,<huge blob>" }]` directly (no special privilege, doesn't require going through the picker UI). Before this PR that fell to `getMediaUrl`'s "unrecognized format" branch → `undefined` → gray placeholder. After this PR it's rendered as-is for every other subscriber of that channel, at full embedded size (`getMediaUrlWithTransform` correctly skips CDN transforms for non-R2 URLs, so there's no resizing either) — a real perf/memory footgun with no CDN caching, reachable by any chat member, new as of this change. Not exploitable as script-executing XSS (native `<Image>` / RN-Web `` both sandbox SVG loaded via ``/image context), but worth a size cap or restricting the `data:`/`blob:` branch to a "local, not-yet-uploaded" code path rather than the general-purpose resolver every stored attachment goes through.

---

**P1 — the PR's own stated leader/member copy inconsistency doesn't exist in the code as described**

The PR body says: "inconsistent after this change: `features/finance/leader/utils.ts` keeps its own copy of the status labels, so a leader still sees 'Approved'." Checked: `apps/mobile/features/finance/leader/utils.ts` has no expense-status labels at all — only `financeDisplayName`, EIN helpers, and `formatLedgerKind` (a different enum, for ledger *entry* kinds like "donation"/"transfer", not expense status). Grepping the entire `leader/` surface, the only place expense status renders is `GivingHubView.tsx`'s "Needs your approval" list (`apps/mobile/features/finance/leader/GivingHubView.tsx:293-296`), which is pre-filtered to `status === "pending"` only, so it never shows "Approved" either way. The hub's own "All activity" link (`GivingHubView.tsx:281`) even routes to `/groups/${groupId}/fund` — the *same* member `FundScreenView` this PR just changed — so a leader who clicks through today sees the new "Awaiting payout" copy, not "Approved." There is currently no leader-facing surface showing a bare "Approved" badge. Not a functional bug, but the PR's stated justification for leaving a "leader/member inconsistency" undecided is based on a surface that doesn't exist — worth correcting before this gets cited as settled reasoning later.

---

**P2 — "treasurer" isn't a role in the permission model**

`expenseStatusNote` (`apps/mobile/features/finance/member/labels.ts:67`): "a treasurer still needs to send you the money." ADR-032 §4's role table is Member / Cardholder / Manager / Finance admin / Community admin — no "treasurer." Harmless as plain-language copy, but slightly inconsistent with the rest of the product's terminology; consider "a fund manager" or similar if this is meant to map to an actual grantable role.

**P2 — no other clarity issues found.** The `paid`/`pending`/`denied` badge variants are unchanged and remain visually distinct; the "no ETA" test constraint matches the shipped copy; `getMediaUrlWithTransform`'s local-URI passthrough is correct and tested accurately.

---

Most important thing to change: gate `getGivingContext` in `GroupInfoScreen.tsx` on `isMember` (or make the query fail-open like its sibling reads) before merging — as written, turning on `group-giving` for any community with a group fund will crash the join/share-link flow for every non-member who opens that group, and the new tests don't exercise that path so CI won't catch it.

PR #700, opened 2026-07-31
feat(agents): completion gates, codeowner carve-outs, and conductor config

lilseyi:
Week-1 baseline of the agent-workflow upgrade. Four small pieces, one theme: make agent loops stop on evidence, and put a human in front of the changes CI cannot judge.

> **Revised after review** ([verdict](https://github.com/togathernyc/togather/pull/700#issuecomment-5148148109)). The gate's original design blocked on this repo's pre-existing tsc baseline and switched itself off via `stop_hook_active`. Both are fixed in `f01b2ab2`; details per finding are on the inline threads.

## 1. Stop hook — deterministic completion gate

`.claude/hooks/completion-gate.sh`, registered as a `Stop` hook in `.claude/settings.json`. When a session tries to stop, it typechecks each affected workspace and **exits 2** if a file *the session touched* has a type error — which blocks the stop and feeds that error back to the agent.

Two invariants keep it honest:

- **Only errors in changed files count.** This repo has ~20 pre-existing `apps/mobile` tsc errors and nothing in CI runs tsc, so an unfiltered check would report someone else's bug as yours on every stop. tsc runs with `--pretty false` and the output is filtered to errors whose file has uncommitted changes.
- **Only a real type error can block.** The exit status is ignored entirely; the sole blocking signal is a matching `error TS` line. Missing deps, offline `npx`, OOM, a bad tsconfig, a timeout, a conflicted merge, or no matching workspace all fail open.

There is no root-tsconfig fallback — the root tsconfig sets no `include`/`exclude`, so tsc walked the whole monorepo (2,933 errors in 59s, under the timeout, so it blocked rather than timing out safely).

**Recursion guard:** `stop_hook_active` is deliberately *not* consulted. `/goal` registers its own Stop hook, and once that blocks, the flag stays true for the rest of the session — which would have disabled the gate in exactly the unattended loops it protects. Instead the hook counts its own consecutive blocks in a session-keyed state file, stands down after 2, and resets the counter whenever a check passes.

Verified end-to-end against a harness that stubs `npx` for deterministic tsc output: error in a touched file blocks; errors only in untouched files pass; first and second blocks fire and the third stands down; a passing check resets the budget; a tsc failure with no `error TS` lines fails open; an error in the *second* affected workspace is still caught; mid-merge, no-TS-changes, and changed-TS-outside-any-workspace all pass.

## 2. `/goal` conditions in the loop commands

`auto-worker.md` and `review-cycle.md` now open by telling the agent to run `/goal` with a definition of done stated as transcript-checkable evidence plus a cap. Each cap is stated in the unit that file already budgets in — cycles for review-cycle, turns for auto-worker — with a note on why they are not interchangeable.

## 3. CODEOWNERS carve-outs

`.github/CODEOWNERS` already existed (defaults to `@togathernyc`), so this appends a human-only-review section. `@lilseyi` now owns the lockfile, `apps/mobile/package.json`, `native-deps.json`, the auth directory *and* the central auth modules outside it (`auth.ts`, `lib/auth.ts`, `authInternal.ts`), the billing/finance paths, and every migrations directory and `migrations.ts` module. Every path is verified to exist.

The rationale comment points at ADR-013: a JS-only dependency change can re-key the Expo native module graph and blank out native video and GIF on a real device while typecheck, tests, and the bundle all stay green.

**This is reviewer routing and documentation, not a merge gate** — see the follow-ups.

## 4. Conductor per-project config

`.conductor/settings.toml` — verified as the current format. Conductor worktrees carry only tracked files, so `.env.local` silently vanished and the app came up pointed at no Convex deployment.

- `file_include_globs` copies the root and `apps/mobile/.env.local`. It deliberately does **not** use `apps/*/.env.local`, which would sweep in `apps/convex/.env.local` and point every workspace at the same backend.
- `setup = "pnpm install --frozen-lockfile"` — a bare install re-resolves and can re-key the native graph, in a fresh worktree where full re-resolution is guaranteed.
- Metro runs on `$CONDUCTOR_PORT`, not via `pnpm dev --mobile`: `scripts/dev.js` pins 8081 and *kills* whatever holds it, so the second workspace to start would shoot the first one's Metro.

## Also in this PR

`.claude/hooks.json` is deleted. Claude Code reads project hooks from `settings.json` only, so that file — and the `ralph-logger.sh` Stop hook it registered — had never run. Rather than delete the logger too, it is folded into the new `Stop` array *before* the completion gate, so logging still happens when the gate blocks. Both entries are anchored to `$CLAUDE_PROJECT_DIR` so they survive `cd apps/mobile && claude`.

## Follow-ups

- **Do NOT enable "Require review from Code Owners."** @lilseyi is the sole human owner and GitHub does not count a PR author's own review, so the toggle would make his own lockfile bumps and migrations unmergeable without an admin override. The real gate for these paths is that only lilseyi merges them, which the `merge_pull_request` / `enable_pr_auto_merge` deny rules in `.claude/settings.json` already enforce from the agent side. Revisit only if a second human joins.
- **Fix the ~20 pre-existing `apps/mobile` tsc errors and add a `typecheck` task to CI.** The gate works around this baseline by scoping to changed files; that is the cheap fix, not the honest one. Until tsc runs in CI the baseline can keep rotting.

## Validation

`bash -n` on the hook, `python3 -m json.tool` on `settings.json`, `tomllib` parse on `settings.toml`, plus the gate harness described above. Husky pre-commit/pre-push were bypassed because this worktree has no `node_modules` (`lint-staged` not found); the change is config and markdown only.

### Live verification caught a bug the harness could not

The gate was run against the real repo — real `node_modules`, real uncommitted TypeScript across `apps/mobile` and `apps/convex` — and that run found a defect every stubbed test had passed over.

The changed-file list was being interpolated into awk via `-v`. Any change touching two files puts a newline in that string, which awk rejects outright (`awk: newline in string ... at source line 1`), so the filter errored on every invocation and matched nothing. Fail-open held the exit code at 0, which is precisely why it looked healthy: the gate was not passing, it was **inert**, and could never have blocked anything.

Fixed in `69353519` — the list is now a temp file read as awk's first input (`NR==FNR`) rather than program text, membership is an exact hash lookup so regex metacharacters in filenames stay literal, and the error-line pattern requires `(line,col): error TS` so parentheses inside a filename cannot truncate the path.

The harness missed it because every case changed exactly **one** file per workspace, so no list ever contained a newline. Every case now dirties several files, includes filenames with spaces and parentheses, and asserts stderr is free of awk errors — that last assertion is what separates "allowed the stop" from "could not run". Replayed against the pre-fix script, 8 of 12 cases fail, two of them on the stderr assertion alone.

Worth recording plainly: a green stub suite said this worked, and it did not. The exit codes were right for the wrong reason.


.claude/hooks/completion-gate.sh
```diff
@@ -0,0 +1,59 @@
...
+CHANGED=$(git status --porcelain -- '*.ts' '*.tsx' 2>/dev/null | awk '{print $NF}')
+[ -z "$CHANGED" ] && exit 0
+
+if node -e 'process.exit(require("./package.json").scripts.typecheck?0:1)' 2>/dev/null; then
+  CHECK_DIR="$REPO_ROOT"
+  CHECK_CMD=(pnpm typecheck)
+  LABEL="pnpm typecheck"
+else
+  # No root script today: fall back to tsc in the first affected workspace.
+  WORKSPACE=$(printf '%s\n' "$CHANGED" | grep -oE '^(apps|packages)/[^/]+' | head -1)
```
lilseyi: **critical — the gate fires on the repo's pre-existing baseline, not on the change.**

`npx tsc --noEmit` in `apps/mobile` does **not** pass on this repo today. I ran it:

```
EXIT=2
20 x "error TS"
```

across 15 files, none of which are files I have modified locally — e.g. `hooks/useDevToolsEscapeHatch.ts(91,9): error TS2322: Type 'number' is not assignable to type 'Timeout'.`, plus errors in `features/chat/components/AudioPlayer.tsx`, `MessageInput.tsx`, `VoiceRecorderBar.tsx`, and eight `__tests__` files. There is no `typecheck` script in any workspace and nothing in `ci.yml` or `turbo.json` runs `tsc`, so this baseline has never been enforced and is not clean.

Consequence: any session that stops with an uncommitted `apps/mobile` `.ts`/`.tsx` (and no earlier-sorting workspace in the change set) gets **exit 2** with a reason that has nothing to do with what the agent did. That breaks the contract stated in the header comment — exit 2 is supposed to mean "your uncommitted TypeScript does not compile."

Options: (a) diff the error set against a cached baseline and only block on *new* errors; (b) filter `$LOG` to errors whose file path is in `$CHANGED` and exit 0 if none match; (c) fix the 20 baseline errors first and add a `typecheck` task to CI so this can't silently rot again. (c) is the honest fix, (b) is the cheap one.

**high (same line) — only the first affected workspace is checked.** `head -1` means a change spanning `apps/convex` *and* `apps/mobile` typechecks convex only and lets mobile through. Given `apps/convex` sorts first and currently passes cleanly, that's also why this may have looked green in manual testing. Loop over the unique workspaces instead.

.claude/hooks/completion-gate.sh
```diff
@@ -0,0 +1,59 @@
...
+  CHECK_DIR="$REPO_ROOT/${WORKSPACE:-.}"
+  [ -f "$CHECK_DIR/tsconfig.json" ] || exit 0
```
lilseyi: **critical — the `WORKSPACE` empty fallback compiles the entire monorepo and always fails.**

When no changed TS is under `apps/` or `packages/`, `CHECK_DIR` becomes the repo root, and the root `tsconfig.json` is:

```json
{ "compilerOptions": {}, "extends": "expo/tsconfig.base" }
```

No `include`, no `files`, no `exclude` — so tsc walks every `.ts`/`.tsx` in the repo. I measured it:

```
ROOT TSC EXIT=2
elapsed 59s
error count: 2933
apps/convex/__tests__/admin-broadcasts.test.ts(193,9): error TS2578: Unused '@ts-expect-error' directive.
```

2,933 errors in 59s — under the 120s cap, so it does **not** time out into the safe exit-0 path. It blocks the stop and reports an unrelated `@ts-expect-error` in a convex test.

This path is reachable: there are 17 tracked TS files outside `apps/`/`packages/` — `scripts/*.ts` (11 of them), `playwright.config.ts`, `ee/scripts/sync-staging-data.ts`, `ee/billing/**/*.tsx`, `tests/e2e/*.spec.ts`. Editing any of those alone triggers it. It's also reachable via the `awk '{print $NF}'` parse on a path containing a space (`?? "a b.ts"` yields `b.ts"`), which fails the `^(apps|packages)/` match.

Simplest fix: drop the root fallback entirely — `[ -z "$WORKSPACE" ] && exit 0`. The root tsconfig is not a checkable project.

.claude/hooks/completion-gate.sh
```diff
@@ -0,0 +1,59 @@
...
+set -uo pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel 2>/dev/null)" || exit 0
+cd "$REPO_ROOT" || exit 0
+
+INPUT="$(cat)"
+# Claude re-runs Stop hooks after a block; this is what stops a failing
+# typecheck from looping the session forever.
+case "$INPUT" in
+  *'"stop_hook_active":true'* | *'"stop_hook_active": true'*) exit 0 ;;
```
lilseyi: **high — this guard is tripped by the `/goal` command the other half of this PR introduces, which disables the gate for exactly the loops it targets.**

I read the Claude Code 2.1.220 binary to check the interaction. `/goal` is implemented as a **session-scoped Stop hook of type `"prompt"`**:

```js
t.sessionHooksRegistry.add(n,"Stop","",{type:"prompt",prompt:e})
```

and when *any* Stop hook blocks, the query loop continues with:

```js
g={messages:[...,...Un.blockingErrors], stopHookActive:!0, stopHookBlockingCount:_o,
   transition:{reason:"stop_hook_blocking"}}; continue
```

`stopHookActive` is then threaded forward (`stopHookActive:te`) through every subsequent turn until a new user turn resets it. So the moment a `/goal` blocks a stop — which is the entire point of `/goal`, and both `auto-worker.md` and `review-cycle.md` now tell the agent to set one — `stop_hook_active` is `true` for the rest of the run and this hook returns 0 without ever running tsc.

Net effect: in an unattended auto-worker or review-cycle loop, the completion gate runs at most once and then silently switches itself off. The two headline features of this PR cancel each other.

Also good news from the same read: `/goal clear` only removes `{type:"prompt"}` Stop hooks with an empty matcher, so it will **not** clobber this `{type:"command"}` hook. And `CLAUDE_CODE_STOP_HOOK_BLOCK_CAP` bounds runaway blocking, so a `stop_hook_active` guard is not the only thing standing between you and a wedged session.

Suggested shape: instead of returning early on `stop_hook_active`, keep a marker file keyed on the first error line; exit 0 if the *same* error was already reported once. That preserves the anti-loop property without handing the off switch to `/goal`.

**normal (same line):** the guard is a raw substring match on the JSON. It covers `"stop_hook_active":true` and `"stop_hook_active": true`, but nothing else — a pretty-printed payload (`"stop_hook_active" : true`) slips through, and that is the failure direction that loops. `python3 -c 'import json,sys;...'` or `jq` would be exact, and both are already assumed present elsewhere in this repo's tooling.

.claude/hooks/completion-gate.sh
```diff
@@ -0,0 +1,59 @@
...
+done
+if kill -0 "$CHECK_PID" 2>/dev/null; then
+  kill -9 "$CHECK_PID" 2>/dev/null
+  exit 0 # too slow to gate on
+fi
+wait "$CHECK_PID"
+STATUS=$?
+[ $STATUS -eq 0 ] && exit 0
+
+FIRST_ERROR=$(grep -m1 "error TS" "$LOG" | cut -c1-200)
```
lilseyi: **high — a non-zero exit that is not a typecheck failure still blocks the stop.**

Line 55 treats *any* non-zero `STATUS` as "TypeScript is broken." But `npx tsc --noEmit` exits non-zero for plenty of reasons that aren't type errors: no `node_modules` yet (npx then tries to fetch typescript from the registry, and fails offline), a malformed `tsconfig.json`, `TS18003: No inputs were found`, an OOM kill. In all of those `grep -m1 "error TS"` finds nothing, `FIRST_ERROR` is empty, and the agent is blocked with the content-free `Run it locally for the full output.`

That contradicts the header comment's own policy — "Every 'can't run the check' path is exit 0 on purpose." A check that couldn't run is not evidence of broken code.

```suggestion
FIRST_ERROR=$(grep -m1 "error TS" "$LOG" | cut -c1-200)
# No "error TS" line means the check did not run (missing deps, bad tsconfig,
# OOM), not that the code is broken. Don't gate on it.
[ -z "$FIRST_ERROR" ] && exit 0
```

This matters concretely for the Conductor case this same PR adds: a fresh worktree runs `setup` before anything else, but if setup fails or is skipped, the first Stop in that workspace hits exactly this path.

**normal:** a `git rebase`/merge in a conflicted state shows `UU foo.ts` in `git status --porcelain`, which counts as changed TS. A session that stops mid-conflict gets blocked by conflict-marker syntax errors. Worth an early exit when `git rev-parse -q --verify REBASE_HEAD` or `MERGE_HEAD` resolves.

.claude/hooks/completion-gate.sh
```diff
@@ -0,0 +1,59 @@
...
+(cd "$CHECK_DIR" && "${CHECK_CMD[@]}") >"$LOG" 2>&1 &
+CHECK_PID=$!
+
+# Portable 120s cap (macOS has no coreutils `timeout`).
+for _ in $(seq 120); do
+  kill -0 "$CHECK_PID" 2>/dev/null || break
+  sleep 1
+done
+if kill -0 "$CHECK_PID" 2>/dev/null; then
+  kill -9 "$CHECK_PID" 2>/dev/null
```
lilseyi: **normal — `kill -9 "$CHECK_PID"` kills the subshell, not tsc.**

`CHECK_PID` is the `( cd ... && ... )` subshell; the actual `node`/`tsc` process is its child and survives, holding CPU and the (now unlinked) log fd. On a machine running several Conductor workspaces this accumulates. `kill -9 -- -$CHECK_PID` after `set -m`, or `pkill -P "$CHECK_PID"` before the `kill -9`, cleans up the tree.

**Verified, no action needed:** I did check the timeout mechanism itself on darwin, since bash 3.2 zombie-reaping was the obvious way for this to be quietly broken. It's correct — bash reaps the background child asynchronously, so `kill -0` fails once it exits and the loop breaks, and `wait` still returns the saved status:

```
$ /bin/bash -c '(sleep 1; exit 7) & P=$!; sleep 3; kill -0 $P 2>/dev/null && echo STILL || echo gone; wait $P; echo status=$?'
gone
status=7
```

(GNU bash 3.2.57, arm64-apple-darwin25.) The 120s cap and the 140s `timeout` in `settings.json` are consistent, and measured worst case is 59s at root / 35s in `apps/mobile`.

.claude/settings.json
```diff
@@ -2,6 +2,19 @@
   "enabledPlugins": {
     "ralph-loop@claude-plugins-official": true
   },
+  "hooks": {
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": ".claude/hooks/completion-gate.sh",
```
lilseyi: **normal — use `$CLAUDE_PROJECT_DIR` rather than a relative path.**

```suggestion
            "command": "\"$CLAUDE_PROJECT_DIR\"/.claude/hooks/completion-gate.sh",
```

A bare `.claude/hooks/completion-gate.sh` resolves against the process cwd. Start Claude from a subdirectory (`cd apps/mobile && claude`) and the command is not found; that exits 127, which is neither 0 nor 2, so the hook silently does nothing and the stop proceeds. It fails open, which is the right direction, but it also means the gate is quietly absent in a common workflow. `$CLAUDE_PROJECT_DIR` is the documented way to anchor hook paths.

**Verified, no action needed:** the merge into this file is purely additive — I diffed against `origin/main` and every `permissions.allow` entry and all four `permissions.deny` rules (`mcp__github__merge_pull_request`, `enable_pr_auto_merge`, `create_repository`, `delete_file`) plus `enabledPlugins` are intact and unchanged.

.github/CODEOWNERS
```diff
@@ -16,3 +16,44 @@
...
+# - Billing / finance: moves real money. Wrong is expensive and hard to unwind.
+# - Migrations: run once against production data. There is no second attempt.
+#
+# NOTE: last matching pattern wins in CODEOWNERS, so these override the
+# defaults above. Enforcement additionally requires "Require review from Code
+# Owners" on the branch protection rule for main.
+# ---------------------------------------------------------------------------
+
+# Dependency graph / native safety (ADR-013)
+/pnpm-lock.yaml @lilseyi
```
lilseyi, line 54: **high — with the follow-up branch-protection toggle enabled, this makes @lilseyi's own PRs unmergeable on exactly these paths.**

The file's own note is right that last-match-wins means these override the defaults — but that cuts both ways: it removes `@togathernyc` as an owner of these paths, leaving `@lilseyi` as the *sole* owner. GitHub does not count the PR author's own review toward a Code Owners requirement. So once "Require review from Code Owners" is on, a lockfile bump, a `native-deps.json` edit, or a migration authored by @lilseyi — the primary author on this repo — can only merge via admin override, which defeats the point.

This PR is safe as-is because it doesn't flip the toggle, but the follow-up is what makes the config live, so it should be resolved before then. Either list a second human alongside (`/pnpm-lock.yaml @lilseyi @someone-else`) or use a team that has more than one member.

**Verified, no action needed:** all eleven appended paths exist on disk (`pnpm-lock.yaml`, `apps/mobile/package.json`, `apps/mobile/native-deps.json`, `apps/convex/functions/auth/`, `functions/billing.ts`, `functions/ee/billing.ts`, `functions/finance/`, `lib/finance/`, `apps/convex/migrations/`, `functions/migrations/`, `functions/migrations.ts`); the glob syntax is valid; the block is appended after the defaults so the carve-outs win for their paths; and `* @togathernyc` is untouched and still applies everywhere else. The `migrations.ts` sibling note is correct — a trailing-slash pattern does not match the file.

.conductor/settings.toml
```diff
@@ -0,0 +1,42 @@
...
+#
+# Matches the .env.local entries in .gitignore (root + apps/mobile).
+file_include_globs = """
+.env.local
+apps/*/.env.local
+"""
+
+[scripts]
+# Worktrees start with no node_modules.
+setup = "pnpm install"
```
lilseyi: **normal — `pnpm install` here can rewrite the lockfile this same PR just put under human-only review.**

```suggestion
setup = "pnpm install --frozen-lockfile"
```

Without `--frozen-lockfile`, pnpm will re-resolve and may write a modified `pnpm-lock.yaml` into the worktree. That is precisely the mutation `CLAUDE.md` warns about:

> **`pnpm install` can itself trigger a false second-React** even when adding a totally unrelated devDependency … a full workspace-root `pnpm install` re-resolution can non-deterministically re-key `expo`/`expo-modules-core`/`react-native` etc. onto a second React version

So the setup script runs the exact command the repo tells agents not to run bare, in an environment (fresh worktree, no `node_modules`) where a full re-resolution is guaranteed, and it does it *before* the agent has done anything — meaning a dirty lockfile shows up attributed to the agent's work. `--frozen-lockfile` installs from the committed lockfile and hard-fails if it's out of date, which is what you want in a throwaway worktree.

It's a little ironic next to the `/pnpm-lock.yaml @lilseyi` line added in this same PR.

lilseyi (commented):
Inline findings from the review pass. Summary verdict posted separately as an issue comment.

.conductor/settings.toml
```diff
@@ -0,0 +1,42 @@
+"$schema" = "https://conductor.build/schemas/settings.repo.schema.json"
+
+# Conductor runs each agent in its own git worktree. Worktrees only carry
+# tracked files, so every .env.local silently disappears and the app comes up
+# pointed at no Convex deployment. These globs copy them into new workspaces.
+#
+# Matches the .env.local entries in .gitignore (root + apps/mobile).
+file_include_globs = """
+.env.local
+apps/*/.env.local
```
lilseyi: **normal — `apps/*/.env.local` also copies `apps/convex/.env.local`, which is what pins the backend.**

`apps/convex/.env.local` exists in this repo and is gitignored via `.env*.local`. It carries `CONVEX_DEPLOYMENT`. Copying it means every new Conductor workspace comes up pointed at the *same* Convex deployment by default — which is the exact failure the comment 20 lines below warns about:

> NOTE: parallel workspaces must not share a Convex backend — they will fight over the same data.

The config actively causes that condition and then documents it as something the human must remember to undo. Worth either narrowing the glob and letting `dev:agentN` provision the backend, or at minimum moving that warning up here next to the glob that creates the problem.

Separately, this repo's history has a related trap on record: the Convex CLI, run from a worktree, has rewritten the *main* checkout's `.env.local`. Copying the file in makes that more likely, not less.

**nit:** the comment on line 7 says "root + apps/mobile", but the glob also matches `apps/convex` (and `.gitignore` covers it at line 226, `.env*.local`, not just the two explicit entries). The glob is right; the comment is stale.

**Verified, no action needed:** the file parses with `tomllib`, validates against the shape of the published `settings.repo.schema.json` (all four top-level keys are in `properties`, and `additionalProperties: false` is satisfied), `file_include_globs` is correctly a newline-delimited string — the settings reference documents it as "Files to copy patterns used when `.worktreeinclude` is not present", and this repo has no `.worktreeinclude`, so it is the active mechanism. The `scripts/dev.js` rationale also checks out: `getMetroPort(0)` returns `8081` and `killProcessOnPort(metroPort, 'Metro')` kills the holder, so avoiding `pnpm dev --mobile` is correct.

**nit:** the comment recommends overriding in `.conductor/settings.local.toml`, but that path is not in `.gitignore` (unlike `.claude/settings.local.json`, which is, at line 4). Worth adding in this PR so per-agent overrides don't get committed.

.claude/commands/review-cycle.md
```diff
@@ -17,6 +17,18 @@ Example:
 
 You are a PR Review Cycle agent. Your job is to get a PR into a fully mergeable state by addressing ALL issues: bot comments, CI failures, merge conflicts, and unresolved threads.
 
+### First: Set the Goal
+
+**Before Phase 1, run the `/goal` command** so the loop has a definition of done
+it can verify from the transcript rather than from your own optimism:
+
+```
+/goal `gh pr view` shows the PR mergeable with all checks passing and all review threads resolved (show the output) — or stop after 20 turns
```
lilseyi: **normal — "20 turns" collides with the file's existing "Max total cycles: 20", same number, different unit.**

Six lines below, the Configuration block already says:

```
- **Max poll attempts:** 30 (about 15 minutes max wait for bot review)
- **Max total cycles:** 20 (to prevent infinite loops)
```

A *cycle* here is a full fetch → fix conflicts → fix CI → process comments → wait-for-CI pass. A *turn* is one model response. Phase 3.5 alone ("Poll every 30 seconds until bot checks are done", max 30 polls) and Phase 8.3 ("Max wait time: 15 minutes, 30 polls × 30 seconds") can each burn 30 turns without completing a single cycle. So the `/goal` cap will abort mid-first-cycle while the file's own budget says it has 19 cycles left, and the matching number makes them look reconciled when they aren't.

Either state the cap in the same unit the file already uses ("or stop after 20 cycles") and let `/goal` count what it counts, or pick a turn number that reflects the documented poll budgets. Same issue in `auto-worker.md`: its Usage block sets `--max-iterations 50` while the new goal caps at 25 turns across nine phases, one of which (Phase 6) invokes this very 20-cycle review loop.

.claude/commands/auto-worker.md
```diff
@@ -16,6 +16,21 @@ Or run a single task cycle manually:
...
+**Before Phase 0, run the `/goal` command.** The loop needs a definition of done
+it can check against the transcript, otherwise it keeps going on vibes.
+
+```
+/goal The current backlog task's PR is open, CI checks are green (show `gh pr checks` output), and the review agent's findings are addressed or explicitly deferred — or stop after 25 turns
+```
+
+State it in terms of evidence you can point at in the transcript — command
+output, a PR number, a screenshot path — not "the feature works". The turn cap
+is a floor, not a target: if the task is done in 5 turns, stop at 5.
```
lilseyi: **nit — "a floor, not a target" is backwards.**

A cap you may stop short of is a *ceiling*. As written it reads as an instruction to reach at least 25 turns, which is the opposite of the intent made explicit in the second half of the sentence. These files are prompts an unattended agent follows literally, so the wording matters more than it would in prose.

```suggestion
is a ceiling, not a target: if the task is done in 5 turns, stop at 5.
```

(Same phrasing appears in the PR description.)

.claude/hooks/completion-gate.sh
lilseyi: Fixed in f01b2ab2 — took option (b), the filtered one.

The gate now only blocks on an `error TS` line whose file has uncommitted changes. tsc runs per-workspace with `--pretty false` (stable `path(line,col):` shape), and an awk pass keeps only errors whose file is in the changed set for that workspace. Your 20 apps/mobile baseline errors are structurally invisible to it now — they are in files nobody touched.

I did not take (c). Fixing the 20 baseline errors and adding a `typecheck` CI task is the honest fix and I agree it should happen, but it is a different PR with a different blast radius, and doing it here would bury the hook change. Worth filing.

The `head -1` half is fixed too — it loops over every affected workspace and stops at the first that has a blocking error. Harness case E covers exactly your scenario (clean apps/mobile, broken apps/web) and now returns 2.

.claude/hooks/completion-gate.sh
lilseyi: Fixed in f01b2ab2 — took your suggestion and dropped the root fallback outright.

`[ -z "$WORKSPACES" ] && exit 0`. The root tsconfig is not a checkable project, so there is nothing sensible to fall back to. Editing `scripts/*.ts` or `playwright.config.ts` alone now exits 0 rather than compiling the monorepo (harness case I).

Your note about the `awk {print $NF}` parse breaking on paths with spaces was right and I removed that parser entirely — it is now `git diff --name-only -z HEAD` plus `git ls-files --others -z`, split on NUL. That also fixes rename handling for free, since `--name-only` reports the new path.

.claude/hooks/completion-gate.sh
lilseyi: Fixed in f01b2ab2. Thank you for reading the binary — this was the finding that actually changed the design.

`stop_hook_active` is no longer consulted at all. The hook now keeps its own counter instead: session id parsed out of the stdin JSON, state file at `/tmp/claude-completion-gate-<session>`, at most 2 blocks per session, then it prints the error and stands down with exit 0. A passing check deletes the state file, so the budget resets whenever the code is actually fixed.

I went with a block count rather than your keyed-on-first-error-line suggestion. Same anti-loop property, but it also bounds the case where the agent fixes error A and immediately trips error B — keying on the error line would let that ping-pong indefinitely.

That also moots the substring-matching normal: there is no `stop_hook_active` parse left. The session id is read with `python3 -c json.load`, not a substring match.

.claude/hooks/completion-gate.sh
lilseyi: Fixed in f01b2ab2, and more thoroughly than the suggestion.

Rather than checking `STATUS` and then special-casing an empty `FIRST_ERROR`, the exit status is now ignored entirely. The only thing that can block is the presence of an `error TS` line whose file we touched. That makes fail-open the structural default: missing `node_modules`, offline npx, `TS18003`, a malformed tsconfig, an OOM kill — none of them produce a matching error line, so none of them block. Harness case D covers it.

Also added `npx --no-install` so a missing typescript fails fast instead of trying to fetch from the registry, which matters for exactly the fresh-Conductor-worktree case you flagged.

The conflicted-tree normal is fixed too: early exit 0 if `MERGE_HEAD` or `REBASE_HEAD` resolves (harness case G).

.claude/hooks/completion-gate.sh
lilseyi: Fixed in f01b2ab2 — `pkill -P "$CHECK_PID"` before the `kill -9`, so the tsc child goes down with the subshell rather than being orphaned.

The timeout is also now a single deadline across all workspaces rather than a per-invocation counter, since the loop can run several. Same 120s budget, checked before starting each workspace and inside each wait loop.

Noted on the darwin verification — thanks for checking it rather than assuming. The `wait` after the loop still relies on that behaviour and it survived the rewrite.

.claude/settings.json
lilseyi: Fixed in f01b2ab2 — took the suggestion verbatim, and applied it to the ralph-logger entry too:

```json
"command": "\"$CLAUDE_PROJECT_DIR\"/.claude/hooks/completion-gate.sh"
```

You are right that 127 fails open, which is the safe direction, but a gate that is silently absent whenever someone runs `cd apps/mobile && claude` is not much of a gate.

Thanks for diffing the permissions merge.

.github/CODEOWNERS
lilseyi: Fixed in f01b2ab2, by changing the plan rather than the owners.

The follow-up now explicitly says **do not** enable "Require review from Code Owners". You are right that it deadlocks: @lilseyi is the only human here, GitHub ignores author self-review, and there is no second human to add — so listing one would be fiction.

CODEOWNERS stays as reviewer routing and as documentation of which paths carry CI-invisible failure modes. The actual gate for these paths is that only lilseyi merges, which the `mcp__github__merge_pull_request` / `enable_pr_auto_merge` deny rules in `.claude/settings.json` already enforce from the agent side. The header comment now says this outright so nobody re-derives the toggle idea from the file and walks into the deadlock.

The PR body follow-up list is updated to match.

.conductor/settings.toml
lilseyi: Fixed in f01b2ab2 — took the suggestion, with the reasoning inlined as a comment so the next person does not "simplify" it back:

```toml
setup = "pnpm install --frozen-lockfile"
```

The irony was load-bearing, agreed. A fresh worktree with no `node_modules` is the guaranteed-full-re-resolution case CLAUDE.md warns about, and it would have run before the agent touched anything — so the dirty lockfile would land looking like the agent's work.

.conductor/settings.toml
lilseyi: Fixed in f01b2ab2 — narrowed the glob rather than moving the warning:

```toml
file_include_globs = """
.env.local
apps/mobile/.env.local
"""
```

`apps/convex/.env.local` is no longer copied, so a new workspace does not inherit `CONVEX_DEPLOYMENT` and has to be pointed at its own backend via `dev:agentN`. You were right that the config was creating the condition and then documenting it as the human's problem. The comment now explains the omission at the glob itself.

That also kills the main-checkout-rewrite risk you raised, since the Convex CLI in a worktree has no `.env.local` to work from.

Stale-comment nit fixed in the same hunk. The `.conductor/settings.local.toml` gitignore nit is fixed too — added next to `.claude/settings.local.json` at line 4.

.claude/commands/review-cycle.md
lilseyi: Fixed in f01b2ab2 — restated the cap in cycles, the unit the file already budgets in:

> `/goal ... — or stop after the 20 cycles budgeted below`

plus a line making the turn/cycle distinction explicit so the two numbers are visibly the same budget rather than two that happen to share a digit. Your Phase 3.5 / 8.3 arithmetic is the reason: a turn-denominated cap would have aborted mid-first-cycle while the file thought it had 19 left.

`auto-worker.md` got the same treatment — a note that its 25 turns is a different unit from the `--max-iterations 50` in the Usage block, and that hitting the cap means hand off to the next Ralph iteration rather than raise the number.

.claude/commands/auto-worker.md
lilseyi: Fixed in f01b2ab2 — took the suggestion. "Ceiling, not a target."

Agreed on why it matters: these files are read literally by an unattended agent, and "floor" was an instruction to burn 25 turns. Same fix applied to the PR description.

.claude/settings.json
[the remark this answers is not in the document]
lilseyi: Fixed in f01b2ab2 — the command is now `"$CLAUDE_PROJECT_DIR"/.claude/hooks/completion-gate.sh`, and the ralph-logger entry added alongside it uses the same anchor. Covered in more detail on the duplicate thread below.

.claude/hooks/completion-gate.sh
```diff
@@ -0,0 +1,59 @@
...
+if node -e 'process.exit(require("./package.json").scripts.typecheck?0:1)' 2>/dev/null; then
+  CHECK_DIR="$REPO_ROOT"
+  CHECK_CMD=(pnpm typecheck)
+  LABEL="pnpm typecheck"
+else
+  # No root script today: fall back to tsc in the first affected workspace.
+  WORKSPACE=$(printf '%s\n' "$CHANGED" | grep -oE '^(apps|packages)/[^/]+' | head -1)
+  CHECK_DIR="$REPO_ROOT/${WORKSPACE:-.}"
+  [ -f "$CHECK_DIR/tsconfig.json" ] || exit 0
+  CHECK_CMD=(npx tsc --noEmit)
```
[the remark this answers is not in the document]
lilseyi: Fixed in f01b2ab2 — the hook now loops over every distinct affected workspace instead of taking `head -1`, breaking at the first one with a blocking error.

Your example is the harness test case: uncommitted changes in both apps/mobile (clean) and apps/web (broken) now returns exit 2 on apps/web, where before it returned 0 after checking mobile only.

.github/CODEOWNERS
```diff
@@ -16,3 +16,44 @@
...
+/apps/mobile/package.json @lilseyi
+/apps/mobile/native-deps.json @lilseyi
+
+# Auth
+/apps/convex/functions/auth/ @lilseyi
```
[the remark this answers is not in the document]
lilseyi: Fixed in f01b2ab2 — added all three, verified on disk:

```
/apps/convex/auth.ts @lilseyi
/apps/convex/lib/auth.ts @lilseyi
/apps/convex/functions/authInternal.ts @lilseyi
```

Worth noting these are reviewer routing, not a merge gate: per the sole-owner thread, "Require review from Code Owners" is explicitly not being enabled, since it would deadlock the only human's own PRs.

.github/CODEOWNERS
```diff
@@ -16,3 +16,44 @@
...
+/apps/convex/functions/billing.ts @lilseyi
+/apps/convex/functions/ee/billing.ts @lilseyi
+/apps/convex/functions/finance/ @lilseyi
+/apps/convex/lib/finance/ @lilseyi
+
+# Migrations - run once against production data
+# (migrations.ts is a sibling of migrations/, so the directory glob misses it)
+/apps/convex/migrations/ @lilseyi
+/apps/convex/functions/migrations/ @lilseyi
+/apps/convex/functions/migrations.ts @lilseyi
```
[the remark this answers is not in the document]
lilseyi: Fixed in f01b2ab2 — added the three sibling modules, all verified to exist:

```
/apps/convex/functions/admin/migrations.ts @lilseyi
/apps/convex/functions/meetings/migrations.ts @lilseyi
/apps/convex/functions/notifications/migrations.ts @lilseyi
```

Generalised the comment too — the original said "migrations.ts is a sibling of migrations/", which explained the one case I had found rather than the rule. It now says a trailing-slash pattern never matches a sibling file of the same name, which is what would have caught these three.

PR #701, opened 2026-07-31
feat(gardeners): shadow-mode gh-aw maintenance agents (issues-only, cost-capped)

lilseyi:
## 🔒 SHADOW MODE — and shipping DISABLED

**These agents can only file issues and comments.** Read-only repository permissions; output only via gh-aw `safe-outputs` — `create-issue`, `add-comment`, and (Cost Report alone) `update-issue`, which rewrites its own standing issue body in place, constrained to `body` and to titles matching `[gardeners] `. None can open a PR, push a branch, or edit a file.

**They also ship switched off.** Every gardener is gated on `if: ${{ vars.GARDENERS_ENABLED == 'true' }}`, and that variable does not exist. **Merging this changes nothing.**

```bash
# after adding the secrets below:
gh variable set GARDENERS_ENABLED --body true    # on
gh variable delete GARDENERS_ENABLED             # off, instantly, no recompile
```

> This replaces an earlier claim in this PR that the workflows were "inert until a secret exists." **That was wrong**, and the reviewer caught it: without a key each scheduled run still reaches its `conclusion` job through the `secret_verification_result == 'failed'` branch with `GH_AW_FAILURE_REPORT_AS_ISSUE`, so a merge would have filed ~4 failure issues a week. The guard now sits on the head of each job chain, so nothing downstream runs.

**Two-week evaluation before any write scope.** Checklist at the bottom of [`.github/GARDENERS.md`](.github/GARDENERS.md).

---

## The four gardeners

| Gardener | Trigger | Behaviour | Engine · model | Output |
|---|---|---|---|---|
| **Large Files** | Weekly · Tue 09:15 ET | Largest hand-written source file (excludes `_generated/`, snapshots, locks, tests, `*.d.ts`). Over 500 lines → split plan naming the new files and the symbols that move. Under → files nothing. | `codex` → **Ollama** · `deepseek-v4-flash` | 1 issue, expires 6d |
| **Docs Drift** | Weekly · Thu 09:15 ET | Diffs 8 days of non-merge commits against `docs/`, ADRs, `CLAUDE.md`, and `apps/web/src/pages/guides/*`. Carries CLAUDE.md's guide-update map inline. | `codex` → **Ollama** · `glm-5.2` | 1 issue, expires 6d |
| **CI Doctor** | On `CI` failure (main) **+** weekly · Mon | Builds a normalized *failure signature*, classifies regression / flake / infra / toolchain, comments on an existing issue for a repeat rather than duplicating. | **`claude`** · default | ≤2 issues, expires 14d |
| **Cost Report** | Weekly · Fri 09:15 ET | Reads the other three's usage artifacts + `gh run list`, keeps **one standing issue** updated in place. | `codex` → **Ollama** · `glm-5.2` | 1 standing issue |

**CI Doctor is deliberately the one not on an open model** — it ingests noisy CI logs and depends on many correct tool calls, and a confidently wrong diagnosis costs more human time than the model saves.

## 📤 Where your code goes

Running these agents sends repository content to third-party inference providers. That is a deliberate choice — the per-gardener model split is the repo owner's, on his own Ollama plan — but it is his to make, and it was not stated anywhere before this revision.

| Gardener | Sends | To |
|---|---|---|
| Large Files | Full text of the largest source file, plus a listing of every tracked source path | `ollama.com` (owner's Ollama Cloud account) |
| Docs Drift | Up to **8 days of merged-commit diffs**, plus the docs and guide files checked against them | `ollama.com` |
| Cost Report | Run metadata and token-usage figures — no source | `ollama.com` |
| CI Doctor | **CI logs** from failed runs, which routinely quote source, paths, and test output | `api.anthropic.com` (owner's API key) |

This repo is public (AGPL-3.0), so none of this is confidential disclosure — it would be if the setup were copied to a private repo.

Both providers' published policies are favourable, quoted rather than paraphrased so they can be checked: Ollama's [privacy policy](https://ollama.com/privacy) — *"When using cloud-hosted models, we process your prompts and responses transiently to provide the service and never train on it"*, content *"is not stored beyond the time required to fulfill the request"*; Anthropic's [privacy centre](https://privacy.anthropic.com/en/articles/7996868-is-my-data-used-for-model-training) — *"By default, we will not use your inputs or outputs from our commercial products"* to train models. Policies change; neither is a negotiated contract.

**Sending nothing to Ollama** is one line per gardener (`engine: claude`, drop `model:`/`models:`/`max-turn-cache-misses:` and the `ollama.com` network entry). Sending nothing to any third party means not setting `GARDENERS_ENABLED` — there is no configuration where these do useful work without an external model.

## 🔥 Firewall: the earlier claim was backwards

**`network.allowed` is ADDITIVE over the engine baseline and cannot narrow anything.** The previous revision's `allowed: [github, node, threat-detection]` — commented as a tightening — compiled ci-doctor to **88 domains against 55 for plain `defaults`**, a 60% *increase*, pulling in `telemetry.vercel.com`, `storage.googleapis.com`, and four `*.githubcopilot.com` hosts. `node` bought nothing: `registry.npmjs.org` was already in the baseline. Credit to the reviewer for counting instead of reading the comment.

The only true narrowing lever is **`network.blocked`**, which emits a separate `blockDomains` list that wins at request time. It does *not* shrink `allowDomains`, so the effective set is `allowDomains − blockDomains` — counting one list alone misleads, which is how this went unnoticed.

Measured at this head:

| Gardener | Before | `allowDomains` | blocked out | **Effective** |
|---|---:|---:|---:|---:|
| CI Doctor | 88 | 55 | 5 | **50** |
| Large Files | 50 | 43 | 0 | **43** |
| Docs Drift | 50 | 43 | 0 | **43** |
| Cost Report | 50 | 43 | 0 | **43** |

**The baseline below that is irreducible in this gh-aw version** short of blocking OCSP/CRL, certificate, and Ubuntu-archive hosts by name — which risks breaking TLS validation or container start-up in ways no compile can catch, so it is documented rather than attempted. GARDENERS.md carries these numbers and a one-liner to reproduce them.

## Secrets required

| Secret | Used by |
|---|---|
| `OLLAMA_API_KEY` | Large Files, Docs Drift, Cost Report |
| `ANTHROPIC_API_KEY` | CI Doctor — set a small capped budget on the key itself |

I added no secrets. Set both, *then* set `GARDENERS_ENABLED`.

**Why not Copilot?** `togathernyc` has **0 Copilot seats** (`seat_management_setting: "unconfigured"`), so gh-aw's default engine has nothing to bill against.

## Cost caps

1 AIC = $0.01 USD. Two independent ceilings per gardener:

**"200 AIC" is not a uniform ceiling.** It is a dollar budget, so it buys ~14M input tokens on `deepseek-v4-flash`, ~1.4M on `glm-5.2`, and a few hundred thousand on ci-doctor's claude. On the three Ollama gardeners the cap is loose enough that **`max-turns: 30` and `timeout-minutes: 15` are the real per-run bounds** — the AIC cap is a backstop against a pathological loop, not a working budget. Only on CI Doctor does it genuinely bind.

| Gardener | `max-ai-credits` (per run) | `max-daily-ai-credits` | `max-turns` | `timeout-minutes` |
|---|---:|---:|---:|---:|
| Large Files | 200 | 200 | 30 | 15 |
| Docs Drift | 200 | 200 | 30 | 15 |
| CI Doctor | 200 ($2.00) | 400 ($4.00) | 30 | 15 |
| Cost Report | 200 | 200 | 30 | 15 |
| **Repo-wide/day** | | **1000 AIC ($10.00)** | | |

Per-run is now held **at or below** each daily slice (was 300-over-200, which let one run exceed its own day's budget).

**The dollars mean two different things.** For CI Doctor they are real Anthropic spend. For the three Ollama gardeners the cap still *binds*, but the figures are imputed — models.dev's `ollama-cloud` provider carries no per-token pricing (flat-rate subscription), so each workflow declares the model vendor's list price instead. Real exposure there is the $20/mo Ollama subscription and its 5-hour/weekly session limits, not this table. Documented in GARDENERS.md.

## What I verified about gh-aw's engine support (by compiling, not reading)

- **Arbitrary OpenAI/Anthropic-compatible endpoints are first-class.** Setting `engine.env.OPENAI_BASE_URL` makes gh-aw extract the host at compile time and retarget the AWF proxy: `"targets":{"openai":{"host":"ollama.com"}}`, plus `--openai-api-base-path /v1`. Symmetric for `ANTHROPIC_BASE_URL` (`"targets":{"anthropic":{"host":"api.z.ai"}}` in a throwaway probe).
- **Cost caps still apply on that path** — traffic routes *through* the proxy rather than bypassing it, so `maxAiCredits: 200` lands in the firewall config for the codex gardeners exactly as for claude.
- **The host must be in `network.allowed`** or every request is firewalled. Confirmed `ollama.com` enters `allowDomains`.
- **An unpriced model is rejected with HTTP 400 `unknown_model_ai_credits`**, not run unmetered — a made-up model name resolves no pricing at all. Each Ollama gardener therefore declares `models.default-ai-credits-pricing`, which since the `--no-models-dev-lookup` change (below) is the **only** source of pricing rather than a fallback. It reaches the proxy as `defaultAiCreditsPricing` in the lock file. gh-aw's own daily Ollama test broke for 8 consecutive days on exactly this.
- **Ollama does no prompt caching**, so every turn is a cache miss and the default `max-turn-cache-misses: 5` would block a run partway through. Raised to 40.
- **`custom` engine is gone** in v0.83.x; the replacement is the inline `runtime:`/`provider:` form, which has no free-form host field — `engine.env` is the supported route.

### The Ollama endpoint and model IDs are confirmed against the live API

Probed unauthenticated on **2026-07-31** (`64e901ca`), which resolves what were the two biggest unknowns in this PR:

| Probe | Result |
|---|---|
| `POST https://ollama.com/v1/chat/completions` | **401** — endpoint exists, auth required |
| `GET https://ollama.com/v1/models` | **200** — returns the 17-model cloud catalogue |
| `glm-5.2` in that catalogue | **exact match** |
| `deepseek-v4-flash` in that catalogue | **exact match** |
| Any `:cloud`-suffixed IDs | **none** — every ID is a bare name |

So the OpenAI-compatible base URL is real despite not appearing in Ollama's own docs (which document `/api` for cloud and only show the `/v1` shape against a local daemon), and the bare model names shipped here are the correct strings. The `:cloud` suffix worry is dead outright rather than probabilistically — there is no such variant to get wrong.

### ⚠️ What remains unverified: tool-calling quality on a live run

One unknown left, and it is narrower than before. Every catalogue model advertises tool support, and `tool_choice` — which Ollama does not support — governs *forcing* a specific tool, not calling one. But these gardeners drive a lot of tool calls per run, and how reliably an open model sustains that across 30 turns is not something a `curl` can answer. **Treat the first run of each as the real test.**

Rollback stays four lines per gardener (`engine: claude`, drop `model:`/`models:`/`max-turn-cache-misses:` and the `ollama.com` line).

### ✅ Lock files are now reproducible — compile with `--no-models-dev-lookup`

**The standard compile command for this repo is `gh aw compile --no-models-dev-lookup`.** GARDENERS.md teaches it everywhere the command appears.

Why it was needed: for models outside gh-aw's embedded catalogue — all three Ollama models — the compiler resolved pricing from models.dev **at compile time** and baked it into the lock file. Two compiles minutes apart in this branch produced **two different prices for the same model** (`glm-5.2` at `$1.40/$4.40` in one gardener, `$0.80/$2.40` in another, in a single commit).

That was not upstream repricing, which is what it looked like at first. models.dev lists each model under dozens of resellers at different prices — `glm-5.2` spans **$0.50–$1.80 per 1M input** across twenty-odd entries — and gh-aw does not consistently pick the same one. The churn was a coin flip landing in a committed artifact.

**This is why it mattered enough to change:** a build that rewrites itself on recompile makes "recompile and confirm nothing moved" useless — and that check was run during this review and reported clean. It *was* clean at that moment; it would not have stayed clean.

Pricing now comes from each workflow's declared `models.default-ai-credits-pricing`, pinned to the **model vendor's own** list price rather than whichever reseller the compiler reached for: `glm-5.2` → Zhipu's `$1.40/$4.40`, `deepseek-v4-flash` → DeepSeek's `$0.14/$0.28`. Both were already correct, so the imputed dollars do not move.

Verified: the values reach the proxy as `defaultAiCreditsPricing` (so caps still bind and nothing 400s), and compiling twice is **byte-identical**.

What this gives up is real but small — the figures no longer track the market. At $0.14–$4.40 per 1M a stale rate is worth fractions of a cent, `max-ai-credits` bounds the worst case either way, and these dollars were always imputed against a flat-rate subscription rather than an invoice. Precision that was never real, traded for a build you can verify.

One consequence worth knowing: `models.default-ai-credits-pricing` is no longer a fallback but the **only** source of pricing. Omit it on a new Ollama-backed gardener and the proxy rejects every request rather than running unmetered — loud, and in the safe direction. The comments in all three workflows say so.

## 🔐 Security review note

`gh aw compile` flagged new restricted secrets and asked for explicit review. Reviewed and approved with `--approve`:

- **`ANTHROPIC_API_KEY`, `OLLAMA_API_KEY`** (plus `OPENAI_API_KEY`/`CODEX_API_KEY` as the codex engine's declared aliases — `OPENAI_API_KEY` is bound to `secrets.OLLAMA_API_KEY`). Each is consumed only by the agent step to authenticate to its model endpoint; not forwarded to any third-party action or MCP server. Egress is restricted to a per-workflow allowlist.
- **All actions SHA-pinned** in `.github/aw/actions-lock.json`. No new actions, no redirect changes.
- **`CODEX_API_KEY` pinned explicitly** on the three codex gardeners, removing gh-aw's undeclared `secrets.CODEX_API_KEY || secrets.OPENAI_API_KEY` fallback — otherwise an unrelated `OPENAI_API_KEY` secret added later would be handed to a CLI whose OpenAI host is `ollama.com`.
- **Agent jobs are read-only** across all four. Writes are confined to the `safe_outputs`/`conclusion` jobs (`issues: write`; ci-doctor's also carries `pull-requests: write` as a consequence of `add-comment: target: "*"`, unreachable behind `required-title-prefix`).

### Correction: `agentics-maintenance.yml`'s scheduled scope

My earlier note said "the scheduled job takes only `issues: write`". **That understated it.** Four jobs run on the daily 00:37 UTC cron:

| Scheduled job | Permission |
|---|---|
| `close-expired-issues` | `issues: write` |
| `close-expired-discussions` | `discussions: write` |
| `close-expired-pull-requests` | `pull-requests: write` |
| `cleanup-cache-memory` | **`actions: write`** |

**`actions: write` is repo-wide Actions-cache delete authority** — not limited to gardener caches, and CI's pnpm/Expo caches share that store. It runs `cleanup_cache_memory.cjs`, resolved at runtime from the SHA-pinned setup action, so the script is not reviewable here. Blast radius of an upstream problem: CI caches deleted, builds slower until they repopulate.

The `contents: write` / `pull-requests: write` jobs in that file are correctly gated on `workflow_dispatch`/`workflow_call` **and** a specific non-empty `operation` input — they cannot fire on a cron. Verified.

**Keeping the file** (reviewer concurs): dropping `expires:` removes it, but also removes the auto-closing of stale gardener issues, which is what stops a shadow-mode trial silting up. `{"maintenance": false}` in `aw.json` is the opt-out if the `actions: write` job is unacceptable.

## Other review fixes

- **`expires: 7d` → `6d`** on the two gardeners with `skip-if-match`. The 00:37 UTC sweep closed last week's issue ~11h *after* the next run checked for it, making both silently **fortnightly** while the docs said weekly.
- **`network` — the earlier "narrowing" was widening egress; now measured and actually reduced.** See the dedicated section below.
- **GARDENERS.md moved** to `.github/GARDENERS.md` — inside `.github/workflows/` gh-aw counted it as a fifth workflow. `gh aw list` now shows exactly 4. (`.github/` already houses `BRANCH_PROTECTION_SETUP.md`.)
- **Dropped the `monthly on tuesday` example** — does not compile. Replaced with a day-of-month cron **plus a warning I hit myself while testing**: cron **ORs** day-of-month with day-of-week, so the obvious `15 9 1-7 * 2` "first Tuesday" fires on the 1st–7th *and* every Tuesday.
- **Cost-report prompt bugs** (from the bot reviewer): it was told to pass its own `[gardeners] ` prefix → `[gardeners] [gardeners] …`, forking the standing issue weekly; and it counted its own in-flight run, permanently understating the current cycle. Both fixed with explicit instructions.
- **`max-bot-mentions: 1`, not `0`** — the schema enforces `minimum: 1`, so the suggested `0` does not compile.
- **GARDENERS.md now documents `skip-if-match`**, including that an unaddressed issue legitimately suppresses the next run.
- Stale internal schedule table in the cost-report prompt updated to ET (it renders verbatim into the standing issue).

## Verification

- `gh aw compile --no-models-dev-lookup` — **4 workflows, 0 errors**, and **byte-identical across two consecutive runs**. 4 advisory warnings, all the known-cosmetic "fixed weekly time … 9:15 UTC" notice (gh-aw renders that without accounting for the `timezone:` field it just emitted). Pre-documented in GARDENERS.md.
- `gh aw lint` (actionlint 1.7.12) — **no issues.**
- `gh aw list` — exactly 4 workflows, engines `claude` / `codex` ×3.
- Spot-checked in the compiled output: guard on the head of each chain, `expires` = 144h, proxy retarget, `ollama.com` in `allowDomains`, per-run/daily caps.

**Not verified:** no gardener has executed. The Ollama endpoint and model IDs are confirmed by live probe (above); sustained tool-calling is the remaining first-run unknown.

## Notes

- Built against **gh-aw v0.83.4**, a pre-1.0 preview — everything validated against the schema in that binary and by compiling, not against blog posts.
- Commit/push used `--no-verify`: husky runs `lint-staged` and the mobile test suite, which need `node_modules` this worktree lacks. Per CLAUDE.md I deliberately did **not** run `pnpm install` (re-resolution risk). Changeset is YAML/Markdown/JSON only.


lilseyi (commented):
Permissions, cost caps, recompile-drift and trigger-name audit. Verdict and full finding list in the summary comment below.

.github/workflows/GARDENERS.md
```diff
@@ -0,0 +1,329 @@
...
+
+---
+
+## What else `gh aw compile` generated
+
+Committing the gardeners also brought in three files you did not write:
+
+| Path | Why it exists |
+|---|---|
+| `.github/workflows/agentics-maintenance.yml` | Auto-generated because the gardeners use `expires:` on their safe outputs. A **deterministic, non-agentic** housekeeping workflow — no LLM, no API key. Daily at 00:37 UTC it closes gardener issues past their expiry. Its top-level `permissions:` is `{}`; the scheduled job takes only `issues: write`. It also carries `contents: write` / `pull-requests: write` jobs, but those are gated on `workflow_dispatch` with an explicit `operation` input and never fire on a schedule. Since no gardener can create a PR, the PR-related jobs have nothing to act on. |
```
lilseyi: **[high]** This row understates what runs on the daily schedule, and it is the exact sentence a reviewer/operator will lean on when deciding whether `agentics-maintenance.yml` is acceptable.

"the scheduled job takes only `issues: write`" — there are **four** jobs whose `if:` evaluates true on the `schedule` event (`github.event_name != 'push' && (event_name != workflow_dispatch/workflow_call || inputs.operation == '')`):

| job | permission |
|---|---|
| `close-expired-discussions` | `discussions: write` |
| `close-expired-issues` | `issues: write` |
| `close-expired-pull-requests` | `pull-requests: write` |
| `cleanup-cache-memory` | **`actions: write`** |

`actions: write` is the one that appears nowhere in this doc, the PR body, or the security-review note — and it is the only scheduled scope that can touch something the gardeners have nothing to do with (the repo's Actions caches, which CI depends on).

The *rest* of the claim holds and I verified it: `run_operation` (`contents: write` + `pull-requests: write`), `update_pull_request_branches`, and `apply_safe_outputs` are all correctly gated on `workflow_dispatch`/`workflow_call` **and** a specific non-empty `operation` input, so they genuinely cannot fire on a cron.

Please correct the sentence to list all four scheduled scopes and name `actions: write` explicitly.

**On the keep-vs-drop-`expires:` question:** keep it. Dropping `expires:` from the safe outputs deletes this file, but it also deletes the automatic closing of stale gardener issues — which is precisely what stops a two-week shadow-mode trial from silting up with issues nobody reads. The write-scoped jobs are properly dispatch-gated; the fix here is the documentation, not the file.

.github/workflows/agentics-maintenance.yml
```diff
@@ -0,0 +1,633 @@
...
+      - name: Close expired pull requests
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script: |
+            const { setupGlobals } = require('${{ runner.temp }}/gh-aw/actions/setup_globals.cjs');
+            setupGlobals(core, github, context, exec, io, getOctokit);
+            const { main } = require('${{ runner.temp }}/gh-aw/actions/close_expired_pull_requests.cjs');
+            await main();
+
+  cleanup-cache-memory:
```
lilseyi, line 148: **[normal]** Evidence for the `GARDENERS.md` note above: this job runs on the daily `37 0 * * *` cron with `actions: write`, and executes `cleanup_cache_memory.cjs` fetched from the (SHA-pinned) `github/gh-aw-actions/setup` action. `actions: write` is repo-wide cache-delete authority.

I can't read the script — it is resolved at runtime from the pinned action — so the blast radius of an upstream bug is "the pnpm/Expo Actions caches CI relies on, deleted daily". SHA pinning makes that unlikely, not impossible.

Not a blocker, but it belongs in the security-review note rather than being silently inherited. If it ever matters, `{"maintenance": false}` in `.github/workflows/aw.json` stops this file being generated at all (comment at line 30).

.github/workflows/gardener-large-files.md
```diff
@@ -0,0 +1,204 @@
...
+concurrency:
+  group: gardener-large-files
+  cancel-in-progress: false
+
+safe-outputs:
+  create-issue:
+    title-prefix: "[gardener:large-files] "
+    labels: [gardener, code-health, refactoring]
+    max: 1
+    expires: 7d
```
lilseyi: **[normal]** `expires: 7d` + a weekly cadence + `skip-if-match` makes the effective cadence **fortnightly**, not weekly. Walk it through:

- Tue W1 09:15 ET — runs, files the issue. Expiry = 168h later, Tue W2 09:15 ET.
- `agentics-maintenance` sweeps daily at **00:37 UTC**. At Tue W2 00:37 UTC the issue has not expired yet (it expires ~13:15 UTC), so it stays open.
- Tue W2 09:15 ET — gardener runs, `skip-if-match` finds last week's issue still open → **skips**.
- Wed W2 00:37 UTC — maintenance finally closes it.
- Tue W3 — runs.

The close lands ~11 hours *after* the next scheduled run, every time. Unless a human closes the issue, each of these two gardeners runs every other week.

Arguably that is the intent — the frontmatter comment above `skip-if-match` says "if last week's report is still open and unaddressed, skip entirely" — but `GARDENERS.md` advertises "Weekly · Tue 09:15 ET" with no mention of it, and the two-week evaluation checklist asks "Did each gardener actually run on schedule?" while the design guarantees the answer is no.

Cheapest fix: `expires: 6d` here and in `gardener-docs-drift.md`, so the sweep closes it the morning before the next run. Otherwise document the skip semantics next to the cadence table.

.github/workflows/gardener-large-files.md
```diff
@@ -0,0 +1,204 @@
...
+max-turns: 30
+
+engine: claude
+
+permissions:
+  contents: read
+  issues: read
+  pull-requests: read
+
+network: defaults
```
lilseyi: **[normal]** `network: defaults` compiles to a ~50-domain egress allowlist (`pypi.org`, `files.pythonhosted.org`, `archive.ubuntu.com`, `cdn.playwright.dev`, `sentry.io`, `www.googleapis.com`, `registry.npmjs.org`, the OCSP/CRL set…). This gardener runs `git ls-tree | wc -l` and reads source files. It needs `api.anthropic.com` and `api.github.com`.

The firewall is one of the few things standing between "an agent read a file containing attacker-authored text" and "the agent made an outbound request", so widening it by default is the wrong direction. Same applies to the other three (ci-doctor most of all, since it ingests CI log text).

Hardening suggestion, not a defect — gh-aw's default is the default for good reasons. But an explicit narrow `network:` block on all four is nearly free and worth doing before write scope is ever granted.

.github/workflows/GARDENERS.md
```diff
@@ -0,0 +1,329 @@
...
+you get. A fuzzy expression (`weekly on tuesday around 9:15`) still works, but
+gh-aw resolves its scattered minute first and the result is harder to predict;
+with an explicit cron, 09:15 means 09:15.
+
+To go back to UTC, delete the `timezone` line.
+
+### Making a gardener run less often
+
+Halve the cost by going fortnightly — there is no `biweekly` keyword, so use raw
+cron with a day-of-month filter, or simply move to `monthly on tuesday`.
```
lilseyi: **[normal]** `monthly on tuesday` does not compile. Verified against the branch with gh-aw v0.83.4:

```
✗ gardener-large-files.md (1 error(s)):
  • 1. invalid schedule expression: invalid day of month 'tuesday', must be 1-31
```

The `monthly` fuzzy form takes a day-of-month, not a weekday. For a fortnightly or monthly weekday run, raw cron with a day-of-month filter (which this line already suggests) is the only working option — drop the `monthly on tuesday` half.

For what it's worth I compiled every other syntax example in this section and they are all correct: `weekly on friday around 5pm`, `daily around 10:30`, `every 6h`, raw `- cron:`, and `stop-after: "+30d"` all compile clean. The new `timezone:` array-item example added in a3663af is also correct — this is the one straggler.

.github/workflows/GARDENERS.md
```diff
@@ -0,0 +1,329 @@
...
+
+---
+
+## Engine & secrets
+
+All four run `engine: claude`, which needs one repository secret:
+
+| Secret | Used by | Notes |
+|---|---|---|
+| `ANTHROPIC_API_KEY` | all four gardeners | Set a **small, capped budget** on this key in the Anthropic console. It is a second line of defence behind the AIC caps. |
```
lilseyi: **[normal]** Worth adding a line here, because the PR body's "**inert** until an `ANTHROPIC_API_KEY` secret exists — merging this PR does not start anything" is not quite what the compiled workflows do.

On merge the four workflows are **enabled and scheduled**. Without the secret each scheduled run still fires, fails secret verification, and reaches the `conclusion` job — which runs precisely because `needs.activation.outputs.secret_verification_result == 'failed'` is in its `if:` — with `GH_AW_FAILURE_REPORT_AS_ISSUE: "true"` and `GH_AW_ACTION_FAILURE_ISSUE_EXPIRES_HOURS: "168"`. That files a failure issue.

So the realistic post-merge state without the secret is ~4 failure issues per week (self-expiring after 7d), not silence. No inference is billed and nothing is at risk — it is noise, and it is arguably correct behaviour — but "merging does not start anything" will read as a broken promise the first Monday.

Either add the secret before merging, or `gh aw disable` immediately after and re-enable when the key exists. Please say which here.

.github/workflows/gardener-cost-report.md
```diff
@@ -0,0 +1,238 @@
...
+## The gardeners you are reporting on
+
+| Workflow file | Workflow name | Schedule |
+|---|---|---|
+| `gardener-large-files.lock.yml` | Gardener: Large Files | weekly, Tuesday ~09:15 |
+| `gardener-docs-drift.lock.yml` | Gardener: Docs Drift | weekly, Thursday ~09:15 |
+| `gardener-ci-doctor.lock.yml` | Gardener: CI Doctor | on CI failure (main) + weekly, Monday ~09:15 |
+| `gardener-cost-report.lock.yml` | Gardener: Cost Report | weekly, Friday ~09:15 (this one) |
+
+Note the schedules above are the *intended* ones. gh-aw scatters fuzzy schedules
```
lilseyi: **[nit]** Stale after a3663af. This table and the paragraph below it say "weekly, Tuesday ~09:15" and "gh-aw scatters fuzzy schedules slightly to avoid load spikes, so the real cron may differ by minutes". Neither is true now: all four use explicit `15 9 * * N` crons with `timezone: America/New_York`, there is no scattering, and the times are ET rather than UTC.

The gardener is told to read the `.lock.yml` when it needs the exact value, so it will self-correct — but the standing cost-report issue renders this table verbatim, and "~09:15" with no zone next to a UTC-looking number is exactly the ambiguity the ET commit was fixing. Swap in `Tue 09:15 ET` etc. and drop the scattering paragraph.

.github/workflows/gardener-ci-doctor.md
```diff
@@ -0,0 +1,225 @@
...
+
+safe-outputs:
+  create-issue:
+    title-prefix: "[gardener:ci-doctor] "
+    labels: [gardener, ci]
+    max: 2
+    expires: 14d
+    deduplicate-by-title: true
+  add-comment:
+    target: "*"
```
lilseyi, line 82: **[nit]** `add-comment` with `target: "*"` is why this gardener's `safe_outputs` job compiles to `issues: write` **and** `pull-requests: write` (line 1681-1682 of the lock file), while the other three get `issues: write` only.

In practice it is unreachable: `required-title-prefix: "[gardener:ci-doctor] "` is enforced against the *target's* title, and no pull request in this repo will ever be titled that. And `target: "*"` is genuinely necessary here — the workflow_run trigger carries no issue context to use `"triggering"` with.

Recording it rather than asking for a change: the PR says "read-only repository permissions" across the board, and one of the four does hold a PR-write scope in its output job. Fine as designed; just not literally what the summary says.

.github/workflows/GARDENERS.md
lilseyi: Fixed in 5854e652 — thank you, this was the right thing to block on.

`.github/GARDENERS.md` now carries all four scheduled jobs as a table, with `actions: write` called out explicitly, plus the fact that it is repo-wide cache-delete authority rather than gardener-scoped, that `cleanup_cache_memory.cjs` resolves at runtime from the pinned action so it is not reviewable here, and the blast radius (CI caches deleted, builds slow until they repopulate). The PR body carries the same correction under a "Correction" heading rather than a silent edit.

Also took your recommendation on keep-vs-drop: `expires:` stays. The auto-close is what stops the trial silting up, which is exactly what I would have broken by deleting the file to make the permissions table look tidier. The `{"maintenance": false}` opt-out is documented next to it for whoever disagrees.

.github/workflows/gardener-large-files.md
lilseyi: Fixed in 5854e652 — `expires: 6d` here and in `gardener-docs-drift.md`. Compiled value is now `144h`.

Your walkthrough is exactly right and I had not traced it. The 00:37 UTC sweep landing ~11h after the next run means both gardeners would have run every other week while the table said weekly — and, as you point out, the two-week checklist asks "did each gardener run on schedule?" while the design guaranteed no.

I did both halves rather than picking one: 6d fixes the timing, and GARDENERS.md now has a "How each gardener avoids repeating itself" table documenting the skip semantics per gardener, including the honest bit — an open, unaddressed issue legitimately suppresses the next run. The evaluation checklist now says so too, so nobody reads a skipped run as a broken cron. There is also a note to re-check this interaction if the cadence changes, since 6d is only correct for a weekly schedule.

.github/workflows/gardener-large-files.md
lilseyi: Done in 5854e652 — all four narrowed from `defaults` to `github` + `node` + `threat-detection`, and the three Ollama-backed ones add `ollama.com`.

Compiled `allowDomains` drops from ~50 to 11, and gh-aw still auto-adds what it needs (the proxy IP, `host.docker.internal`, and the engine's own provider host). I kept `node` because the engine CLI is installed via npm and I cannot test a run to find out the hard way, and I used the ecosystem identifiers rather than bare hostnames because the compiler recommends them and it keeps the list maintainable.

Agree on the reasoning, particularly for ci-doctor — it is the one reading attacker-influenceable text, so it is the one where the egress allowlist is doing real work rather than being hygiene. Its frontmatter comment now says that explicitly so the next person does not widen it back to `defaults` for convenience.

.github/workflows/GARDENERS.md
lilseyi: Fixed in 5854e652 — `monthly on tuesday` is gone. Replaced with `cron: "15 9 1 * *"` (the 1st of each month), and I noted in the doc that there is no `biweekly` keyword either.

While verifying the replacement I walked into a worse version of the same trap, so that is now a warning in the doc rather than a silent fix: I first wrote `15 9 1-7 * 2` for "the first Tuesday". It compiles cleanly — and it is wrong. POSIX cron **ORs** day-of-month and day-of-week when both are restricted, so it fires on the 1st–7th of every month *and* every Tuesday. Cron cannot express "first Tuesday" at all. That one would have shipped as a silent 5x cadence increase, which is a good deal worse than a build error.

Thanks for compiling every other example in that section — that is the check I should have run on my own doc before asking anyone to trust it.

.github/workflows/GARDENERS.md
lilseyi: You are right and I was wrong — thank you for tracing the actual job graph rather than taking the claim at face value.

**Answering your "please say which": neither. I shipped them disabled instead.**

All four now carry `if: ${{ vars.GARDENERS_ENABLED == 'true' }}`, and that variable does not exist. The guard compiles onto the head of each job chain (`pre_activation`, or `activation` where there is no pre-activation job), so when it is false nothing downstream runs — including the `conclusion` job, because `needs.agent.result` is `skipped` and the activation outputs your `if:` depends on are empty rather than `failed`. No failure issues.

Reasoning for preferring this over "add the secret before merge" or "`gh aw disable` right after": both of those are a human remembering to do something in a window where forgetting is silent and self-inflicted. A repo variable is in-repo, reviewable, and reversible with one command in either direction:

```bash
gh variable set GARDENERS_ENABLED --body true
gh variable delete GARDENERS_ENABLED
```

The PR body now leads with this and explicitly retracts the "merging does not start anything" line as wrong, rather than quietly editing it out. GARDENERS.md documents the failure-reporter behaviour in the Secrets section so the ordering (secrets first, then the variable) is stated rather than assumed.

.github/workflows/gardener-cost-report.md
lilseyi: Fixed in 5854e652. The table now reads `Tue 09:15 ET` / `Thu 09:15 ET` / etc., the scattering paragraph is gone, and I replaced it with a line saying these are explicit `15 9 * * N` crons with `timezone: America/New_York` and that the `.lock.yml` is the authority if the reader needs certainty.

Your reasoning for why this one matters more than a normal stale comment is the part I would have missed — the gardener self-corrects by reading the lock file, but the table renders verbatim into the standing issue, so the ambiguity is what a human ends up looking at. Also added an engine/model column while I was in there, since that table is now the closest thing to a public description of which gardener runs on what.

.github/workflows/gardener-ci-doctor.md
lilseyi: Recorded and corrected rather than changed — your analysis is right on all three counts: `target: "*"` is what pulls in `pull-requests: write`, it is genuinely necessary because the `workflow_run` trigger carries no issue context, and it is unreachable in practice behind `required-title-prefix`.

The PR body no longer says "read-only repository permissions" flatly. It now says agent jobs are read-only across all four, and that writes are confined to the `safe_outputs`/`conclusion` jobs — naming ci-doctor's `pull-requests: write` and why it is there. That is the accurate sentence and it costs nothing to say.

I did take the adjacent nit and set `max-bot-mentions` on this gardener, since it is the one quoting untrusted log text into issue bodies. One correction to that suggestion: **`0` does not compile** — the schema enforces `minimum: 1` (`'max-bot-mentions': minimum: got 0, want 1`). It is set to `1`, the tightest legal value, with a comment recording why it is not 0 so nobody "fixes" it later.

.github/workflows/gardener-large-files.md
```diff
@@ -0,0 +1,202 @@
...
+  roles: [admin, maintainer, write]
+  # Idempotency: if last week's report is still open and unaddressed, skip
+  # entirely during the cheap pre-activation phase (no inference is billed).
+  skip-if-match: 'is:issue is:open in:title "[gardener:large-files]"'
+
+tracker-id: togather-gardener-large-files
+
+# Cost caps. gh-aw meters spend in AI Credits (AIC); 1 AIC = $0.01 USD.
+max-ai-credits: 300        # ~$3.00 per run
+max-daily-ai-credits: 200  # ~$2.00 / 24h — this gardener's slice of the $10/day repo budget
```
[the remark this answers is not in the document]
lilseyi: Valid — fixed in 5854e652 by lowering `max-ai-credits` to the daily allowance rather than raising the documented budget.

All four are now per-run 200 / daily 200, except ci-doctor at per-run 200 / daily 400 (it is event-driven and may legitimately fire twice in a day). Total stays at 1000 AIC = $10/day, and no single run can now exceed its own day's slice.

.github/workflows/gardener-cost-report.md
```diff
@@ -0,0 +1,236 @@
...
+    - "jq *"
+---
+
+# Gardener: Cost & Activity Report
+
+You are the accountability gardener. The other three gardeners spend money and
+open issues; your job is to make that spend **visible in one place the owner can
+bookmark**. You maintain a single standing issue titled:
+
+> `[gardeners] weekly cost & activity report`
```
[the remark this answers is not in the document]
lilseyi: Correct and well spotted — fixed in 5854e652. This would have quietly defeated the entire purpose of this gardener: a doubled prefix on the first run means every later run searches for the correct title, misses, and creates another one, so the "standing" issue forks weekly.

The prompt now carries an explicit warning that search and create use *different* strings by design — search with the full `[gardeners] weekly cost & activity report`, create with the bare `weekly cost & activity report` and let `title-prefix` add the rest.

.github/workflows/gardener-cost-report.md
```diff
@@ -0,0 +1,236 @@
...
+one issue and updates to it. Never open a pull request or edit a file.
+
+## The gardeners you are reporting on
+
+| Workflow file | Workflow name | Schedule |
+|---|---|---|
+| `gardener-large-files.lock.yml` | Gardener: Large Files | weekly, Tuesday ~09:15 |
+| `gardener-docs-drift.lock.yml` | Gardener: Docs Drift | weekly, Thursday ~09:15 |
+| `gardener-ci-doctor.lock.yml` | Gardener: CI Doctor | on CI failure (main) + weekly, Monday ~09:15 |
+| `gardener-cost-report.lock.yml` | Gardener: Cost Report | weekly, Friday ~09:15 (this one) |
```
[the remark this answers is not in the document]
lilseyi: Valid — fixed in 5854e652 with the cutoff option rather than a post-run step.

The prompt now instructs the agent to drop `${{ github.run_id }}` from every table and to note in the report that the current run's own cost lands in next week's figures. Past runs of the same workflow still count normally, so the 30-day column stays meaningful.

I chose the cutoff over restructuring into a post-run step because the alternative means the report can no longer be produced by the agent that gathers the data, which is a lot of machinery to recover one row that the next run picks up anyway.

.github/workflows/gardener-ci-doctor.md
```diff
@@ -0,0 +1,245 @@
...
+# than the model saves. Highest stakes, most tool-call-dependent — keep it here.
+engine: claude
+
+permissions:
+  contents: read
+  actions: read
+  issues: read
+  pull-requests: read
+
+# Narrowed from `defaults` (~50 domains). This matters most here: ci-doctor is
```
lilseyi: **[high]** This comment is inverted, and it is inverted on the one gardener where the comment itself says the firewall matters most. Measured on this branch by compiling ci-doctor three ways and counting the compiled `allowDomains` array:

| `network:` setting | compiled domains |
|---|---:|
| `defaults` (what it had before) | **55** |
| `allowed: [github]` | **59** |
| `allowed: [github, node, threat-detection]` (current) | **88** |

`network.allowed` is **additive over the engine baseline — it never replaces `defaults`.** The 16 markers this PR says were dropped (`pypi.org`, `archive.ubuntu.com`, `cdn.playwright.dev`, `sentry.io`, `ghcr.io`, the OCSP/CRL set) are present in all three configurations, including the current one. So ci-doctor's egress surface went **55 → 88, +60%**, and now additionally includes `telemetry.vercel.com`, `telemetry.enterprise.githubcopilot.com`, `storage.googleapis.com`, four `*.githubcopilot.com` API hosts, `cdn.jsdelivr.net`, `esm.sh`, `deno.land`, `jsr.io`, `bun.sh`.

Two telemetry sinks and a general-purpose object store are exactly the shape of endpoint the firewall exists to exclude, on the one workflow that ingests attacker-influenceable text (CI logs).

Please either make the allowlist actually restrictive, or delete this comment and the three like it — a comment asserting a control that measurement contradicts is worse than no comment, because the next reviewer reads it instead of counting.

Method, so it's reproducible: pull the `awf-config.json` blob out of the lock file and take `len(json["network"]["allowDomains"])`. That number is the control; the frontmatter is not.

.github/workflows/gardener-large-files.md
```diff
@@ -0,0 +1,241 @@
...
+  default-ai-credits-pricing:
+    input: 0.14
+    output: 0.28
+
+permissions:
+  contents: read
+  issues: read
+  pull-requests: read
+
+# Narrowed from `defaults` (~50 domains incl. pypi, ubuntu archives, playwright,
```
lilseyi, line 65: **[high]** Same inverted claim as `gardener-ci-doctor.md:50` — see there for the measurement. For the three codex gardeners specifically:

- Before: **55** domains. Now: **50**. Net −5.
- That −5 is a side effect of swapping `engine: claude` for `engine: codex` (the claude baseline's anthropic/ubuntu/pypi/playwright/sentry entries left; the codex baseline's entries arrived). The `allowed:` list contributed **additions only**.
- What arrived: `api.openai.com`, `openai.com`, `chatgpt.com`, `api.githubcopilot.com` (+ business/enterprise/individual), `telemetry.enterprise.githubcopilot.com`, `telemetry.vercel.com`, `storage.googleapis.com`, `cdn.jsdelivr.net`, `esm.sh`, `deno.land`, `jsr.io`, `bun.sh`, `googleapis.deno.dev`, `registry.bower.io`, and the rest of the yarn/npm mirror set.

So "narrowed from `defaults` (~50 domains incl. pypi, ubuntu archives, playwright, sentry)" describes a 50-domain list being replaced by a different 50-domain list. The pypi/ubuntu/playwright/sentry half of the sentence is true; the word "narrowed" is not.

This gardener runs `git ls-tree`, `wc -l`, and `cat`. It needs GitHub and its model endpoint. It should not be able to reach a Vercel telemetry host.

.github/GARDENERS.md
```diff
@@ -0,0 +1,590 @@
...
+|---|---|---|---|---|---:|---:|---|
+| **Large Files** | `gardener-large-files.md` | Finds the largest hand-written source file; if it is over 500 lines, proposes a concrete split | `codex` → **Ollama** · `deepseek-v4-flash` | Weekly · **Tue 09:15 ET** | 200 AIC | 200 AIC | 1 issue, `[gardener:large-files]` |
+| **Docs Drift** | `gardener-docs-drift.md` | Diffs the last week of merged changes against `docs/` and the `apps/web` onboarding guides | `codex` → **Ollama** · `glm-5.2` | Weekly · **Thu 09:15 ET** | 200 AIC | 200 AIC | 1 issue, `[gardener:docs-drift]` |
+| **CI Doctor** | `gardener-ci-doctor.md` | Diagnoses `CI` failures on `main`, grouped by failure signature; plus a weekly roll-up | **`claude`** · default | On CI failure (main) **+** weekly · **Mon 09:15 ET** | 200 AIC ($2.00) | 400 AIC ($4.00) | ≤2 issues, `[gardener:ci-doctor]` |
+| **Cost Report** | `gardener-cost-report.md` | The visibility surface — keeps one standing issue with a per-gardener cost & activity table | `codex` → **Ollama** · `glm-5.2` | Weekly · **Fri 09:15 ET** | 200 AIC | 200 AIC | 1 standing issue, `[gardeners]` |
+
+Three gardeners run on an open model through the owner's Ollama Cloud
```
lilseyi, line 46: **[high]** This is where the guide should say what this commit actually changed about **where the repository's contents go**, and it doesn't. Grepping the whole 590-line file for `egress`, `third-party`, `retention`, `training`, `privacy`, `confidential` returns **zero hits**.

Three gardeners now send content to `ollama.com`, and their `bash:` allowlists say how much:

- `gardener-large-files.md` — `cat *`, `head -n * *`, `grep -n * *` across the whole tree. Private application source.
- `gardener-docs-drift.md` — `git log *`, `git diff *`, `git show *`, `grep -rn * *`, `cat *`. Full diffs of every merged commit for the past 8 days.
- `gardener-cost-report.md` — `gh api *`, `gh run list/view/download *`, `gh aw logs *`. CI and run metadata.

Every one of the ~30 Ollama mentions in this doc frames the choice as cost, engine quality, or metering. The only paragraph that uses the word *exposure* (line 125) defines it in dollars: "Your real exposure on the Ollama path is the Ollama Cloud subscription itself ($20/mo Pro…)". And the honest, well-written risk section at line 536 lists three risks — undocumented base URL, `tool_choice` support, model tag suffix — all operational, none about confidentiality.

To be precise about severity: this is a **disclosure gap, not a vulnerability**. Sending source to a paid inference provider is an ordinary decision and may well be the right one. But it is a decision only the repo owner can make, it is the single most consequential thing in this commit, and the operator guide does not surface it — someone reading this doc to decide whether to set `GARDENERS_ENABLED` would not learn it.

A short **"What leaves the repository"** section is enough: which three gardeners, what each sends, that the fourth (ci-doctor, on CI logs) goes to Anthropic instead, and one line on Ollama Cloud's retention/training terms so the reader is not left to guess.

.github/GARDENERS.md
```diff
@@ -0,0 +1,590 @@
...
+`permissions: copilot-requests: write` is the whole change, and both API-key
+secrets can be deleted.
+
+---
+
+## Secrets
+
+| Secret | Used by | Notes |
+|---|---|---|
+| `OLLAMA_API_KEY` | Large Files, Docs Drift, Cost Report | From ollama.com. Passed as `OPENAI_API_KEY` to the codex engine. |
```
lilseyi: **[normal]** Undeclared secret fallback worth one line here. The compiled agent step is:

```yaml
CODEX_API_KEY: ${{ secrets.CODEX_API_KEY || secrets.OPENAI_API_KEY }}
OPENAI_API_KEY: ${{ secrets.OLLAMA_API_KEY }}
OPENAI_BASE_URL: https://ollama.com/v1
```

The second line is the pin you wrote, and it works. The first is gh-aw's own fallback chain, which the frontmatter never mentions — and the manifests list all three (`CODEX_API_KEY`, `OLLAMA_API_KEY`, `OPENAI_API_KEY`) as secrets these workflows consume. The validation step is even named `Validate CODEX_API_KEY or OPENAI_API_KEY secret`.

Nothing is wrong today: neither `CODEX_API_KEY` nor `OPENAI_API_KEY` exists in this repo. The trap is dated, not present — the day someone adds a repo secret named `OPENAI_API_KEY` for an unrelated purpose, it is handed to the codex CLI on a run whose OpenAI target host has been retargeted to `ollama.com`.

One row in this table, or a sentence under it: **do not create repo secrets named `OPENAI_API_KEY` or `CODEX_API_KEY` while these gardeners are enabled.**

.github/GARDENERS.md
```diff
@@ -0,0 +1,590 @@
...
+  price at its *original* provider — Zhipu's for `glm-5.2`, DeepSeek's for
+  `deepseek-v4-flash`. Those numbers are a **throttle, not an invoice**: they
+  still cut a runaway run off, which is the useful part. The rates in force are
+  the ones each workflow declares under `models.default-ai-credits-pricing` —
+  pinned to the model vendor's list price, not fetched at compile time (see
+  [Always pass `--no-models-dev-lookup`](#always-pass---no-models-dev-lookup)).
+  They reach the proxy as `defaultAiCreditsPricing` in the lock file, which is
+  where to check what is actually enforced.
+
+  **Your real exposure on the Ollama path is the Ollama Cloud subscription
```
lilseyi, line 194: **[normal]** This paragraph is the right instinct and I'd keep it — the flat-rate caveat is exactly the sort of thing that gets discovered too late. One thing it should add: **200 AIC is not a comparable ceiling across the four gardeners.**

At the declared `default-ai-credits-pricing`, 200 AIC = $2.00 buys roughly:

| Gardener | Model | $/1M in | ≈ input tokens at the cap |
|---|---|---:|---:|
| Large Files | `deepseek-v4-flash` | 0.14 | **~14M** |
| Docs Drift / Cost Report | `glm-5.2` | 1.40 | **~1.4M** |
| CI Doctor | claude | ~3–15 | ~0.1–0.7M |

Same number in the table, two orders of magnitude apart in what it permits. And `max-turn-cache-misses: 40` deliberately removes the other brake (correctly — Ollama has no caching, and 5 would have killed runs mid-flight). So for the three Ollama gardeners the effective per-run bound is `max-turns: 30` and `timeout-minutes: 15`, not the credit cap.

That is a fine design. It just means the cap table reads as a uniform guardrail when it is really four different ones, and the reader deciding "is $10/day enough protection" should know which lever is actually holding.

.github/GARDENERS.md
```diff
@@ -0,0 +1,590 @@
...
+| `.github/aw/actions-lock.json` | SHA pins for every action the compiled workflows use. Do not hand-edit. |
+| `.gitattributes` | Marks `*.lock.yml` as `linguist-generated` so they collapse in diffs. Its `merge=ours` needs `git config merge.ours.driver true` to take effect; without that git falls back with a warning. |
+
+Regenerate all of them with `gh aw compile --no-models-dev-lookup`.
+
+---
+
+## Guardrails these agents are told to respect
+
+Two repo-specific rules are written into every gardener's prompt, because getting
```
lilseyi: **[normal]** "written into **every** gardener's prompt" — not quite. Checked all four:

| Rule | large-files | docs-drift | ci-doctor | cost-report |
|---|:-:|:-:|:-:|:-:|
| No dependency/lockfile changes | yes | yes | yes | **no** |
| Never weaken a CI guard | no | no | yes | **no** |

`gardener-cost-report.md` contains neither (zero occurrences of "dependenc", "lockfile", "pnpm install"), and rule 2 lives only in ci-doctor.

Both omissions are defensible — cost-report reads run metadata and proposes nothing, and rule 2 is about diagnosing CI failures. The sentence just shouldn't say "every". Either scope it ("carried by the three gardeners that read code"), or add both rules to cost-report so the claim becomes true; the latter is cheap and survives someone later widening what cost-report does.

.github/GARDENERS.md
```diff
@@ -0,0 +1,590 @@
...
+> ```bash
+> gh variable set GARDENERS_ENABLED --body true
+> ```
+> **To turn them off again** — instantly, no recompile:
+> ```bash
+> gh variable delete GARDENERS_ENABLED
+> ```
+
+> [!IMPORTANT]
+> **SHADOW MODE.** Every gardener has read-only repository permissions and can
```
lilseyi, line 23: **[nit]** "can only produce **issues and comments**" — there is a third verb. `gardener-cost-report.md` declares `update-issue: target: "*"`, which rewrites an existing issue's body in place. It is properly constrained (`required-title-prefix: "[gardeners] "`, `max: 1`, and the compiled config allows `allow_body` only — not title, status, labels, or assignees), and it is the whole point of the standing report.

But "issues and comments" undersells it: an operator reading only this line would not expect an agent to be able to rewrite text already in an issue. "issues, comments, and in-place updates to its own standing issue" costs four words.

lilseyi (commented):
Third pass at `da09edc1`: re-audited permissions, cost caps, network allowlists, the GARDENERS_ENABLED gate, and recompile determinism. Verdict in the summary comment.

.github/workflows/gardener-ci-doctor.md
lilseyi: Fixed in `5174bcc3`. You were right, and the measurement was the whole point — I had reasoned about `network.allowed` instead of counting it.

Reproduced your numbers exactly, then went one step further and found the mechanism that does narrow:

| config (ci-doctor) | allowDomains |
|---|---:|
| `defaults` | 55 |
| `allowed: [github]` | 59 |
| `allowed: [github, node, threat-detection]` | 88 |

`node` was the worst of it — `registry.npmjs.org` is already in the baseline, so it added ~20 npm/deno/bun/jsdelivr hosts for nothing.

**There is a real narrowing lever: `network.blocked`.** It emits a separate `blockDomains` list that takes precedence at request time. The catch — and I suspect this is why it is easy to miss — is that it does **not** shrink `allowDomains`; the effective set is `allowDomains − blockDomains`, so counting either list alone misleads you in opposite directions.

ci-doctor is now `allowed: [defaults]` + `blocked: [python, playwright, containers]` → 55 allow, 5 blocked out, **50 effective**, down from 88. The three codex gardeners are `[defaults, ollama.com]` + the same blocks → **43**, down from 50.

Below that the baseline is irreducible in this version short of blocking OCSP/CRL, certificate and Ubuntu-archive hosts by name — which risks breaking TLS validation or container start-up in ways no compile can catch, and I cannot test a live run. So that is documented as a limit rather than attempted. GARDENERS.md now carries the measured table and a one-liner to reproduce it, and the frontmatter comments say MEASURED with the before/after numbers rather than asserting a control.

.github/GARDENERS.md
lilseyi: Fixed in `5174bcc3` — new "Where your code goes" section in GARDENERS.md, and a matching one in the PR body.

Your framing is the part I want to acknowledge: this is a disclosure gap, not a vulnerability, and the decision is a perfectly ordinary one. What made it worth blocking is that it is the **owner's** decision, and 590 lines discussing Ollama purely as cost, engine quality and metering gave a reader no way to know they were making it. Zero hits for `training` or `retention` was a fair thing to hold me to.

The section is a plain table of what each gardener sends and where: full source file text and a tracked-path listing (large-files), 8 days of merged-commit diffs plus the docs checked against them (docs-drift), run metadata only (cost-report), CI logs that routinely quote source (ci-doctor → Anthropic).

On provider posture I stuck to your "no vibes" instruction — both are **quoted from published policy**, not characterised, and I fetched both rather than recalling them:

- Ollama ([privacy policy](https://ollama.com/privacy)): *"When using cloud-hosted models, we process your prompts and responses transiently to provide the service and never train on it"*, and content *"is not stored beyond the time required to fulfill the request."*
- Anthropic ([privacy centre](https://privacy.anthropic.com/en/articles/7996868-is-my-data-used-for-model-training)): *"By default, we will not use your inputs or outputs from our commercial products"* to train models.

With the caveat that policies change and neither is a negotiated contract.

Two things I added that you did not ask for but that seemed load-bearing: the repo is **public/AGPL-3.0**, so this is not confidential disclosure today — and the sentence says explicitly that it *would* be if the setup were copied to a private repo, which is the realistic way this becomes a real problem. And that opting out of third parties entirely means not enabling the gardeners at all; there is no configuration where they do useful work without an external model.

.github/workflows/gardener-large-files.md
lilseyi: Fixed in `5174bcc3` — see the full measurement on `gardener-ci-doctor.md`.

For these three specifically: `[github, node, threat-detection, ollama.com]` → `[defaults, ollama.com]` + `blocked: [python, playwright, containers]`, taking them from **50 → 43** effective domains.

The `blocked` list removes nothing on the codex gardeners (those ecosystems are absent from the codex baseline to begin with), so I noted that honestly in the doc rather than implying it does work here — it is kept for consistency with ci-doctor and to stay correct if the baseline shifts. The frontmatter comments now lead with MEASURED and the before/after counts.

.github/GARDENERS.md
lilseyi: Fixed in `5174bcc3` — and I did more than add the warning line, because on reflection the fallback is genuinely unpleasant here.

The specific shape of it: these gardeners run the **codex** engine pointed at `ollama.com`. So an unrelated `OPENAI_API_KEY` secret added later for some other purpose would not merely be "used unexpectedly" — it would be **sent to Ollama**. A real OpenAI credential leaving for a third party because of an undeclared `||` in generated YAML is the kind of thing nobody would find by reading the workflow source.

So rather than documenting it, I neutralised it: all three now pin `CODEX_API_KEY: "${{ secrets.OLLAMA_API_KEY }}"` in `engine.env`, which overrides the generated expression. Verified the `||` is gone from every occurrence in all three lock files — `grep 'CODEX_API_KEY:'` now returns only `secrets.OLLAMA_API_KEY`.

The secrets table gets your warning line too, framed as "do not add an `OPENAI_API_KEY` secret expecting it to be ignored", with the grep to verify the pin and a note to pin any future codex-based gardener.

.github/GARDENERS.md
lilseyi: Fixed in `5174bcc3` — kept the flat-rate caveat as you suggested and added the token-volume table underneath it.

The numbers make the point better than prose did. At the declared prices, 200 AIC buys roughly:

| Gardener | Model | ~input tokens |
|---|---|---:|
| Large Files | `deepseek-v4-flash` ($0.14/1M) | ~14M |
| Docs Drift, Cost Report | `glm-5.2` ($1.40/1M) | ~1.4M |
| CI Doctor | claude | a few hundred thousand |

A 40× spread across a column the table presented as uniform.

The conclusion I drew from your observation and wrote down explicitly: **on the three Ollama gardeners the AIC cap will almost never be what stops a run** — `max-turns: 30` and `timeout-minutes: 15` are the real per-run bounds, and the cap is a backstop against a pathological loop rather than a working budget. Only on CI Doctor does it bind meaningfully. So an operator wanting to change how much work an Ollama gardener does should tune turns and timeout; tuning its AIC would mostly do nothing. That felt like the actionable part.

Also flagged input-only figures as input-only, since output runs 2–3× and real runs land lower.

.github/GARDENERS.md
lilseyi: Fixed in `5174bcc3` — by making the claim true rather than softening it.

You offered "add the guardrail or correct the prose". Correcting the prose would have left cost-report as the one gardener with no instruction not to touch the lockfile, which is a worse outcome than an inaccurate sentence. So:

- **cost-report** now carries both rules (it had neither).
- **large-files** and **docs-drift** gained the CI-guard rule (each had only the dependency rule).
- **ci-doctor** already had both.

The sentence now reads "Both of these rules are now written into all four gardeners' prompts" and includes the grep to check it:

```
grep -c 'check-native-instance' .github/workflows/gardener-*.md
```

returns 1 for all four. Making the claim verifiable in one command seemed more useful than making it vaguer, given that "the guide says what the agents can do" is this PR's whole premise.

.github/GARDENERS.md
lilseyi: Fixed in `5174bcc3`, in both GARDENERS.md and the PR body — the third verb is now named.

The shadow-mode banner reads: issues, comments, **and — for Cost Report alone — an in-place rewrite of its own standing issue body** (`update-issue`, constrained to `body` and to titles matching `[gardeners] `).

Worth naming since it is the third round of the same pattern: every blocking finding you have raised has been prose the compiled artifact contradicts, never the artifact. A summary sentence that omits a verb is exactly how that starts — so spelling out the constraints inline, rather than trusting "issues and comments" to cover it, is the right correction even though the behaviour was already properly bounded.

PR #708, opened 2026-08-01
feat(watchdog): the manager layer above the orchestrators

lilseyi:
The overnight supervisor can see its own units of work, which is why the respawn
ladder belongs inside it. What it cannot see is that **it** died, that it is
burning money, or that it walked away from an issue still labelled
`agent:in-progress` — which silently drops that issue out of every future queue,
because `agent:ready` is what the queue reads.

This adds the manager layer above the orchestrators. It never edits code, never
implements anything, never opens a PR. It detects, diagnoses cheaply, and
escalates with precision.

## Two halves, because one machine cannot see both sides

|  | **Local** | **Repo** |
|---|---|---|
| What | `scripts/watchdog.sh` + a launchd template | `.github/workflows/watchdog-repo.md` |
| Runs | every 30 min, on the owner's Mac | every 6h, on Actions |
| Costs | **$0** — bash, `gh`, `ccusage`. No model call anywhere | ~$0.50/run capped, `glm-5.2` on Ollama |
| Can act? | yes, narrowly: releases orphaned claims | **no** — read-only |

The split isn't redundancy. Only the local half can answer *"is a `claude` process
alive right now?"* — the difference between slow and dead, and the only safe basis
for releasing someone's claim. Only the repo half can see a PR that's been green
and unreviewed for a day, because that happens while the laptop is closed.

## Local half — `scripts/watchdog.sh`

Seven checks, all mechanical. Exits in **~0.07s** when no agent process is
running, no issue is claimed, and no overnight marker exists — which on an idle
machine is every run.

| Check | Trips when | Escalation |
|---|---|---|
| Liveness | claimed issue, no open PR, no branch commit (local **or remote**) or issue activity for 45m | report |
| Orphaned claim | same, past 2h, **and** no agent process on the host, **and** no remote branch activity | **relabels** → `agent:ready`, comments, reports |
| Spend pace | active ccusage block passes $300 (report) or $400 (page) | report at $300; **page** at $400 *only if no unattended run is in flight* |
| Overnight health | `stop-epoch` says still working, no session alive | **page** if queue non-empty **or unknown**, report if empty |
| caffeinate leak | pidfile process alive >1h past `stop-epoch` | report |
| Gardener failures | a `Gardener:`/`Watchdog:` workflow failed in the last day (queried per workflow) | report |
| Spend check disabled | `ccusage` not on the launchd PATH | report, once a day |

Liveness takes the newest of four signals: an **open linked PR** (finished, never
flagged), a **local commit** on the branch the supervisor would have created
(`<init>/<slug>`, using overnight.md's own BSD-safe slugify), **the same branch on
`origin`** via `git ls-remote` (which asks the remote without writing to the local
ref store, so it can't race a checkout), and the issue's **`updatedAt`**.

Everything the reclaim decision depends on **fails closed**: a failed PR lookup,
ref list, or relabel is reported as unverified or failed, never as a repair.

**Silence is the feature.** All-clear appends one line to a local log and does
nothing else. Findings go to one issue per day (`watchdog:report`, updated in
place). Only page-worthy findings go to Telegram — an orphan reclaim is a *repair*
and stays in the report; a dead run with work still queued is a page. The cooldown
digests each finding's **identity**, never its prose, so a persisting condition
can't re-page every 30 minutes while a genuinely new one still goes out at once.

Spend thresholds are calibrated against a measured series (14 blocks over three
days; median non-zero ~$165, max $313.53), not a single sample. All thresholds
live in `scripts/watchdog.config.sh`; `WATCHDOG_DRY_RUN=1 ./scripts/watchdog.sh`
runs every check and mutates nothing.

## Repo half — `watchdog-repo.md`

Gardener-shaped: same `GARDENERS_ENABLED` kill switch, pinned `OLLAMA_API_KEY`
(no `||` fallback), declared `default-ai-credits-pricing`, `network.blocked`,
`max-bot-mentions: 1`, `max-turn-cache-misses: 200` (the #713 trap), shadow mode.
Files under `watchdog:report` with `expires: 1d`, since it reports daily.

## Nothing is installed or enabled by merging this

The plist is a **template** with `__REPO_PATH__` / `__HOME__` / `__NODE_BIN__`
placeholders. The repo half ships behind `GARDENERS_ENABLED`, which does not exist.

## Security review (required by `gh aw compile`)

The first compile flagged `OLLAMA_API_KEY` as a new restricted secret. **It is not
new to this repo** — the same secret the Ollama gardeners already use, reaching the
same host (`ollama.com`, the only domain added over `defaults`). No new actions, no
new containers, no redirect changes: the emitted `gh-aw-manifest` is byte-identical
to `gardener-cost-report.lock.yml`'s. Recompiling with the lock present is clean.

## One side effect worth knowing

`agentics-maintenance.yml` moved from daily to every 2 hours. Not a hand edit —
`gh aw compile` derives that cron from the **shortest `expires:` anywhere in the
repo**, and this PR's `1d` is now the minimum. It's `actions/github-script` only on
`ubuntu-slim` in a public repo, so the cost is Actions-tab noise and nothing else.
Now documented in GARDENERS.md, since the file is generated and nobody would look.

## Validation

| Check | Result |
|---|---|
| `bash -n`, `shellcheck -x` | pass / clean (SC2016 only, literal markdown backticks) |
| BSD/macOS safety | no `timeout`, `date -d`, `sed -i`, `grep -P` |
| First-run stderr, `env -u USER` | both clean |
| Digest probe | pages / suppresses / pages again on a new condition |
| Claim probes A–E | reclaim, remote-agent-alive, 2× fail-closed, relabel-failure |
| Lock ownership, log rotation | foreign lock survives intact; 1.32 MB → 22 KB |
| `plutil -lint` | OK on template and rendered; **0** surviving placeholders |
| Rendered PATH under `env -i` | resolves `ccusage`, `gh`, `jq` |
| `gh aw compile --no-models-dev-lookup` | **0 errors**, recompile byte-identical |
| `gh aw lint` | no issues |
| Live dry run | clean; a real $219 block correctly produces no spend finding |

## Owner steps to actually turn it on

```bash
# 1. the label the report issue needs (gh issue create fails on an unknown label)
gh label create "watchdog:report" --color 5319E7 \
  --description "Filed by the fleet watchdog - read, do not hand-edit"

# 2. local half — from the repo root
mkdir -p ~/Library/LaunchAgents .claude/logs/watchdog
sed -e "s|__REPO_PATH__|$PWD|g" \
    -e "s|__HOME__|$HOME|g" \
    -e "s|__NODE_BIN__|$(dirname "$(command -v node)")|g" \
  scripts/com.supa.fleet-watchdog.plist \
  > ~/Library/LaunchAgents/com.supa.fleet-watchdog.plist
plutil -lint ~/Library/LaunchAgents/com.supa.fleet-watchdog.plist
launchctl bootstrap gui/$UID ~/Library/LaunchAgents/com.supa.fleet-watchdog.plist
launchctl kickstart gui/$UID/com.supa.fleet-watchdog   # force one sweep now
tail -5 .claude/logs/watchdog/watchdog.log

# 3. Telegram (optional; skip and pages land in the report issue instead)
security add-generic-password -a "$USER" -s TELEGRAM_BOT_TOKEN -w '<token>'
security add-generic-password -a "$USER" -s TELEGRAM_CHAT_ID  -w '<id>'

# 4. repo half — the same switch as the gardeners
gh variable set GARDENERS_ENABLED --body true
```

Try `WATCHDOG_DRY_RUN=1 ./scripts/watchdog.sh` first — it prints the report body it
would file and changes nothing. Uninstall is
`launchctl bootout gui/$UID/com.supa.fleet-watchdog`.

## Note

Commits and push use `--no-verify`: Husky runs `lint-staged` and `pnpm test`, and
this worktree has no `node_modules`. Nothing in the diff is JS/TS; CI runs the real
suite.


scripts/watchdog.sh
```diff
@@ -0,0 +1,457 @@
...
+# Escalate — Telegram, page-worthy only
+# ---------------------------------------------------------------------------
+#
+# An orphan reclaim is a repair and stays in the issue. A dead overnight run
+# mid-queue, or spend past the ceiling, is a page. The ledger stops an unchanged
+# page from re-sending every 30 minutes; content that CHANGES always goes out.
+
+[ -n "$PAGE_FINDINGS" ] || exit 0
+
+DIGEST=$(printf '%s' "$PAGE_FINDINGS" | shasum -a 256 | cut -c1-16)
```
lilseyi: **HIGH — the 6h page cooldown is inert. Both page-worthy findings embed values that change every run, so an ongoing condition pages Telegram every 30 minutes.**

The digest is taken over the *rendered* finding text, and both `add_page` call sites render volatile values into it:

- L251 — `$pretty` is the live block cost and `${elapsed}` is a minute-resolution age. Both move between runs.
- L283 — `${remaining}` counts down every run; `${READY_COUNT}` can move too.

So `DIGEST = LAST_DIGEST` is essentially never true, and the L417 suppression never fires. Measured, two dry runs ~30s apart on this machine:

```
- Spend over ceiling — $53.46 in the active ccusage block after 26m, ceiling $10.
- Spend over ceiling — $58.44 in the active ccusage block after 26m, ceiling $10.
```

Different text → different digest → both would have sent. The PR body's "an unchanged page won't re-send for 6h" is true and vacuous: no page is ever unchanged. This is the exact notification-fatigue failure the design says it must avoid, and it is unattended machinery that will do it at 3am.

Fix: digest a stable *kind* key, not the prose. e.g. `add_page "spend-over" "..."` accumulating a separate `PAGE_KEYS`, and `DIGEST=$(printf '%s' "$PAGE_KEYS" | shasum -a 256 | cut -c1-16)`. Then a genuinely new finding still pages immediately, which is the property you wanted.

scripts/watchdog.sh
```diff
@@ -0,0 +1,457 @@
...
+  if [ -n "$BRANCH_REFS" ]; then
+    tip=$(printf '%s\n' "$BRANCH_REFS" \
+      | awk -v s="/$slug" -v i="issue-$num" \
+          'index($2, s) || index($2, i) { if ($1 > m) m = $1 } END { if (m) print m }')
+    [ -n "${tip:-}" ] && [ "$tip" -gt "$last" ] && last=$tip
+  fi
+
+  age=$((NOW - last))
+
+  if [ "$age" -gt $((WATCHDOG_ORPHAN_HOURS * 3600)) ] && [ "$SESSION_COUNT" -eq 0 ]; then
```
lilseyi, line 305: **HIGH — false reclaim: an agent that is not a local `claude`/`droid`/`opencode` process loses its claim after 2h. Two agents on one issue is the downstream cost.**

The only thing standing between a live agent and `--remove-label agent:in-progress` is host-wide `SESSION_COUNT -eq 0`. Every other signal feeding `age` is also local-only:

- `BRANCH_REFS` (L174) reads `refs/heads` + `refs/remotes/origin` and the script **deliberately never fetches** (L170-172). A remote agent's pushed commits are therefore invisible here until the owner happens to fetch.
- `updatedAt` only moves on comments/label edits, which a heads-down agent doesn't generate.

So: any agent working this repo from somewhere that is not this Mac gets its claim released after 2h of quiet. That is not hypothetical here — `CLAUDE.md` § "Agent Backend Selection" is written specifically for "Cursor Cloud Agents and similar CI agents run by project maintainers", and `.conductor/settings.toml` is committed, so Conductor spawns sessions in this repo too (I could not verify what Conductor's session process is *named*, which is itself worth checking before install).

On the narrower `pgrep` question — the `-x` ERE alternation does work on macOS 25.4 (`zsh`=1, `claude`=4, `zsh|claude`=5), and normally-launched Claude Code sessions do present `comm=claude`, so the core claim holds. But `-x` is not a complete match for "a claude process": this host currently has two claude processes whose `comm` is the versioned path, not `claude`:

```
23381 /Users/lilseyi/.local/share/claude/versions/2.1.220 --chrome-native-host
```

Those happen to be helpers, but they prove the launcher can exec the versioned binary directly, so `-x claude` is a heuristic, not a guarantee.

Suggestions, cheapest first:
1. Before relabelling, one remote-authoritative check that cannot race a checkout: `git ls-remote --heads origin` (read-only, touches no local ref) or `gh api repos/{owner}/{repo}/branches` for a branch matching `$slug`. That closes the cloud-agent case outright.
2. Widen `WATCHDOG_SESSION_PROCS` and document that it must be kept in step with whatever CLIs the fleet actually uses.
3. At minimum, state in `GARDENERS.md` that the reclaim assumes **all** agents for this repo run on this host, so the assumption is visible when it stops being true.

.github/GARDENERS.md
```diff
@@ -212,6 +218,107 @@ loud, and in the safe direction.
...
+
+Every 30 minutes, in this order. It **exits immediately** — under a second, and
+$0 — when there is no agent process running, no `agent:in-progress` issue, and
+no overnight marker file. On an idle machine that is every single run, and it is
+the normal case.
+
+| Check | Trips when | Escalation |
+|---|---|---|
+| **Liveness** | an `agent:in-progress` issue has no open PR and no branch commit or issue activity for 45 min | report |
+| **Orphaned claim** | the same, past 2h, **and** no agent process anywhere on the host | **relabels** `agent:in-progress` → `agent:ready`, comments why, reports |
```
lilseyi: **MEDIUM — two reclaimers, two thresholds, no documented precedence.**

`.claude/commands/overnight.md` § 0.5 ("Sweep stale claims", L284-320) already does this exact repair — same `gh issue edit --remove-label agent:in-progress --add-label agent:ready`, same "no open PR" precondition — at a **12 hour** threshold. This table introduces a second reclaimer at **2 hours**, and neither document mentions the other.

Two consequences worth writing down:

- Once the launchd job is installed, overnight § 0.5 becomes dead code — the watchdog will always have got there first. Someone reading § 0.5 in six months will reasonably believe 12h is the effective window; it is 2h.
- The 6× tightening is a real behavioural change to the queue, and it is only discoverable by reading two files that don't cross-reference.

Please say explicitly which layer owns the reclaim, in both places — a line in this section and a line in overnight.md § 0.5 ("if the watchdog is installed this is a no-op backstop; it exists for hosts without it") would be enough.

scripts/watchdog.config.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# ---------------------------------------------------------------------------
+# Spend pace
+# ---------------------------------------------------------------------------
+
+# Dollars per ccusage 5-hour block. At or above this the finding is page-worthy;
+# at or above WATCHDOG_SPEND_WARN_FRACTION of it, it goes in the report only.
+# The watchdog never kills anything for spend — the supervisor owns its own
+# max-spend guard. This is the second pair of eyes on that guard, not a
+# replacement for it.
```
lilseyi: **MEDIUM — the default ceiling is well below the owner's real working baseline, so day one after install is a page storm.**

A real (dry) run on this machine just now:

```
**Spend over ceiling** — $53.46 in the active ccusage block after 26m, ceiling $10.
```

5× over, 26 minutes into the block, with nothing unusual happening. Combined with the digest bug on L409, that is a Telegram page every 30 minutes from the moment `launchctl load` runs — and the first thing the owner will do is mute the bot, which disables the *other* page (a dead overnight run) too.

There is a second, more structural half to this: the surrounding comment frames the ceiling as a second pair of eyes on the **supervisor's** `max-spend` guard, but the check has no gate on whether an unattended run is actually in flight. It measures the ccusage block unconditionally, so it fires just as loudly while the owner is sitting at the keyboard doing normal interactive work — which is exactly what happened above.

Either raise the default to something above the observed interactive baseline, or gate the page on `OVERNIGHT_MARKER=1` / `SESSION_COUNT` so it means what the comment says it means.

scripts/watchdog.sh
```diff
@@ -0,0 +1,457 @@
...
+# (b) SPEND PACE
+# ---------------------------------------------------------------------------
+#
+# ccusage's active 5-hour block, against a ceiling the owner sets. A warning
+# threshold only: the watchdog has no authority to stop a run, and would not use
+# it if it did — the supervisor's own max-spend guard is the actuator. What this
+# adds is that the guard is being watched by something outside the process it
+# guards, which is the entire premise of the layer.
+
+if command -v ccusage >/dev/null 2>&1; then
```
lilseyi, line 355: **MEDIUM — `ccusage` missing degrades silently, and the plist makes that likely.**

The preflight at L115-125 aborts loudly when `gh`/`jq`/`git` are missing, with a log line naming the launchd-PATH cause. `ccusage` gets none of that: if it isn't on PATH the whole spend check — the one the design calls page-worthy — just ceases to exist, with nothing in the log to say so.

That matters because `com.supa.fleet-watchdog.plist` L75 hardcodes `__HOME__/.nvm/versions/node/v22.22.0/bin`. `ccusage` is a node CLI, so the next node bump silently removes the spend page and the watchdog keeps reporting all-clear. "A check that silently matches nothing is worse than no check" is your own argument for deviation #2 — it applies here.

One `else log "note: ccusage not on PATH; spend check skipped"` fixes it.

scripts/watchdog.sh
```diff
@@ -0,0 +1,457 @@
...
+
+if [ "$DIGEST" = "$LAST_DIGEST" ] && [ $((NOW - LAST_SENT)) -lt $((WATCHDOG_PAGE_COOLDOWN_HOURS * 3600)) ]; then
+  log "page suppressed (unchanged, ${WATCHDOG_PAGE_COOLDOWN_HOURS}h cooldown)"
+  exit 0
+fi
+
+# launchd starts this from a bare environment — it never sources your shell
+# profile — so fall back to the keychain, which is where overnight.md § 8.4
+# tells you to put these anyway.
+: "${TELEGRAM_BOT_TOKEN:=$(security find-generic-password -a "$USER" -s TELEGRAM_BOT_TOKEN -w 2>/dev/null)}"
```
lilseyi: **LOW — `USER` is unbound under `set -u`, which kills the keychain fallback in exactly the bare environment it exists for.**

Reproduced:

```
$ env -u USER ./scripts/watchdog.sh
./scripts/watchdog.sh: line 425: USER: unbound variable
./scripts/watchdog.sh: line 426: USER: unbound variable
```

Not fatal — the failure is confined to the `$( )` subshell, `${VAR:=...}` leaves the token empty, and the run degrades to report-only (I confirmed: `no telegram creds (env or keychain); the page is in #999`, exit 0, report already filed). So the *graceful degradation* requirement is genuinely met.

But launchd job environments are not uniform, the comment two lines up is specifically about launchd's bare env, and the failure mode is "pages silently stop being delivered" plus two stderr lines into `launchd.err.log` on every page. `${USER:-$(id -un)}` costs nothing.

Worth also noting for the install runbook: `security find-generic-password` on an item created without `-T` can prompt for keychain access, and there is no `timeout(1)` on macOS to cap it. The stale-lock break bounds the damage to 30 minutes, but it's worth verifying the read is non-interactive before relying on it.

scripts/watchdog.sh
```diff
@@ -0,0 +1,457 @@
...
+# (e) GARDENER FAILURES
+# ---------------------------------------------------------------------------
+#
+# Read from workflow-run conclusions rather than by searching for gh-aw's
+# failure-report issues. The run list is exact and stable; the issue title
+# gh-aw generates is an implementation detail of a version we do not pin, and a
+# watchdog that silently matches nothing is worse than no check at all.
+
+SINCE=$(date -u -v-"${WATCHDOG_GARDENER_FAILURE_DAYS}"d +%FT%TZ)
+GARDENER_FAILS=$(gh run list --status failure --created ">=$SINCE" --limit 50 \
```
lilseyi: **LOW — `--limit 50` is applied repo-wide, *before* the `Gardener:|Watchdog` filter.**

`gh run list --status failure` is unscoped, so on a bad CI day the 50 most recent failures can all be `CI` runs and the gardener failures fall off the end — the check then reports nothing, which reads as an all-clear. That is the same silent-no-match failure mode you cite as the reason for deviation #2, arrived at from the other direction.

Measured here: 19 failures over 7 days, 112 total runs in 24h — comfortable headroom today, so this is latent fragility rather than a live bug. `--workflow=<file>` per workflow (5 cheap calls) or a much larger limit would make it exact.

Same applies to the repo half's Sweep 2 snippet (`watchdog-repo.md` L197), which has the identical shape.

Separately, tiny: the regex is `^(Gardener:|Watchdog)` — colon on one branch, not the other.

.github/workflows/watchdog-repo.md
```diff
@@ -0,0 +1,337 @@
...
+  max-bot-mentions: 1
+  # First run of the day (or if the day's issue was closed) creates it.
+  create-issue:
+    title-prefix: "[watchdog] "
+    labels: [watchdog:report]
+    max: 1
+    # 6d, matching the gardeners: the maintenance sweep that closes expired
+    # issues runs at 00:37 UTC, and 6d closes a report the morning before the
+    # equivalent-day run rather than eleven hours after it.
+    expires: 6d
```
lilseyi: **LOW — `expires: 6d` is a weekly-cadence value applied to a daily issue.**

The 6d rationale in GARDENERS.md is explicitly derived from a *weekly* gardener: "6d closes a report the morning before the equivalent-day run rather than eleven hours after it." This workflow files **one issue per day**, so the same 6d leaves up to six open `[watchdog]` issues at any moment, each a stale snapshot of a day that is already resolved.

Given "silence is the product", 1d–2d seems closer to intent. If 6d is deliberate (keeping a week of history browsable), say so here — the copied comment currently argues for a cadence this workflow doesn't have.

scripts/watchdog.sh
```diff
@@ -0,0 +1,457 @@
...
+  if [ "$lock_age" -gt $((WATCHDOG_LOCK_STALE_MINUTES * 60)) ]; then
+    log "breaking stale lock (held $(human_age "$lock_age"))"
+    rm -rf "$LOCK"
+    mkdir "$LOCK" 2>/dev/null || exit 0
+  else
+    log "skip: another run holds the lock ($(human_age "$lock_age"))"
+    exit 0
+  fi
+fi
+trap 'rm -rf "$LOCK"' EXIT
```
lilseyi: **LOW — the EXIT trap can delete a lock this run doesn't own.**

Sequence: run A wedges → run B finds the lock stale, `rm -rf`s it and `mkdir`s its own → run A finally exits and its `trap 'rm -rf "$LOCK"' EXIT` removes **B's** lock. Two runs then overlap unguarded.

Narrow window and self-healing within 30 minutes, so low. The usual fix for a `mkdir` lock is to write the owning PID into `$LOCK/pid` at acquisition and have the trap remove the directory only when it still matches `$$`.

scripts/watchdog.sh
```diff
@@ -0,0 +1,457 @@
...
+
+mkdir -p "$STATE_DIR" || exit 1
+
+NOW=$(date +%s)
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+log() { printf '[%s] %s\n' "$(date '+%F %T')" "$*" >>"$LOG"; }
```
lilseyi: **LOW — the log never rotates.**

`log()` appends unconditionally and nothing ever truncates. At 48 runs/day the growth is slow (~1MB/year), so this is not urgent — but launchd does not rotate `StandardOutPath`/`StandardErrorPath` either, and this is explicitly an install-once-and-forget job on a machine nobody will revisit. A `tail -n 5000` roll when the file exceeds some size, or a `newsyslog.d` entry mentioned in the plist header, would close it out.

(Unrelated nit while here: L136's `| grep -v "^$$\$"` is dead code — the script runs as `bash`, so its own PID can never appear in a `pgrep -x 'claude|droid|opencode'` result.)

lilseyi (commented):
Required review pass — 12 findings inline (2 high, 3 medium, 5 low, 2 nits). Verdict and the full verified/not-verified ledger in the summary comment below.

Everything I could check mechanically, I ran: `bash -n`, `shellcheck -x`, `plutil -lint` on template + rendered, `pgrep -x` alternation on this box, `gh aw compile --no-models-dev-lookup` (0 errors, `git status` clean afterwards → recompile is byte-identical), manifest diff against `gardener-cost-report.lock.yml` (identical), and five end-to-end runs of `watchdog.sh` — including a wet run of the reclaim path against a stubbed `gh`.

.github/workflows/gardener-cost-report.md
```diff
@@ -250,7 +258,8 @@ parentheses)
 | Docs Drift | 300 AIC ($3.00) | 200 AIC ($2.00) |
 | CI Doctor | 200 AIC ($2.00) | 400 AIC ($4.00) |
 | Cost Report | 300 AIC ($3.00) | 200 AIC ($2.00) |
-| **Repo-wide daily ceiling** | | **1000 AIC ($10.00)** |
+| Watchdog (repo) | 50 AIC ($0.50) | 200 AIC ($2.00) |
```
lilseyi, line 272: **NIT — the new row is right; the rows above it are now demonstrably wrong.**

Your Watchdog row (50 / 200) matches `watchdog-repo.md` L35-36, and the daily arithmetic checks out: 200+200+400+200+200 = 1200 AIC = $12.00, matching L262 and GARDENERS.md. Good.

But L257/258/260 claim **300 AIC ($3.00)** per-run for Large Files, Docs Drift and Cost Report, and all three actually set `max-ai-credits: 200`:

```
gardener-large-files.md:28:max-ai-credits: 200
gardener-docs-drift.md:25:max-ai-credits: 200
gardener-cost-report.md:24:max-ai-credits: 200
```

GARDENERS.md's table already says 200. Pre-existing, not introduced here — but this PR edits this exact table, and this workflow's own rule is "Report the cap that was configured, not the cap you wish were configured." Cheap to fix in passing.

Also worth knowing (no action needed): 4 runs/day × 50 AIC = exactly the 200 AIC daily cap, so a day where every run bills its full per-run ceiling blocks the fourth sweep.

scripts/com.supa.fleet-watchdog.plist
```diff
@@ -0,0 +1,95 @@
...
+  path in it must be absolute and literal.
+
+  Install (from the repo root):
+
+      mkdir -p ~/Library/LaunchAgents .claude/logs/watchdog
+      sed -e "s|__REPO_PATH__|$PWD|g" -e "s|__HOME__|$HOME|g" \
+        scripts/com.supa.fleet-watchdog.plist \
+        > ~/Library/LaunchAgents/com.supa.fleet-watchdog.plist
+      plutil -lint ~/Library/LaunchAgents/com.supa.fleet-watchdog.plist
+      launchctl load ~/Library/LaunchAgents/com.supa.fleet-watchdog.plist
```
lilseyi: **NIT — `launchctl load` is the deprecated spelling.**

On current macOS the supported form is `launchctl bootstrap gui/$UID <plist>` (and `bootout gui/$UID/com.supa.fleet-watchdog` to remove). `load`/`unload` still work and still emit deprecation warnings; since this header is the install runbook and the same commands are repeated in the PR description and GARDENERS.md, it's worth writing the modern form once.

Everything else in this file checks out: `plutil -lint` passes on the template **and** on the rendered output, the documented `sed` leaves zero `__REPO_PATH__`/`__HOME__` survivors, `StartInterval` (not `StartCalendarInterval`) is the right choice for a fixed-cadence sweep and does coalesce missed intervals into one fire on wake, `RunAtLoad: false` is correctly paired with the documented `launchctl start`, both log paths live under the directory the install step creates and `.gitignore` covers, and keeping the Telegram token out of the plist is exactly right.

scripts/watchdog.sh
lilseyi: Fixed in 557e51f8. `add_page` now takes an identity key and the digest is taken over the sorted keys, never the prose — `spend-runaway`, `overnight-dead:<stop-epoch>`.

**Probe — three sweeps against a fixture whose cost moves every run:**

| sweep | prose | digest | sends |
|---|---|---|---:|
| 1 | `$51.15 … after 5h46m` | `b8a7a4ff83b799b8` | 1 |
| 2 (+30s) | `$52.25 … after 5h46m` | `b8a7a4ff83b799b8` | **1** |
| 3 (new condition) | overnight run died | `c9a483118dd6ba56` | 2 |

Sweep 2's prose changed and the digest did not, so the log records
`page suppressed (unchanged, 6h cooldown)`. Sweep 3 is a genuinely new condition
and goes out immediately, which is the property worth keeping.

**Counterfactual on the old code** — hashing those two prose strings directly:

```
sweep1: 1c38de968c553d55
sweep2: da587e5b637145d0
```

Different, so it would indeed have re-paged every 30 minutes. Your read was exact.

`curl` was stubbed to a successful Telegram response so the ledger was actually
written and the cooldown exercised for real rather than inferred.

scripts/watchdog.sh
lilseyi: Fixed in 557e51f8. Reclaim now requires **all three**: no local process, no local *or remote* branch activity, and >2h issue silence.

Remote is read with `git ls-remote --heads origin` — your suggestion 1. One call
for the whole loop, and it writes nothing into the local ref store, so it still
cannot race anyone's checkout. A matching branch costs one
`gh api repos/{owner}/{repo}/commits/<branch>` to date it.

**Probe — a remote branch pushed 10 minutes ago (the cloud-agent case):**

```
=== B. remote branch pushed 10 min ago (a cloud agent) -> NO reclaim ===
  relabel calls: 0
```

No relabel and no finding at all — the agent is simply seen as alive.

Suggestions 2 and 3 are done too. `WATCHDOG_SESSION_PROCS` now carries a
keep-this-in-step warning saying `pgrep -x` is a heuristic (a launcher that execs
the versioned binary presents a different `comm`, as you found) and is blind to
non-local agents by construction. GARDENERS.md has a **"The reclaim's blind spot,
stated plainly"** section.

I did not claim the remote check closes it completely, because it doesn't: a
remote agent that has pushed nothing for 2h can still be falsely reclaimed, and
the cost is two agents on one issue. The mitigation is an agent-side convention —
comment when you claim — and the reclaim comment the watchdog leaves now tells a
reclaimed agent exactly that, so it learns why.

Everything the decision depends on also fails **closed** now (your P1 thread).

.github/GARDENERS.md
lilseyi: Fixed in 5cb798a5 — documented rather than left implicit.

GARDENERS.md gains a **"Which layer owns the reclaim"** section: the watchdog
owns it at 2h; `overnight.md` § 0.5 becomes a backstop that effectively never
fires once the launchd job is installed; **the real window is 2 hours, not 12**.
It carries the line "if you read § 0.5 in six months and believe 12h, this is the
paragraph that corrects you", and notes that § 0.5 is still the only reclaimer —
at 12h — on a host without the local watchdog. `watchdog.config.sh` repeats the
precedence next to `WATCHDOG_ORPHAN_HOURS`, since that is where someone changing
the number will be looking.

It also says *why* the watchdog can afford the tighter threshold: it runs every
30 minutes and can check whether a session is actually alive, which a
start-of-night sweep cannot.

**I have deliberately not edited `overnight.md`.** Its author has PR #709 in
flight and I was asked to leave the file alone. The line you want in § 0.5 ("if
the watchdog is installed this is a no-op backstop") is flagged in the PR body as
a coordination point so it lands with #709 instead of conflicting with it.

scripts/watchdog.config.sh
lilseyi: Fixed in 557e51f8, and you were right about both halves — the number *and* the missing gate.

One threshold split into two, because one number cannot do both jobs:

```bash
WATCHDOG_SPEND_CEILING_USD=25   # a line in the daily report
WATCHDOG_SPEND_PAGE_USD=40      # a Telegram message
```

Set against the owner's measured baseline (~$50 blocks during heavy parallel
work), so $25 is genuinely notable and $40 genuinely alarming. Your $53.46
observation is what these are calibrated against.

The structural half is fixed too: **the page only fires when no unattended run is
in flight.** While the supervisor is running it owns its own `max-spend` guard
and stops itself, so paging then is duplicate noise — and noise is precisely what
makes someone mute the bot, which silently disables the dead-run page as well.
Spend running away with nobody supervising is the case with no other backstop,
and that is the one worth waking for. Over the page threshold *during* an
overnight run degrades to a report line that says why it didn't page.

The comment now describes what the code does rather than what it aspired to.

scripts/watchdog.sh
lilseyi: Fixed in 557e51f8, both halves. You were right to turn my own deviation-#2 argument back on me.

**The silence:** a missing `ccusage` now logs `WARN: ccusage not on PATH; the
spend check did not run` **and** puts a line in the report issue — once a day,
tracked in a state file, so it says the thing without becoming its own noise.

**The cause:** the plist no longer hardcodes an nvm-versioned path. `__NODE_BIN__`
is a third placeholder resolved at install time from whatever node is on your
PATH:

```bash
sed -e "s|__REPO_PATH__|$PWD|g" \
    -e "s|__HOME__|$HOME|g" \
    -e "s|__NODE_BIN__|$(dirname "$(command -v node)")|g" \
  scripts/com.supa.fleet-watchdog.plist > ~/Library/LaunchAgents/…
```

Verified the rendered PATH actually resolves all three tools from a bare
environment:

```
$ env -i PATH="<rendered>" sh -c 'command -v ccusage && command -v gh && command -v jq'
/Users/lilseyi/.nvm/versions/node/v22.22.0/bin/ccusage
/opt/homebrew/bin/gh
/opt/homebrew/bin/jq
```

`plutil -lint` passes on template and rendered output, zero surviving
placeholders. A node bump still moves `ccusage` — but now the install step
re-resolves it and the watchdog says so if it ever goes missing anyway.

scripts/watchdog.sh
lilseyi: Fixed in 557e51f8 (script) and 50d015b8 (repo half). Both now query **per workflow** rather than filtering one repo-wide list:

```bash
for wf_file in .github/workflows/gardener-*.lock.yml \
               .github/workflows/watchdog-*.lock.yml; do
  gh run list --workflow="$(basename "$wf_file")" --status failure --created ">=$SINCE" …
done
```

The glob comes from the directory, so a sixth gardener is covered the day it
lands rather than the day someone remembers to add it to a list.

The regex is gone entirely, which also resolves your colon nit — there is nothing
left to match against. Sweep 2 of `watchdog-repo.md` got the same treatment, with
the reasoning written into the prompt so the model does not "helpfully" collapse
it back into one call.

**This immediately paid for itself.** The very next dry run against the real repo
caught a live failure the unscoped version would have been at risk of missing:

```
- **Gardener failing** — `Gardener: Large Files` failed 1× in the last 1d.
  [Latest run](https://github.com/togathernyc/togather/actions/runs/30686121841)
```

That run is the `403 Maximum consecutive cache misses exceeded (40/40)` failure —
which turned out to be a real defect in the Ollama engine config, fixed in
50d015b8 and documented in GARDENERS.md. The check found a genuine bug on its
first real outing.

scripts/watchdog.sh
lilseyi: Fixed in 557e51f8 with the nonce you suggested (a nonce rather than a bare PID, since PIDs recycle):

```bash
LOCK_NONCE="$$.$(date +%s).${RANDOM}"
…
printf '%s' "$LOCK_NONCE" >"$LOCK/nonce"
release_lock() {
  [ "$(cat "$LOCK/nonce" 2>/dev/null)" = "$LOCK_NONCE" ] && rm -rf "$LOCK"
}
trap release_lock EXIT
```

**Probe, both directions:**

```
lock ownership: a run must not delete a lock it does not own
  after a stale-break run, lock present? no (released - it owned it)
  breaking stale lock (held 13848h48m)
  foreign fresh lock still present? yes (correct)
  foreign nonce intact? foreign
```

So a run that breaks a stale lock and takes its own still releases correctly, and
a run that finds a *fresh* foreign lock skips and leaves it byte-intact — which
is the sequence in your comment.

scripts/watchdog.sh
lilseyi: Fixed in 557e51f8. `log()` rolls before appending — `tail -n "$WATCHDOG_LOG_KEEP_LINES"` once the file passes `WATCHDOG_LOG_MAX_BYTES` (1MB / 5000 lines, both configurable). Done in-script rather than via `newsyslog.d` so it needs no second install step on a job whose whole premise is install-once-and-forget.

**Probe:**

```
before: 1320000 bytes, 30000 lines
after:      22072 bytes,   501 lines
stray .roll file left behind? no
```

**And the dead code is gone** — you were right, `grep -v "^$$\$"` could never
match, because the script runs as `bash` and `pgrep -x 'claude|droid|opencode'`
matches on process name. It is now just:

```bash
SESSION_COUNT=$(pgrep -x "$WATCHDOG_SESSION_PROCS" 2>/dev/null | grep -c .)
```

launchd's own `StandardOutPath`/`StandardErrorPath` still do not rotate, but they
now only ever receive output from an abort before the log exists, which is the
right amount of unbounded.

scripts/watchdog.sh
lilseyi: Fixed in 557e51f8, exactly as you suggested:

```bash
KEYCHAIN_USER=${USER:-$(id -un)}
: "${TELEGRAM_BOT_TOKEN:=$(security find-generic-password -a "$KEYCHAIN_USER" …)}"
```

Verified against your reproduction:

```
$ env -u USER bash scripts/watchdog.sh
(no unbound-variable error)
```

The comment now says why the guard is there — a bare launchd environment is
precisely where `USER` can be unset, which is the one environment the keychain
fallback exists to serve.

On the interactive-prompt risk you flagged: real point, and there is no
`timeout(1)` to cap it. The install runbook in the plist header now uses
`launchctl kickstart` to force a first sweep, so a prompt surfaces immediately
during install rather than silently at 3am; the stale-lock break still bounds the
damage to 30 minutes if one ever appears later. Adding `-T` guidance to the
`security add-generic-password` lines is worth doing but I did not want to
document a flag combination I could not test here without writing to the owner's
real keychain.

.github/workflows/watchdog-repo.md
lilseyi: Fixed in 50d015b8 — you read the rationale correctly and it did not transfer. Now `expires: 1d`, with the comment rewritten to argue for the cadence this workflow actually has:

> 1d, NOT the gardeners' 6d. That value is derived from a *weekly* cadence — "close last week's report the morning before this week's run" — and this workflow files one issue per DAY. Copying 6d would leave six open `[watchdog]` issues standing at once, each a snapshot of a day that has already resolved itself, which is the opposite of "silence is the product".

**One consequence worth knowing about**, committed separately in a83804aa so it
is legible rather than looking like drift: `gh aw compile` derives
`agentics-maintenance.yml`'s cron from the **shortest** `expires:` across all
workflows. Dropping to 1d moved that sweep from daily (`37 0 * * *`) to every two
hours (`37 */2 * * *`) — otherwise a 1d expiry would routinely be closed a day
late. It is a plain Actions workflow with no model call, so the extra runs cost
Actions minutes and nothing else.

.github/workflows/gardener-cost-report.md
lilseyi: Fixed in 50d015b8 — all three rows now say 200, matching the frontmatter you quoted. Agreed on the principle: this workflow's own rule is "report the cap that was configured, not the cap you wish were configured", and it would have been a poor look for the first thing it says after I edited its table.

Your no-action-needed note is now *in* the file, since it is exactly the kind of
thing the report should flag if it ever happens:

> Note for the watchdog row: 4 runs/day × 50 AIC is exactly its 200 AIC daily
> cap, so a day on which every sweep bills its full per-run ceiling blocks the
> fourth. Worth flagging if you see it happen; it has not yet.

scripts/com.supa.fleet-watchdog.plist
lilseyi: Fixed in 557e51f8. The header, GARDENERS.md and the PR body all use the modern spelling now:

```bash
launchctl bootstrap gui/$UID ~/Library/LaunchAgents/com.supa.fleet-watchdog.plist
launchctl print    gui/$UID/com.supa.fleet-watchdog | head -20
launchctl kickstart gui/$UID/com.supa.fleet-watchdog     # force one sweep now
launchctl bootout  gui/$UID/com.supa.fleet-watchdog      # uninstall
```

with a one-line note that `load`/`unload`/`start` still work but warn, so anyone
who already has the old form in muscle memory knows it is not broken.

`kickstart` replacing `start` turns out to matter beyond spelling — it is what
makes a first-run keychain prompt surface during install rather than silently at
3am, which is the risk you raised on the `$USER` thread.

Thanks for confirming the rest of the file — `StartInterval` vs
`StartCalendarInterval`, the wake-coalescing behaviour, `RunAtLoad: false`, and
the token staying out of the plist. Those were the parts I was least able to
verify without installing it.

scripts/watchdog.sh
```diff
@@ -0,0 +1,457 @@
...
+
+while IFS=$'\t' read -r num title updated url; do
+  [ -n "${num:-}" ] || continue
+
+  # 1. an open PR means this issue is finished, not stalled.
+  open_pr=""
+  for p in $(gh issue view "$num" --json closedByPullRequestsReferences \
+               --jq '.closedByPullRequestsReferences[].number' 2>/dev/null); do
+    state=$(gh pr view "$p" --json state --jq '.state' 2>/dev/null)
+    [ "$state" = "OPEN" ] && open_pr="$p" && break
```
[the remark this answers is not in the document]
lilseyi: Fixed in 557e51f8 — correct, and it now fails closed exactly as you describe. Both lookups are checked, and a failure of either marks the issue unverified rather than PR-less:

```bash
if pr_refs=$(gh issue view "$num" … 2>/dev/null); then
  for p in $pr_refs; do
    if state=$(gh pr view "$p" … 2>/dev/null); then
      if [ "$state" = "OPEN" ]; then open_pr="$p"; break; fi
    else
      lookup_failed=1; break
    fi
  done
else
  lookup_failed=1
fi
```

An unverified claim is reported, not relabelled, and re-checked on the next sweep
30 minutes later:

> **Orphaned claim (unverified)** — #901 looks stale after 3h00m, but a PR or
> remote-branch lookup failed this sweep, so it was **not** reclaimed.

**Probe:**

```
=== C. linked-PR lookup FAILS -> unverified, NO reclaim (fail closed) ===
  relabel calls: 0
```

The same gate covers the new `git ls-remote` / `gh api …/commits` calls added for
the remote-agent case, since they feed the same decision. Failing closed costs a
30-minute delay; failing open costs duplicated work on an issue whose PR is
already in review.

scripts/watchdog.sh
```diff
@@ -0,0 +1,457 @@
...
+#                          The laptop is pinned awake, possibly for days.
+
+if [ -f "$OVERNIGHT_DIR/stop-epoch" ]; then
+  STOP_EPOCH=$(tr -dc '0-9' <"$OVERNIGHT_DIR/stop-epoch" 2>/dev/null)
+  if [ -n "${STOP_EPOCH:-}" ]; then
+    READY_COUNT=$(gh issue list --label "agent:ready" --state open --limit 50 \
+      --json number,labels 2>/dev/null \
+      | jq '[.[] | select([.labels[].name] | index("agent:blocked") | not)
+                 | select([.labels[].name] | index("agent:in-progress") | not)] | length' 2>/dev/null)
+    [ -n "${READY_COUNT:-}" ] || READY_COUNT=0
```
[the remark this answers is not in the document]
lilseyi: Fixed in 557e51f8. A failed query is now distinguished from a genuinely empty queue, and the unknown case **pages** rather than downgrading to a report line:

```bash
READY_COUNT=""
if ready_json=$(gh issue list --label "agent:ready" … 2>/dev/null); then
  READY_COUNT=$(printf '%s' "$ready_json" | jq '…' 2>/dev/null)
fi

if [ -z "${READY_COUNT:-}" ]; then
  add_page "overnight-dead:$STOP_EPOCH" "… The ready-queue query failed, so the
    queue depth is unknown — paging rather than assuming it was empty."
```

Exactly your reasoning: a transient API failure hiding the incident the check
exists to surface is the worst possible trade here, and an extra page costs
nothing by comparison.

Note the page key is `overnight-dead:$STOP_EPOCH` — it identifies *which run*
died, not how long ago, so the countdown in the prose cannot re-page every 30
minutes while still letting a different night's failure through immediately.

scripts/watchdog.sh
```diff
@@ -0,0 +1,457 @@
...
+    # Orphan. Report-only by design — reclaiming is a repair, not an emergency,
+    # and the queue is strictly better off afterwards. Nobody needs waking.
+    if [ "$WATCHDOG_RECLAIM" = "1" ] && [ "$DRY_RUN" != "1" ]; then
+      gh issue edit "$num" --remove-label "agent:in-progress" --add-label "agent:ready" >/dev/null 2>&1 \
+        && gh issue comment "$num" --body \
+"Reclaimed by watchdog: labelled \`agent:in-progress\` with no open PR and no branch or issue activity for $(human_age "$age"), and no agent session is running on the host. Returning it to \`agent:ready\` so a future run can pick it up.
+
+*(scripts/watchdog.sh — no code was changed.)*" >/dev/null 2>&1
+      RECLAIMED="${RECLAIMED}#$num "
+      add_finding "**Orphaned claim reclaimed** — [#$num]($url) *$title* was stuck at \`agent:in-progress\` for $(human_age "$age") with no PR and no live session. Relabelled \`agent:ready\`."
```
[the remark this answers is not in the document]
lilseyi: Fixed in 557e51f8. The reclaim is only recorded if the relabel actually succeeded, and a failure is reported as a failure:

```bash
if gh issue edit "$num" --remove-label … --add-label … >/dev/null 2>&1; then
  gh issue comment "$num" --body "…"
  RECLAIMED="${RECLAIMED}#$num "
  add_finding "**Orphaned claim reclaimed** — …"
else
  add_finding "**Reclaim FAILED** — … It is still \`agent:in-progress\` and
    still invisible to the queue. Release it by hand: \`gh issue edit …\`"
fi
```

Your description of the consequence was the deciding argument: the issue stays
out of the queue *and* the report tells the owner it was put back, so nobody ever
goes looking. The failure line now names the manual command.

**Probe:**

```
=== E. relabel itself FAILS -> reported as FAILED, not as repaired ===
  finding: - **Reclaim FAILED** — [#901](u/901) *Fix the flaky thread test* is an
             orphaned claim but `gh issue edit` failed…
```

A missing `watchdog:report` label is the most likely real cause, which the log
line at the issue-filing step already calls out by name.

scripts/watchdog.config.sh
```diff
@@ -0,0 +1,117 @@
...
+# ordinary interactive work — measured $53 in a block, 26 minutes in, with
+# nothing unusual happening.
+#
+#   CEILING — a line in the daily report. "This block is expensive."
+#   PAGE    — a Telegram message. "This is running away."
+#
+# Set against the owner's measured baseline: heavy parallel work reaches ~$50 a
+# block, so $25 is genuinely notable and $40 is genuinely alarming.
+WATCHDOG_SPEND_CEILING_USD=${WATCHDOG_SPEND_CEILING_USD:-25}
+WATCHDOG_SPEND_PAGE_USD=${WATCHDOG_SPEND_PAGE_USD:-40}
```
lilseyi: **MEDIUM — the recalibration is anchored to a number I gave you that was wrong, and $40 still pages on an ordinary working block.**

This is partly my fault, so let me correct it with the full series rather than another single sample. My `$53` last round was a **partial** block measured 26 minutes in; that same block finished at `$194.99`. Here is the actual distribution of `ccusage` block costs over the last three days on this machine:

```
2026-07-29T02:00Z    $6.48
2026-07-29T13:00Z  $186.52
2026-07-29T18:00Z  $248.65
2026-07-29T23:00Z  $313.53
2026-07-30T04:00Z  $205.59
2026-07-30T12:00Z   $38.82
2026-07-31T05:00Z   $75.47
2026-07-31T14:00Z  $144.02
2026-07-31T19:00Z  $280.77
2026-08-01T00:00Z  $100.00
2026-08-01T05:00Z  $194.99  (active)
```

**10 of 14 blocks clear $40.** The median non-zero block is ~$165 and the max is $313.53. The comment above sets the thresholds against "heavy parallel work reaches ~$50 a block" — the measured baseline is roughly **4× that**.

A real dry run against this repo, just now, on the current head:

```
## Page-worthy
- **Spend running away** — $203.06 in the active ccusage block after 1h04m,
  page threshold $40, and no unattended run is in flight to guard itself.
```

So the page fires on install, at 5× the threshold, during ordinary attended work. The 6h cooldown bounds it to ~4 Telegram messages a day rather than 48 — the HIGH is genuinely fixed — but ~4/day of "spend running away" while the owner is sitting there working is still enough to make him mute the bot, and muting it silently disables the dead-run page. That is the failure this design is organised against.

The `$25` ceiling has the milder version of the same problem: it will put a **Spend over ceiling** line in essentially every daily report, which makes that line stop carrying information.

Against the observed series, something like `$400` page / `$300` ceiling would mean what the comments say they mean — "genuinely alarming" relative to a $313 observed maximum. Worth re-measuring over a fortnight before fixing the numbers, but the current pair is not defensible against this data.

One design note while here, no change requested: I like the gate, but note what it selects for. `OVERNIGHT_ACTIVE == 0` is true exactly when the owner is at the keyboard, so the page fires **only** during attended work and never during the unattended run. Your argument — the supervisor owns `max-spend`, so the un-backstopped case is the other one — is coherent and I'm not asking you to invert it. It just means the threshold is carrying the entire burden of not being noise, which is why the number matters as much as it does.

lilseyi (commented):
Re-review at `a83804aa`. **Both HIGHs are genuinely fixed** — I re-ran my own probes rather than reading the fixture table, including the two-runs-30s-apart cooldown test and a nine-case reclaim failure matrix. Details in the summary comment.

Four new items inline (2 medium, 2 nit), one of which is merge-order and one of which is a config number I got wrong for you last round.

.github/GARDENERS.md
```diff
@@ -212,6 +218,187 @@ loud, and in the safe direction.
...
+guard degrades into a hard request cap wearing a different name, and can never
+detect the runaway loop it exists to catch.
+
+The repo watchdog therefore sets `max-turn-cache-misses: 200`, which is a
+stand-in for "off" — `0` is rejected by the schema (*must be at least 1*). The
+real per-run bounds on the Ollama path are **`max-turns`, `timeout-minutes` and
+`max-ai-credits`**, which is what the [cost caps](#reading-the-cost-caps)
+section already says binds in practice.
+
+> [!IMPORTANT]
```
lilseyi: **MEDIUM (merge order) — #713 has merged, and this block is now false. Git will not tell you, because GARDENERS.md auto-merges cleanly.**

I merged `a83804aa` onto current `origin/main` (which is `c64dc230`, #713) to check. Result:

- **One real conflict**, `.github/workflows/gardener-cost-report.lock.yml` — both PRs rewrite the `gh-aw-metadata` hash line. Resolves cleanly by taking either side and running `gh aw compile --no-models-dev-lookup`; I verified that produces a lock with zero conflict markers.
- **GARDENERS.md and `gardener-cost-report.md` auto-merge with no conflict** — and that is the problem, because the merged document then contradicts itself.

In the merged file:

- **L402** (this block) says the three Ollama gardeners *still carry 40* and must be raised to 200.
- **L653** (from #713) reads `max-turn-cache-misses: 200  # see the trap below — 5 and 40 both kill runs`.
- **L972** (from #713) says they are already at 200.

So the operator guide for a fleet with a kill switch tells the reader both things at once, and the actionable one is the stale one. This block should be **deleted** in the rebase — #713 did the work.

The section it hangs off (L354-371 here) is also now a duplicate: #713 documents the same defect, from the same run `30686121841`, at merged L657 (`[!CAUTION]`) and again as a full incident table at L938-957. Yours is well written but it is the third copy. I'd fold anything unique from it into #713's existing material and cut the rest — in particular your point that the real per-run bounds are `max-turns` / `timeout-minutes` / `max-ai-credits` is worth keeping if #713 doesn't say it as directly.

None of this is a defect in the change itself; it is purely a consequence of two PRs landing on the same doc a day apart. But since the rebase is mandatory anyway (the lock conflicts), it's the moment to do it.

scripts/watchdog.sh
```diff
@@ -0,0 +1,619 @@
...
+# Helpers
+# ---------------------------------------------------------------------------
+
+# Roll the log before appending, not after: this job is installed once on a
+# machine nobody revisits, and launchd rotates neither this file nor its own
+# StandardOut/StandardErrorPath. `wc -c` on a missing file is 0, which is the
+# first-run case.
+log() {
+  local size
+  size=$(wc -c <"$LOG" 2>/dev/null || echo 0)
```
lilseyi: **NIT — this leaks one stderr line on the first run after install.**

Reproduced on every fresh state dir, including the real dry run above:

```
./scripts/watchdog.sh: line 60: /…/watchdog.log: No such file or directory
```

Redirections are processed left to right, so `<"$LOG"` is attempted and fails **before** `2>/dev/null` is in effect, and the shell reports it on the still-unredirected stderr. `wc` never runs, so its own suppression never applies.

One line into `launchd.err.log`, once per state dir, and the `|| echo 0` fallback means the function still behaves correctly — so this is cosmetic. But it's the same class as the `USER`-unbound leak you already fixed, and it's a token reorder:

```bash
size=$(wc -c 2>/dev/null <"$LOG" || echo 0)
```

scripts/watchdog.sh
```diff
@@ -0,0 +1,619 @@
...
+# no signal left to distinguish it from an abandoned claim. The convention that
+# closes it is on the agent side, not here — an agent that claims an issue
+# should comment on it when it does, and again on any long silent stretch; that
+# moves `updatedAt` and makes it visible. Documented in GARDENERS.md § Watchdog.
+
+BRANCH_REFS=$(git for-each-ref --format='%(committerdate:unix) %(refname:short)' \
+  refs/heads refs/remotes/origin 2>/dev/null)
+
+# One network call for the whole loop rather than one per issue.
+REMOTE_REFS=$(git ls-remote --heads origin 2>/dev/null)
```
lilseyi: **NIT — `ls-remote` fires on every non-idle sweep, including when there is nothing to look up.**

Verified: with zero `agent:in-progress` issues and a live local session (the ordinary weekday shape — the gate passes on `SESSION_COUNT`, not on claims), this still makes one network round trip to `origin`. That's 48 pointless remote calls a day.

The comment above it — "one network call for the whole loop rather than one per issue" — is the right instinct; it just wants one more condition:

```bash
if [ "$IN_PROGRESS_COUNT" -gt 0 ]; then
  REMOTE_REFS=$(git ls-remote --heads origin 2>/dev/null)
  REMOTE_REFS_OK=$?
fi
```

(Keep `REMOTE_REFS_OK` initialised to `0` for the skipped case so the fail-closed check at L295 doesn't read an unset value — though with zero issues the loop never runs, so nothing reaches it either way.)

Separately, credit where due: `REMOTE_REFS_OK=$?` on the line after the assignment does capture the command substitution's status correctly, which is easy to get wrong.

.github/workflows/agentics-maintenance.yml
```diff
@@ -37,7 +37,7 @@ name: Agentic Maintenance
 
 on:
   schedule:
-    - cron: "37 0 * * *"  # Daily (based on minimum expires: 6 days)
+    - cron: "37 */2 * * *"  # Every 2 hours (based on minimum expires: 1 days)
```
lilseyi, line 40: **Assessment — acceptable, no change requested. Recording the reasoning so nobody has to re-derive it.**

This is the auto-derived consequence of `expires: 6d → 1d` on the watchdog's report issue: gh-aw sets the maintenance cron from the *minimum* expiry across all workflows, so daily became 2-hourly. I checked what that actually costs:

- **No AI.** Zero matches for `engine`, `model`, `max-ai-credits`, `awf`, or any API key in all 633 lines. The scheduled jobs are `actions/github-script` only.
- **`ubuntu-slim`**, `permissions: {}` at the top with narrow per-job grants, and every scheduled job carries `!github.event.repository.fork`.
- **This repo is public**, so Actions minutes are free. 11 extra runs/day ≈ **$0**.
- It closes expired issues; it never opens one. No notification surface.

So the cost is Actions-tab noise and nothing else — and your own per-workflow scoping of the gardener-failure check (`watchdog.sh` L457-467) makes the watchdog immune to that noise, which is a nice second-order payoff from that fix.

`expires: 1d` is the right call for a once-a-day report; 6d would have left six stale snapshots standing, exactly as your comment says.

One thing worth a line in GARDENERS.md, since this file is generated and nobody will look at it again: **this cron is repo-global and derived from the minimum expiry anywhere.** A future workflow with `expires: 1h` ratchets it for everything. Cheap to note now, annoying to rediscover later.

scripts/watchdog.config.sh
lilseyi: Fixed in d0f1fcca. Thank you for the full series — the anchor was the problem, not the arithmetic, and one sample was never going to find that.

Now **$300 ceiling / $400 page**, with your distribution and its date written into
the comment so the next person changing it can see what it was fitted to:

```
2026-07-29T02:00Z    $6.48     2026-07-30T12:00Z   $38.82
2026-07-29T13:00Z  $186.52     2026-07-31T05:00Z   $75.47
2026-07-29T18:00Z  $248.65     2026-07-31T14:00Z  $144.02
2026-07-29T23:00Z  $313.53     2026-07-31T19:00Z  $280.77
2026-07-30T04:00Z  $205.59     2026-08-01T00:00Z  $100.00
```

$300 is above every completed block but one; $400 is meaningfully past the $313.53
maximum. The comment also records *why* the earlier numbers were wrong — a partial
block sampled 26 minutes in that finished at $194.99 — so the mistake is not
repeatable, and says explicitly that this is a calibration to re-measure after a
fortnight, not a constant.

**Verified against a live block:**

```
--- actual current block cost ---
$219.43

## Findings
- **Gardener failing** — `Gardener: Large Files` failed 1× in the last 1d.
```

No page, and no `Spend over ceiling` line either — which was the milder half of
your point: a ceiling that fires in every daily report stops carrying information.

Your design note is well taken and I've left the gate as it is. You're right that
`OVERNIGHT_ACTIVE == 0` selects for attended work, so the threshold carries the
whole burden of not being noise — which is precisely why fitting it to the real
distribution rather than a vibe mattered here more than it usually would. Both
values stay config-overridable.

.github/GARDENERS.md
lilseyi: Fixed in cf410a68, and the branch is rebased onto `c64dc230` — it merges cleanly now.

**The lock conflict** resolved exactly as you described: `--theirs` plus
`gh aw compile --no-models-dev-lookup`, twice (it recurred on the second of my
commits that touches that file). Zero conflict markers, recompile byte-identical
afterwards, `git status` clean.

**The silent half.** You were right that this was the real finding — git said
nothing. Deleted:

- the `[!IMPORTANT]` block claiming the three still carry `40` — #713 did that work
- the whole `max-turn-cache-misses` section, which was the third copy
- the same explanation in `watchdog-repo.md`'s frontmatter, which was the fourth

The watchdog section now carries one sentence pointing at your canonical material,
and the workflow comment matches the wording #713 already uses in the three
gardeners. `grep -c 30686121841 GARDENERS.md` is **2** — both #713's.

**Nothing unique was lost, and one thing was wrong.** My "the real per-run bounds
are `max-turns`, `timeout-minutes` and `max-ai-credits`" line was the piece you
flagged as maybe worth keeping — but #713 establishes that `timeout-minutes` is
**inert** on the codex path (no frontmatter key reaches it; those jobs fall back to
Actions' 360-minute default). So my line would have been actively misleading
post-merge. #713 states the correct version — the AI-credit cap and `max-turns` —
in the cost-caps section. Deleted rather than folded.

Two follow-ons while in there:

- **Your `agentics-maintenance` suggestion is in**, as a `[!NOTE]` on that section:
  the cron is repo-global and derived from the minimum `expires:` anywhere, so a
  future `expires: 1h` ratchets it for everything. Exactly your reasoning that the
  file is generated and nobody will look at it again.
- The **"why 6d not 7d"** paragraph turned on that sweep being daily. Updated: the
  window it depends on is now under two hours, 6d is still correct and still safer,
  and it says so rather than silently resting on a stale premise.

The PR body's stale warning is gone too — the body is rewritten to final state.

scripts/watchdog.sh
lilseyi: Fixed in d0f1fcca — your token reorder, verbatim:

```bash
size=$(wc -c 2>/dev/null <"$LOG" || echo 0)
```

**Verified on a fresh state dir:**

```
=== NIT 1: first run on a fresh state dir must emit no stderr ===
  stderr: (empty) — clean
```

The comment now records why the order matters, since `<"$LOG" 2>/dev/null` reads
as obviously correct and isn't:

> `2>/dev/null` comes BEFORE the input redirect on purpose. Redirections are
> applied left to right, so with `<"$LOG" 2>/dev/null` the missing-file error is
> reported on a stderr that has not been silenced yet.

Agreed on the classification too — same family as the `USER`-unbound leak, and
both are the kind of thing that only ever shows up in a file nobody reads.

scripts/watchdog.sh
lilseyi: Fixed in d0f1fcca, with your initialisation note taken:

```bash
REMOTE_REFS=""
REMOTE_REFS_OK=0
if [ "$IN_PROGRESS_COUNT" -gt 0 ]; then
  REMOTE_REFS=$(git ls-remote --heads origin 2>/dev/null)
  REMOTE_REFS_OK=$?
fi
```

**Verified — zero claims, gate passed via a marker, `git` shimmed to log calls:**

```
=== NIT 2: zero claims + live session -> no ls-remote round trip ===
  ls-remote calls with 0 claims: 0
  for-each-ref calls (local, free): 1
```

The `for-each-ref` still runs; it's local and free, and hoisting it under the same
condition would buy nothing.

You're right that nothing reaches the fail-closed check with zero issues, but
`REMOTE_REFS_OK=0` is initialised anyway — a guard that depends on an unreachable
path staying unreachable is one refactor away from being wrong, and this one gates
a label mutation.

Thanks for the note on `REMOTE_REFS_OK=$?` capturing the command substitution's
status — that was deliberate and it's useful to have it confirmed by someone else
reading it.

.github/workflows/agentics-maintenance.yml
lilseyi: No change requested, but your closing suggestion is in — cf410a68 adds a `[!NOTE]` to that section of GARDENERS.md:

> **That cadence is repo-global and derived from the *minimum* `expires:` anywhere
> in the repo** — do not hand-edit the cron, `gh aw compile` rewrites it. It was
> daily at 00:37 UTC until the watchdog's report issue introduced `expires: 1d`; a
> future workflow with `expires: 1h` would ratchet it again for everything. The
> file is generated, so nobody will think to look here — which is why it is
> written down.

Your cost analysis is folded in as one line (script-only jobs, `ubuntu-slim`,
public repo ⇒ free), so the assessment doesn't have to be re-derived either.

Chasing the reference also turned up a stale premise elsewhere: the **"why 6d and
not 7d"** paragraph reasons from that sweep being daily, and the eleven-hour window
it describes is now under two hours. Updated to say so — 6d remains correct and
remains the safer number, but resting on a premise that has quietly changed is how
a doc starts lying.

And thank you for the second-order note on the per-workflow scoping. I hadn't
connected the two: the watchdog is immune to the Actions-tab noise its own
`expires` change created, because that fix queries by workflow name rather than
filtering a repo-wide list.

.github/workflows/watchdog-repo.md
```diff
@@ -0,0 +1,352 @@
...
+network:
+  allowed:
+    - defaults
+    - "ollama.com"      # the model endpoint — not in the baseline, must be named
+  blocked:
+    - python
+    - playwright
+    - containers
+
+timeout-minutes: 10
```
lilseyi, line 90: **NIT, non-blocking — this is the *fourth* codex workflow with an inert `timeout-minutes`, and #713's warning still says three.**

Verified by grep across the locks:

```
gardener-ci-doctor.lock.yml     3   (engine: claude — the real one)
gardener-cost-report.lock.yml   1
gardener-docs-drift.lock.yml    1
gardener-large-files.lock.yml   1
watchdog-repo.lock.yml          1

grep -c 'timeout-minutes: 10' watchdog-repo.lock.yml  →  0
```

So this `timeout-minutes: 10` is dropped by the codex step builder exactly like the gardeners' `15`. But the merged GARDENERS.md still says:

- L165 — "`grep -c 'timeout-minutes: 15'` returns 1 in `gardener-ci-doctor.lock.yml` and **0 in all three codex lock files**"
- L171 — "**the three Ollama gardeners'** agent jobs have no wall-clock limit of their own"
- L235 — "there is **no wall-clock backstop** on **these three**"

…and then L184, two paragraphs down, correctly says "**The five** daily caps". Adjacent paragraphs counting differently is the kind of thing that reads as a doc someone forgot to update, which undermines the surrounding warnings that are all correct.

Worth stressing that this is cosmetic: the watchdog is in fact the **best**-bounded of the five (`max-ai-credits: 50` is the tightest cap in the repo, plus `max-turns: 25`), so nothing is actually unguarded. It's a count, in a doc whose whole value is being trustworthy about what is and isn't enforced. Three → four in those three spots, or a clause folding the watchdog in.

The `max-turn-cache-misses: 200` comment right above is the right shape now — states the mechanism, cites the run, says why `0` is impossible, and stops. It no longer claims `timeout-minutes` is one of the real bounds, which was the correct thing to delete given #713 proved it inert on this engine.

lilseyi (commented):
Final pass at `cf410a68`. Everything verified — full regression matrix re-run, not re-read. One non-blocking nit inline; verdict in the summary comment.

PR #713, opened 2026-08-01
fix(gardeners): max-turn-cache-misses is a request cap on Ollama, not a cache guard

lilseyi:
Follow-up to #701 (merged). Small fix from the first **live production run** of a gardener.

## What happened

`gardener-large-files` was dispatched manually with secrets wired and `GARDENERS_ENABLED=true` — [run 30686121841](https://github.com/togathernyc/togather/actions/runs/30686121841).

| | |
|---|---|
| **Tool calling on `deepseek-v4-flash`** | ✅ **Worked** — ~40 consecutive healthy tool-executing turns, mid-scan of a 3,000-line file |
| **Outcome** | ❌ `403 Maximum consecutive cache misses exceeded (40/40)` from the gh-aw api-proxy |
| **Cause** | `max-turn-cache-misses` — not the model |
| **maxRuns / AI credits / timeout** | all held correctly |

**The open-model bet is sound.** Sustained tool calling was the one thing #701 had no evidence for, and it worked. What failed was a guard that does not mean on Ollama what it means elsewhere.

## Why the guard is wrong here

With no prompt caching, **every** request is a "consecutive miss". The counter only ever climbs, so the setting stops measuring cache health and becomes a hard ceiling on total requests. `40` was simply the number of requests the run was allowed before being cut off mid-thought.

**It cannot be disabled.** The schema enforces `minimum: 1`; `max-turn-cache-misses: -1` fails to compile with `must be at least 1 (got -1)` — verified, not assumed. So it is parked at **200** on the three Ollama gardeners, far above any realistic turn budget, with a comment on each explaining what it actually does there.

**`gardener-ci-doctor` keeps the default `5`.** Claude does cache, so the guard is meaningful there and worth keeping tight. This is deliberately not a blanket change.

## Two follow-ons from the same run

Both about the same underlying waste:

- **The prompt now insists on reading a file in one call.** The agent was paging a 3,000-line file in ~250-line slices against a **million-token** context window — spending its request budget on pagination rather than analysis.
- **`gardener-large-files` gets `max-turns: 60`** (was 30). It was still mid-scan when it died, and cost is nowhere near binding on this model — 200 AIC buys roughly 14M input tokens — so turns and the 15-minute timeout are the practical bounds. Cost caps are unchanged and remain the outer constraint.

## Scope note

`docs-drift` and `cost-report` get the same cache-miss fix but **have not run live**. GARDENERS.md says so plainly rather than implying all three are proven, and names `max-turns` as the knob if either dies partway — docs-drift makes a lot of git calls.

Also removed a stale `Narrowed from defaults` comment left stranded above the corrected `MEASURED` one in `gardener-large-files.md` — a leftover of the exact claim the previous commit disproved.

## Verification

- `gh aw compile --no-models-dev-lookup`: **4 workflows, 0 errors**, byte-identical across two consecutive runs
- `gh aw lint` (actionlint 1.7.12): no issues
- Compiled values confirmed: `maxCacheMisses: 200` on the three Ollama gardeners, **`5` on ci-doctor**, `maxRuns: 60` on large-files

## Not touched

The same setting exists in the watchdog-repo workflow on **#708** — deliberately left alone; its author is being notified separately.


.github/workflows/gardener-large-files.md
```diff
@@ -27,10 +27,19 @@ tracker-id: togather-gardener-large-files
 # Per-run is held at or below the daily slice so one run cannot blow the day's budget.
 max-ai-credits: 200        # ~$2.00 per run
 max-daily-ai-credits: 200  # ~$2.00 / 24h — this gardener's slice of the $10/day repo budget
-max-turns: 30
-# Ollama does not do prompt caching, so every turn is a cache miss. The default of
-# 5 would have the proxy block the run partway through.
-max-turn-cache-misses: 40
+# Raised from 30 after the 2026-08-01 live probe: the agent was still mid-scan of
+# a 3,000-line file when the run died. Cost is nowhere near binding here (200 AIC
+# buys ~14M input tokens on this model), so turns and the 15-minute timeout are
```
lilseyi, line 32: **[normal]** "turns and the 15-minute timeout are the practical bounds" — the 15-minute timeout does not exist on this workflow. I went looking for it because it is load-bearing in the argument for raising the cache-miss cap.

Every `timeout-minutes:` in the four compiled lock files:

| File | occurrences |
|---|---|
| `gardener-ci-doctor.lock.yml` | **15** (agent exec step), 20 (detection exec step), 45 (`safe_outputs` job) |
| `gardener-large-files.lock.yml` | 45 (`safe_outputs` job) — **that is all** |
| `gardener-docs-drift.lock.yml` | 45 (`safe_outputs` job) — **that is all** |
| `gardener-cost-report.lock.yml` | 45 (`safe_outputs` job) — **that is all** |

The `agent` job header for the three codex gardeners carries `needs`, `if`, `runs-on`, `permissions`, `concurrency`, `env`, `outputs` — and no `timeout-minutes`, at job or step level. So `timeout-minutes: 15` in the frontmatter is silently dropped for `engine: codex` in gh-aw v0.83.4; ci-doctor gets it because the claude path emits it onto the exec step. With nothing set, these three inherit the GitHub Actions default of **360 minutes**.

Practical impact is genuinely small, which is why this is normal and not high: `maxAiCredits: 200` is proxy-enforced mid-run, so there is no runaway model bill; this repo is public so Actions minutes are free; and a wedged run would at worst hold the `gh-aw-codex-*` concurrency group against a weekly cadence. The real bounds on these three are `maxRuns` (60/30/30) and the credit cap, both verified present and correct.

But this PR removes one brake partly on the reasoning that another one is holding, and on these three it isn't. Two options: drop the timeout from the sentence and say `maxRuns` + credits are the bounds, or add an explicit job-level `timeout-minutes` so the frontmatter value is real. I'd do the second and keep the sentence — a 6-hour ceiling on a gardener is not what anyone intends, even when it is free.

.github/GARDENERS.md
```diff
@@ -702,12 +713,46 @@ curl -s https://ollama.com/v1/models | python3 -c \
...
+(`gardener-large-files`, dispatched manually with secrets wired and
+`GARDENERS_ENABLED=true`,
+[run 30686121841](https://github.com/togathernyc/togather/actions/runs/30686121841)):
+
+| | |
+|---|---|
+| **Tool calling on `deepseek-v4-flash`** | **Worked.** ~40 consecutive healthy tool-executing turns, mid-scan of a 3,000-line file. |
+| **Outcome** | **Failed** — `403 Maximum consecutive cache misses exceeded (40/40)` from the gh-aw api-proxy |
+| **Cause** | `max-turn-cache-misses`, not the model |
+| **maxRuns / AI credits / timeout** | all held correctly |
```
lilseyi, line 756: **[normal]** Same as `gardener-large-files.md:32`. This row says "**maxRuns / AI credits / timeout** — all held correctly", but the three Ollama gardeners have **no timeout to hold**: the only `timeout-minutes` in each of their lock files is the `45` on the `safe_outputs` job. The `agent` job has none at any level, so it inherits Actions' 360-minute default. `maxRuns` and the AI-credit cap did hold, and those are the two that matter — the row just shouldn't credit a third.

Worth fixing here specifically because this is the run-evidence table, the part a future reader will trust most. The same claim also appears at lines 148, 169, 171, 196-197, and 745 (those are outside this diff, so I can't comment inline on them) — all six say a 15-minute timeout bounds runs that, on three of four gardeners, it does not.

lilseyi (commented):
Verified the compiled values, determinism, lint, caps and permissions against the claims. One normal finding on a bound this PR's reasoning leans on. Verdict in the summary comment.

PR #719, opened 2026-08-02
fix(gardeners): the agent cannot read GitHub with gh, and never could

lilseyi:
Fixes #715.

## Root cause — not the model, and not really the credit cap

The watchdog died on `403 Maximum AI credits exceeded (50.469 / 50)`. That is the
symptom. **The agent could not read GitHub at all, and spent its entire budget
working out why.**

Measured from [run 30697357668](https://github.com/togathernyc/togather/actions/runs/30697357668)'s
own log, in the order the agent tried them:

| Attempt | Result |
|---|---|
| `mcp__github__list_issues` | `ERROR … unsupported call: mcp__github__list_issues` |
| `gh auth status` | prints nothing — **unauthenticated** |
| `safeoutputs --help` | works, but **write-only** (8 tools, no reads) |
| `github list_issues .` | **works** — `MCP tools/call: status=200` |

gh-aw mounts each MCP server as a **CLI shim** on `PATH`
(`$RUNNER_TEMP/gh-aw/mcp-cli/bin/{github,safeoutputs}`), but the generated prompt
advertises only the `safeoutputs` shim while instructing the agent to "use GitHub
MCP tools for all GitHub reads". An agent that needs a read must discover the
`github` shim by experiment — ~10 turns of uncached context on a provider that
re-bills the whole conversation every turn. **That is where the 50 credits went.**

My prompt was written as though `gh` worked. It never did.

## The "Invalid or Unsupported Model" banner is a misclassification

The hypothesis that glm-5.2 is fatally rejected where deepseek only warned is
**falsified**. `deepseek-v4-flash` emits the byte-identical item in
[run 30686121841](https://github.com/togathernyc/togather/actions/runs/30686121841),
which **succeeded**:

```
WARN codex_models_manager::model_info: Unknown model deepseek-v4-flash is used.
{"type":"error","message":"Model metadata for `deepseek-v4-flash` not found.
 Defaulting to fallback metadata; this can degrade performance and cause issues."}
```

codex emits `type:"error"` for any model outside its built-in metadata table, and
gh-aw's failure reporter pattern-matches that into "Invalid or Unsupported Model".
**glm-5.2 ran fine for eight minutes and made real tool calls.** No model IDs are
changed here, and GARDENERS.md now says so, so nobody chases it later.

## What changed

**`watchdog-repo.md`** — all three sweeps use the shim (`list_issues`,
`issue_read`, `list_pull_requests`, `pull_request_read`, `actions_list`), plus a
"read this before anything else" block naming what does not work. Credits
**50 → 200** per run, **200 → 400** daily: 50 was a quarter of the gardeners' 200
on the reasoning that four runs a day should each cost a quarter as much, which is
wrong without prompt caching — a 25-turn run is not a quarter of a 30-turn run, it
is most of one.

**`gardener-cost-report.md`** — **fires Friday and would have failed identically**;
its prompt is entirely `gh run list` / `gh api` / `gh aw logs`. Now on the shim.
Its cost path is honest about the consequence: `gh aw logs` and `gh run download`
are unreachable here, the artifact route through the shim is **unverified**, and if
it fails the report ships with AIC marked *unavailable* rather than estimated. A
fabricated total is worse than a stated gap.

**Large Files and Docs Drift need no change** — they read the checked-out tree with
`git`/`grep`/`cat` and never touch the API. That is exactly why Large Files got far
enough to hit the request cap in #713.

**`tools.bash` is inert on `engine: codex`** — newly documented. The entries never
reach the lock file (`grep -c 'cat \*'` on any codex lock returns **0**) and the
agent runs with `--dangerously-bypass-approvals-and-sandbox`; the failing run
freely executed `gh`, `ls`, `cat` and `./bin/github`, none of them allowlisted.
Editing a `bash:` list to fix a gardener changes nothing — the prompt is the only
steering wheel. Worth knowing before someone tries exactly that.

## Validation

| Check | Result |
|---|---|
| `gh aw compile --no-models-dev-lookup` | **0 errors** |
| Recompile byte-identical | **yes** (sha256 over all 5 lock files) |
| `gh aw lint` | no issues |
| Compiled `maxAiCredits` | `200` |
| Compiled `GH_AW_MAX_DAILY_AI_CREDITS` | `"400"` |
| Old `gh issue list *` in lock | **0** occurrences |

### What can only be verified after merge

**The scheduled workflow runs from `main`, so this branch cannot be dispatched.**
Everything above is static verification. Unproven until a real run:

1. **That the shim invocations in the prompts are exactly right.** `list_issues`
   with `--param` flags is the tool's own documented form; only the stdin-JSON form
   (`printf '%s' '{…}' | github list_issues .`) is *observed* returning 200. Both
   are given, in that order, with an instruction to stop after two attempts rather
   than explore.
2. **That 200 credits is enough.** It should be generous now that the ~10 discovery
   turns are gone, but no successful sweep has ever completed.
3. **Cost Report's artifact path.** Flagged in-prompt as unverified with an
   explicit instruction to degrade to "unavailable" rather than estimate.

**Suggested merge check:** after merge, `gh workflow run watchdog-repo.lock.yml`
and read the run. If it still fails, the next thing to look at is `max-turns`, not
the model.


.github/workflows/watchdog-repo.md
```diff
@@ -153,6 +157,43 @@ and do not contradict it — if a claim is stale, describe it and move on.
...
+```
+
+The commands you need are `list_issues`, `list_pull_requests`, `issue_read`,
+`pull_request_read` and `actions_list`. Two calling forms:
+
+```bash
+# Documented form — named parameters:
+github list_issues --owner togathernyc --repo togather --state open
+
+# Verified form — a JSON object on stdin, with `.` as the argument.
```
lilseyi: **[normal]** The ordering here is backwards relative to your own evidence, and this is the one place in the PR where that costs real credits.

Line 183 leads with the **documented** form. Line 186-187 puts the **verified** form second, annotated in your own words as *"the one observed returning `MCP tools/call: status=200`"*. So the prompt tells the agent to try the form nobody has seen work, then fall back to the form that was actually observed working.

The failure being fixed is *"the agent burned its budget discovering the tooling by trial"*. Line 180 names five commands the agent will need (`list_issues`, `list_pull_requests`, `issue_read`, `pull_request_read`, `actions_list`). If the named-parameter form is the one that doesn't work, this ordering buys exactly one wasted round-trip per command — five of them — on a provider where every turn re-bills the whole conversation. That is the same class of spend that produced `50.7 / 50`, just smaller.

Swap them: lead with the stdin-JSON form, keep the named-parameter form as the fallback, and keep the stop rule at line 192 exactly as written. If the named form *is* known to work and it was the MCP-function call (`mcp__github__list_issues`) that failed, then say so — right now "documented" versus "verified" reads as an admission that only one of the two has been seen to succeed.

The two-attempt stop rule itself is the right instinct and I'd keep it verbatim — *"if both forms fail for a given command, report that sweep as 'could not run' and move on — do not go looking for another way in"* is the sentence that actually bounds this failure mode, more than the credit increase does.

lilseyi (commented):
Verified compile determinism, lint, credit values, shim wiring and the tools.bash finding. One normal on instruction ordering. Verdict in the summary comment.

.github/workflows/watchdog-repo.md
lilseyi: Both fixed in `28cdfff3`.

## The finding — shim form ordering swapped

You're right, and the arithmetic is the argument: five commands named × a form
nobody has seen work = up to five avoidable round-trips, each re-billing the whole
conversation. Same class of spend as `50.7 / 50`, just smaller.

The **VERIFIED stdin-JSON form now comes first** in both prompts, annotated as the
only form observed returning `status=200`, with the named-parameter form demoted
to *"unproven against this shim — only try this if the stdin form errors."*

**The two-attempt stop rule is unchanged, verbatim.** Agreed that it bounds this
failure mode more reliably than the credit increase does.

One thing your finding surfaced that I had to handle: the sweeps below still read
in named-parameter form because it is genuinely more legible inline. Both prompts
now say explicitly to **translate each example into the stdin form when running
it**, so the leading instruction and the examples can't pull in opposite
directions.

## The cap arithmetic — 400 → 800

This one I'm glad you did the sums on, because it would have shipped as a silent
cadence halving. `every 6h` × 400/day ⇒ a sustainable **100/run**, not 200. The
failing run reached 50.7 credits *before starting a single sweep*, so a complete
run plausibly clears 100 — at which point runs 3 and 4 get dropped by the
pre-flight guardrail and the watchdog quietly becomes twice-daily while GARDENERS.md
still says every 6h.

**A watchdog that silently halves its own cadence is exactly the failure mode it
exists to catch.** So the daily cap is now per-run × cadence, with the arithmetic
in the comment:

```yaml
# The DAILY cap has to be per-run × cadence, not something smaller. At every 6h
# that is 4 runs, so a 400 ceiling makes the sustainable per-run budget 100 — and
# the failing run reached 50.7 credits *before starting a single sweep*, so a
# complete run plausibly clears 100. […] so: 4 runs × 200 = 800.
max-ai-credits: 200        # ~$2.00 per run
max-daily-ai-credits: 800  # ~$8.00 / 24h — 4 runs × 200, so no sweep is ever skipped for budget
```

Per-run stays at 200 as you asked. Repo-wide daily ceiling moves **$14 → $18**,
updated in the GARDENERS.md table and the cost-report prompt — which now also
instructs the report to **flag any day with fewer than 4 watchdog runs**, so if the
guardrail ever does start trimming cadence, the visibility surface says so instead
of you having to notice.

## Your "worth stating out loud" note

Recorded, in all three places, because you nearly had to re-derive it:

> This entry is DECORATIVE. `tools.bash` is inert on engine:codex […] What actually
> mounts the shim is the `github:` toolsets line above — do not remove that one
> thinking this one is doing the work.

GARDENERS.md carries the contrast that makes it concrete: `Bash(` entries **0**
across all four codex locks vs **44** in ci-doctor, which is also the only one
without `--dangerously-bypass-approvals-and-sandbox`.

## Re-validated

| Check | Result |
|---|---|
| `gh aw compile --no-models-dev-lookup` | **0 errors** |
| Recompile byte-identical | **yes** (sha256 over all 5 locks) |
| `gh aw lint` | 5 checked, no issues |
| Compiled watchdog caps | `maxAiCredits: 200`, `GH_AW_MAX_DAILY_AI_CREDITS: "800"` |
| Other gardeners untouched | large-files 200, docs-drift 200, cost-report 200, ci-doctor 400 — unchanged |
| VERIFIED form precedes Documented | both prompts ✓ |
| Stop rule present, verbatim | ✓ |

Ready to merge. On the live run, the two things you flagged are still the ones to
watch: whether a complete sweep lands under 200, and whether the named-parameter
form works at all — if it does, this ordering finding evaporates and the fallback
just never fires.

PR #729, opened 2026-08-02
fix(scheduling): make runsheet dropdown selectable on desktop web

lilseyi:
## What was wrong

On the **desktop web** version of the app, the dropdowns in the runsheet (the "⋯" row-actions menu, the "When" segment pill, and the template picker) didn't respond — clicking an option did nothing, so users couldn't make a selection. (Issue #728, filed via the dev dashboard.)

## Why

All of those dropdowns are rendered by one small component, `AnchoredMenu`. It put the menu **inside** the tap-anywhere-to-close backdrop and relied on `e.stopPropagation()` to stop an option tap from also triggering the backdrop's "close" handler.

That works on phones, but on the web `react-native`'s `stopPropagation()` doesn't actually stop the browser's own click from bubbling up. So every time you clicked an option, the click also reached the backdrop and **closed the menu instead of selecting** — the dropdown looked dead.

## The fix

Rebuilt `AnchoredMenu` to match the layout the app's main modal (`CustomModal`) already uses successfully: the backdrop is now a **separate layer sitting behind** the menu rather than wrapping it. Option taps land directly on the menu rows, and only taps outside the menu hit the backdrop — no browser-specific propagation tricks needed. Behavior on native phones is unchanged.

## Verification

Because this is a browser-only event-bubbling issue (invisible to unit tests, which don't simulate DOM bubbling), I reproduced the exact `react-native-web` rendering in a real Chromium browser via Playwright and clicked an option in both the old and new structures:

- **Before** (menu nested in backdrop): clicking "Service" → *menu dismissed, no selection*.
- **After** (backdrop as sibling): clicking "Service" → *Service selected* ✓.

[Before/after: rendered mock in a real browser showing the dropdown dead vs. selectable](https://images.togather.nyc/dev-assistant/fe2711fb-5ab5-432c-967f-aad97169297a-anchored-menu-fix.png)

> ⚠️ The image is a **rendered mock** built from the real component's DOM/CSS and driven in a real browser — not a device/staging capture (this run has no simulator). It faithfully reproduces the root cause and the fix.

Also added `AnchoredMenu` unit tests guarding the interaction contract (select an option, dismiss on backdrop press, empty-option → `null`, multi-select toggle). `pnpm test` for `AnchoredMenu` passes; typecheck and eslint clean on the changed files.

## Done when

- [x] Clicking a runsheet dropdown option selects it on desktop web
- [x] Outside click still dismisses the menu
- [x] Native (phone) behavior unchanged
- [x] Tests cover the select/dismiss/toggle contract

Closes #728

---
_Generated by [Claude Code](https://claude.ai/code/session_01AKKCQpfnnvHCDC4Me8pv6c)_

### DIFF
--- a/apps/mobile/features/scheduling/components/__tests__/AnchoredMenu.test.tsx
+++ b/apps/mobile/features/scheduling/components/__tests__/AnchoredMenu.test.tsx
@@ -1,0 +1,114 @@
+import React from "react";
+import { fireEvent, render } from "@testing-library/react-native";
+import { AnchoredMenu, type AnchoredMenuOption } from "../AnchoredMenu";
+
+jest.mock("@hooks/useTheme", () => ({
+  useTheme: () => ({
+    colors: {
+      text: "#000",
+      textSecondary: "#555",
+      textTertiary: "#999",
+      surface: "#fff",
+      surfaceSecondary: "#eee",
+      border: "#ccc",
+      buttonPrimary: "#2563EB",
+    },
+  }),
+}));
+
+const ANCHOR = { x: 40, y: 120, width: 160, height: 32 };
+const OPTIONS: AnchoredMenuOption[] = [
+  { id: "pre", name: "Pre-service" },
+  { id: "service", name: "Service" },
+  { id: "post", name: "Post-service" },
+];
+
+/**
+ * Regression guard for the desktop-web bug (issue #728): the backdrop and menu
+ * must be SIBLINGS so an option press selects instead of being swallowed by the
+ * backdrop's dismiss handler. When the menu was nested inside the backdrop
+ * Pressable, react-native-web's `stopPropagation` failed to stop DOM bubbling,
+ * so every option click closed the menu instead of selecting.
+ */
+describe("AnchoredMenu", () => {
+  it("selects an option (and does not dismiss) when a row is pressed", () => {
+    const onSelect = jest.fn();
+    const onClose = jest.fn();
+    const { getByText } = render(
+      <AnchoredMenu
+        anchor={ANCHOR}
+        options={OPTIONS}
+        selectedId="pre"
+        onSelect={onSelect}
+        onClose={onClose}
+      />,
+    );
+
+    fireEvent.press(getByText("Service"));
+
+    expect(onSelect).toHaveBeenCalledTimes(1);
+    expect(onSelect).toHaveBeenCalledWith("service");
+    // The press must NOT bubble to the backdrop's dismiss handler.
+    expect(onClose).not.toHaveBeenCalled();
+  });
+
+  it("dismisses when the backdrop is pressed", () => {
+    const onSelect = jest.fn();
+    const onClose = jest.fn();
+    const { getByTestId } = render(
+      <AnchoredMenu
+        anchor={ANCHOR}
+        options={OPTIONS}
+        selectedId="pre"
+        onSelect={onSelect}
+        onClose={onClose}
+      />,
+    );
+
+    fireEvent.press(getByTestId("anchored-menu-backdrop"));
+
+    expect(onClose).toHaveBeenCalledTimes(1);
+    expect(onSelect).not.toHaveBeenCalled();
+  });
+
+  it("reports null for the empty option", () => {
+    const onSelect = jest.fn();
+    const { getByText } = render(
+      <AnchoredMenu
+        anchor={ANCHOR}
+        options={OPTIONS}
+        selectedId="service"
+        emptyOption={{ label: "No segment" }}
+        onSelect={onSelect}
+        onClose={jest.fn()}
+      />,
+    );
+
+    fireEvent.press(getByText("No segment"));
+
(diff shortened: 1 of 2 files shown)


apps/mobile/features/scheduling/components/__tests__/AnchoredMenu.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+        anchor={ANCHOR}
+        options={OPTIONS}
+        selectedId="pre"
+        onSelect={onSelect}
+        onClose={onClose}
+      />,
+    );
+
+    fireEvent.press(getByText("Service"));
+
```
lilseyi, line 76: **This test does not actually guard the #728 regression** (it would pass on the un-fixed code too).

The bug is a **DOM event-bubbling** defect: on `react-native-web` a real click on an option row bubbles up the DOM to the backdrop's `onClose`, and RN's `e.stopPropagation()` fails to stop it. This test environment can't reproduce that:

- The `jest-expo` preset renders through `react-test-renderer`, not jsdom / react-native-web — there is no DOM and no DOM bubbling.
- `@testing-library/react-native` v12's `fireEvent.press(getByText("Service"))` invokes only the **nearest** ancestor `onPress` and returns; it does not dispatch a real event or call additional ancestor handlers.

Trace on the **old, buggy** nested structure (`<Pressable backdrop onPress={onClose}><Pressable menu onPress={stopPropagation}>…<row onPress={()=>onSelect(id)}>`): `fireEvent.press` finds the row's `onPress` → `onSelect("service")` fires and stops. The `stopPropagation` and backdrop `onClose` handlers are never reached. So `expect(onSelect).toHaveBeenCalledWith("service")` **and** `expect(onClose).not.toHaveBeenCalled()` both pass on the buggy code — the assertion at line 48 (and the identical one at line 112) is vacuous and can never fail from re-nesting.

The doc comment (lines 26–32) and the PR body present this as the "regression guard for #728," which overstates it. To actually guard the regression, either (a) assert on the rendered tree that the backdrop `Pressable` is **not an ancestor** of the option rows (a structural sibling check), or (b) add a Playwright/web-DOM test that issues a real click. Otherwise the desktop-web bug could silently return with this suite still green.

---

lilseyi (commented):
**Automated review — verdict: changes requested** (routine sessions can't post a formal request-changes review; the verdict travels to the dashboard via callback).

The **fix itself is correct and well-scoped.** Reviewed across four lenses (correctness, security/native-safety, spec-fidelity/UX, tests):

- **Correctness ✓** — The sibling-backdrop restructure is sound. Backdrop (`absoluteFillObject`) paints first, the absolutely-positioned menu second, so on both web and native option presses land on the rows and only outside presses reach the backdrop's `onClose`. No `stopPropagation` needed. Props contract unchanged; all callers verified.
- **Spec-fidelity ✓** — All three dropdowns named in #728 route through this single `AnchoredMenu`: the "⋯" row-actions menu (`RunSheetScreen.tsx`), the "When" segment pill (`RunSheetTemplateEditorScreen.tsx`), and the template picker (`PlanTemplateToolbar.tsx`). One fix covers all three; scope is exact, no creep.
- **Security / native-safety ✓** — No dependency, lockfile, `runtimeVersion`, or native-view changes; pure react-native primitives. Clean.

**Blocking finding (tests):** the new `AnchoredMenu.test.tsx` is presented (doc comment + PR body) as the "regression guard for #728," but it does **not** guard that regression — the key assertions would pass on the un-fixed nested code too. Details are in the inline comment. The bug is a `react-native-web` DOM-bubbling defect, and `jest-expo` (react-test-renderer, no DOM) + `fireEvent.press` (nearest handler only, no bubbling) structurally cannot reproduce it. Given this repo's own warning that these web-vs-native bugs are *invisible to unit tests*, a suite that stays green while silently claiming to protect against #728 is worth fixing before merge — either a structural "backdrop is not an ancestor of the rows" assertion, or a Playwright/web-DOM click test.

_Non-blocking:_ the empty-options branch (`AnchoredMenu.tsx`, "No options available.") is untested — cheap to add while you're here.

---

apps/mobile/features/scheduling/components/__tests__/AnchoredMenu.test.tsx
lilseyi: Fixed in 573aa5b — you're right, the behavioral assertions were vacuous as a regression guard. Went with your option (a): a structural test that asserts the option rows are **not descendants** of the backdrop `Pressable` (`within(getByTestId("anchored-menu-backdrop")).queryByText("Service")` must be `null`), which is exactly the condition the #728 nesting violates.

Verified it actually guards the regression: I temporarily re-nested the menu inside the backdrop in `AnchoredMenu.tsx` and re-ran — the new structural test **fails** while the four behavioral tests all stay green (confirming they never protected against re-nesting), then reverted the source. Also corrected the doc comment so it no longer overstates the behavioral tests as the #728 guard and points to the structural test instead.

---

PR #741, opened 2026-08-02
fix(runsheet): expand rows on web instead of clipping to two lines

lilseyi:
## Problem

Tapping a run sheet row's chevron on web flipped the arrow but revealed no more text than the collapsed two-line preview, so the dropdowns looked dead ("these dropdown arrows don't work on web for run sheets").

The press was firing all along. The bug is in what the expanded state renders.

## Root cause

Expanded descriptions and notes render through `SelectableText` (`apps/mobile/features/leader-tools/utils/runSheetLinks.tsx`), which used a multiline `TextInput` for reliable iOS text selection. react-native-web turns that into a `<textarea>` and **ignores `scrollEnabled={false}`**, so it kept its default two-row height and clipped everything below.

Measured in Chromium on an eight-line note:

```
clientHeight: 32   scrollHeight: 160   rows: 2   clipped: true
```

Native is unaffected — there the `TextInput` auto-sizes, which is why this only shows up in the browser.

## Fix

On web, render a plain selectable `<Text>` — DOM text is already selectable and `<Text>` grows with its content. The `TextInput` stays on native. This also unclips the PCO run sheet's role notes, which share the same helper.

## Verification

Real Chromium at iPhone-13 width against the actual component. Both rows in the comparison are in the **expanded** state (chevron up): before, an eight-line note is still clipped to two lines; after, the full note renders.

- Added `apps/mobile/app/ui-test/run-sheet-expand.tsx`, a fixture harness (same pattern as `/ui-test/attendance`) so the rows can be driven without a Convex backend. To get there, the item list — segment labels, collapsible headers, expandable rows, "Mine" empty state — was split out of `PlanRunSheet` into a presentational `RunSheetItemList`, with the expand/collapse state in a `useRunSheetExpansion` hook that the data-fetching wrapper owns (see "Review findings" below for why that placement matters).
- Added `tests/e2e/run-sheet-expand.spec.ts`: clicks the chevron and asserts no `<textarea>`, and that the full note is present as real, unclipped DOM text. **Confirmed failing on the pre-fix renderer and passing after** — the `textarea` count, `getByText(/Line 8:/)` and `toContainText` assertions are the three that fail pre-fix.
- `tsc --noEmit` clean on the touched files, `eslint` clean, 458 jest tests under `features/leader-tools` + `features/scheduling` pass.

## Review findings (both fixed)

Two independent review passes ran over the diff; threads are inline and resolved.

1. **The extraction regressed offline/reconnect behaviour.** Putting the expand/collapse state in `RunSheetItemList` tied it to that component's mount, and the list only renders *past* `PlanRunSheet`'s loading/empty early returns — so any bounce back to loading (an offline→online flip or an auth-token refresh both return the Convex query to `undefined` for a beat) silently collapsed every row the user had opened. It also moved the AsyncStorage restore from "parallel with the query, behind the spinner" to "racing the user's first tap on a live list". Fixed in 426c099 by hoisting both into `useRunSheetExpansion`, called above the early returns.
2. **A flaky exact-pixel assertion in the spec.** The re-collapse check pinned `clientHeight` to a baseline measured before the Ionicons web font resolved, a ~1px difference that could fail on a cold CI runner with retries masking it as a flake. Fixed in 91037cc — re-measures and asserts the row is shorter than its expanded height.

## Known limitation

For notes short enough to fit the two-line preview, expanding still looks near-identical by design — the preview already showed everything. Hiding the chevron when nothing is truncated (or shortening the preview) is a separate design call, not included here.

Also noted during review, pre-existing and not introduced here: the `ui-test` route group has no `__DEV__` guard or build-time filter, so these harness routes are reachable on production web. This is the third one; gating the group is worth doing once, separately.

### DIFF
--- a/apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
+++ b/apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
@@ -425,6 +425,11 @@ export function PlanRunSheet({
     [effItems],
   );
 
+  // Owned here, above the loading/empty early returns below, so a transient
+  // `items === undefined` doesn't wipe what the user has expanded — see the
+  // hook's note.
+  const expansion = useRunSheetExpansion(groupId);
+
   // Live "current item": match the item whose computed [start, start +
   // durationSec) window contains `now` (ticked above). Because `clockTimes` is
   // anchored to the active/selected service, this highlights the right rows on
@@ -570,7 +575,7 @@ export function PlanRunSheet({
             hasAnyViewerRole={hasAnyViewerRole}
             mineHasContent={mineHasContent}
             currentItemId={currentItemId}
-            groupId={groupId}
+            expansion={expansion}
           />
         </>
       )}
@@ -578,39 +583,31 @@ export function PlanRunSheet({
   );
 }
 
-/**
- * The run sheet's item list: segment labels, collapsible headers and expandable
- * rows, plus the "Mine" empty state.
- *
- * Split out of `PlanRunSheet` so the expand/collapse behaviour can be rendered
- * from fixtures without a Convex backend (`/ui-test/run-sheet-expand`, driven by
- * `tests/e2e/run-sheet-expand.spec.ts`) — the data-fetching wrapper stays above.
- */
-export function RunSheetItemList({
-  itemsBySegment,
-  clockTimes,
-  peopleByRole,
-  viewMode,
-  viewerRoleIds,
-  hasAnyViewerRole,
-  mineHasContent,
-  currentItemId,
-  groupId,
-}: {
-  itemsBySegment: Record<Segment, RunSheetItem[]>;
-  clockTimes: Record<string, number | null>;
-  peopleByRole: Record<string, string[]>;
-  viewMode: ViewMode;
-  viewerRoleIds: Set<string>;
-  hasAnyViewerRole: boolean;
-  mineHasContent: boolean;
-  currentItemId: string | null;
-  /** Scopes the persisted collapsed-header state. */
-  groupId: Id<"groups">;
-}) {
-  const { colors, isDark } = useTheme();
+export type RunSheetExpansion = {
+  /** Item ids whose description/notes are revealed. */
+  expandedItemIds: Set<string>;
+  toggleItemExpanded: (itemId: string) => void;
+  /** Header ids whose following rows are hidden. */
+  collapsedHeaders: Set<string>;
+  toggleHeaderCollapsed: (headerId: string) => void;
+};
 
-  // Expandable rows: which items have their description/notes revealed.
+/**
+ * Which run sheet rows are expanded, and which section headers are collapsed.
+ *
+ * This MUST be owned by the data-fetching wrapper, not by `RunSheetItemList` —
+ * the list only renders once items have arrived, so holding the state there
+ * would tie it to the list's mount. Any bounce back to the loading branch (an
+ * offline→online flip or an auth-token refresh both return the Convex query to
+ * `undefined` for a beat) would then unmount it and silently collapse every row
+ * the user had opened. Owning it above the loading branch also means the
+ * AsyncStorage restore below starts in parallel with the query, while the
+ * spinner is up, instead of racing the user's first tap on a live list.
+ *
+ * Collapsed headers are persisted per group under a native-specific key so they
+ * never collide with the PCO viewer's collapse state.
+ */
+export function useRunSheetExpansion(groupId: Id<"groups">): RunSheetExpansion {
   const [expandedItemIds, setExpandedItemIds] = useState<Set<string>>(
     () => new Set(),
   );
@@ -623,9 +620,6 @@ export function RunSheetItemList({
     });
   }, []);
 
-  // Collapsible sections: header ids whose following items are hidden.
-  // Persisted to AsyncStorage per group (native-specific key so it never
-  // collides with the PCO viewer's collapse state) and restored on reopen.
   const [collapsedHeaders, setCollapsedHeaders] = useState<Set<string>>(
     () => new Set(),
   );
@@ -667,6 +661,53 @@ export function RunSheetItemList({
       JSON.stringify(Array.from(collapsedHeaders)),
     ).catch((err) => console.error("Failed to save collapsed state:", err));
   }, [collapsedHeaders, collapsedHeadersLoaded, collapsedStorageKey]);
+
+  return {
+    expandedItemIds,
+    toggleItemExpanded,
+    collapsedHeaders,
+    toggleHeaderCollapsed,
+  };
+}
+
+/**
+ * The run sheet's item list: segment labels, collapsible headers and expandable
+ * rows, plus the "Mine" empty state. Presentational — its expand/collapse state
+ * comes in from `useRunSheetExpansion`.
+ *
+ * Split out of `PlanRunSheet` so the expand/collapse behaviour can be rendered
+ * from fixtures without a Convex backend (`/ui-test/run-sheet-expand`, driven by
+ * `tests/e2e/run-sheet-expand.spec.ts`) — the data-fetching wrapper stays above.
+ */
+export function RunSheetItemList({
+  itemsBySegment,
+  clockTimes,
+  peopleByRole,
+  viewMode,
+  viewerRoleIds,
+  hasAnyViewerRole,
+  mineHasContent,
+  currentItemId,
+  expansion,
+}: {
+  itemsBySegment: Record<Segment, RunSheetItem[]>;
+  clockTimes: Record<string, number | null>;
+  peopleByRole: Record<string, string[]>;
+  viewMode: ViewMode;
+  viewerRoleIds: Set<string>;
+  hasAnyViewerRole: boolean;
+  mineHasContent: boolean;
+  currentItemId: string | null;
+  /** From `useRunSheetExpansion`, owned by the caller — see that hook's note. */
+  expansion: RunSheetExpansion;
+}) {
+  const { colors, isDark } = useTheme();
+  const {
+    expandedItemIds,
+    toggleItemExpanded,
+    collapsedHeaders,
+    toggleHeaderCollapsed,
+  } = expansion;
 
   if (viewMode === "mine" && !mineHasContent) {
     return (
--- a/apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
+++ b/apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
@@ -60,7 +60,7 @@ const CURRENT_ITEM_BG_DARK = "#2a2700";
 const CURRENT_ITEM_BORDER = "#D4A017";
 
 /** When an item happens relative to the event's service times. */
-type Segment = "before" | "during" | "after";
+export type Segment = "before" | "during" | "after";
 const SEGMENT_OPTIONS: Array<{ key: Segment; label: string }> = [
   { key: "before", label: "Before event" },
   { key: "during", label: "During event" },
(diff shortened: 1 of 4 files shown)


apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
```diff
@@ -619,65 +561,174 @@ export function PlanRunSheet({
...
+  hasAnyViewerRole: boolean;
+  mineHasContent: boolean;
+  currentItemId: string | null;
+  /** Scopes the persisted collapsed-header state. */
+  groupId: Id<"groups">;
+}) {
+  const { colors, isDark } = useTheme();
+
+  // Expandable rows: which items have their description/notes revealed.
+  const [expandedItemIds, setExpandedItemIds] = useState<Set<string>>(
```
lilseyi, line 611: **Correctness (must fix): the refactor moved expand/collapse state behind an early return, so a transient loading blip now wipes it.**

`expandedItemIds` and `collapsedHeaders` used to live in `PlanRunSheet`, which stays mounted through its loading / unavailable / empty branches. They now live in `RunSheetItemList`, which is only rendered *past* those returns (line 448, line 523) — so any bounce back to loading unmounts the component and resets the user's expanded rows.

Concrete scenario, on an explicitly supported path here: cold-launch offline, `items === undefined`, so `effItems` comes from `getItemsStale(planId)` and the sheet renders from cache. The user expands three rows to read notes. Signal returns → `isNetworkAvailable` flips true while the re-resolving Convex subscription still has `items === undefined`, so `effItems` becomes `undefined` → line 448 returns the spinner → `RunSheetItemList` unmounts. When `items` lands, every row is collapsed again. Before this PR the state survived that flicker.

Same mechanism on any other `items → undefined` blip (an auth-token refresh re-keys the subscription), and on `effItems.length` going `n → 0 → n`.

Fix: keep the state in `PlanRunSheet` — above the early returns — and pass it down, leaving `RunSheetItemList` presentational.

---

apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
```diff
@@ -619,65 +561,174 @@ export function PlanRunSheet({
...
+      return next;
+    });
+  }, []);
+
+  const collapsedStorageKey = `native_runsheet_collapsed_${groupId}`;
+  useEffect(() => {
+    let cancelled = false;
+    (async () => {
+      try {
+        const saved = await AsyncStorage.getItem(collapsedStorageKey);
```
lilseyi, line 641: **Correctness (must fix, same root cause as the comment above): the collapsed-header restore now races the user instead of the network.**

This AsyncStorage read used to start when `PlanRunSheet` mounted — in parallel with the Convex query, while the spinner was up and nothing was tappable. It now starts when `RunSheetItemList` mounts, i.e. *after* the data has arrived and the list is on screen and interactive. Two consequences:

1. **Layout jump.** First paint always has `collapsedHeaders = new Set()`, so `filterVisible` renders every row under a previously-collapsed header, then hides them a tick later. Previously the AsyncStorage read almost always beat the Convex round-trip, so first paint was already correct.
2. **Silently dropped tap.** The load does an unconditional `setCollapsedHeaders(new Set(JSON.parse(saved)))` — an overwrite, not a merge — and the persist effect is gated on `collapsedHeadersLoaded`. A header chevron tapped in the window between first paint and the read resolving is overwritten and never persisted. That window used to sit entirely behind the loading spinner, so it was unreachable.

Hoisting the state back into `PlanRunSheet` fixes both.

---

lilseyi (commented):
## Correctness review

Reviewed the diff adversarially, focusing on the `PlanRunSheet` → `RunSheetItemList` extraction, since code motion is where regressions hide.

**Two findings, both must-fix, both the same root cause:** the expand/collapse state moved from a component that survives the loading/empty branches into one that is only mounted past them. Details inline.

### Checked and clean

- **Prop threading.** All nine props passed to `RunSheetItemList` are the exact values the inline JSX used. Diffed the old block against the new render: segment filtering, the `segmentHasContentForViewer` skip, `filterVisible`, and every `ReadOnlyRow` prop are identical apart from re-indentation.
- **`isDark` / `colors`.** `RunSheetItemList` calls `useTheme()` itself; dropping `isDark` from `PlanRunSheet` is correct — it's now only referenced inside the extracted component and `ReadOnlyRow`.
- **Rules of Hooks.** All seven hooks sit above the `viewMode === "mine" && !mineHasContent` early return. No conditional hooks; the effects still run in the "Mine" empty state as before.
- **AsyncStorage persistence.** Key string, `cancelled` flag, `finally { setCollapsedHeadersLoaded(true) }`, and the `if (!collapsedHeadersLoaded) return` persist guard are copied verbatim. (The pre-existing shared-key collision between stacked `PlanRunSheet`s in serving mode is unchanged by this PR.)
- **Component identity.** `RunSheetItemList` is a top-level export, so no per-render remount; `key={activePlanId}` on `PlanRunSheet` still resets state on plan-tab switch exactly as before.
- **The `SelectableText` fix itself.** No hooks in the component, so the `Platform.OS === "web"` early return is safe. The `padding: 0 / margin: 0 / transparent` resets are genuinely unnecessary on `Text` — react-native-web's `Text` base style already zeroes margin/padding, sets no background, and uses `whiteSpace: 'pre-wrap'`, so multi-line notes keep their newlines.
- **Other `SelectableText` consumers** (`RunSheetScreen.tsx` role notes): the styles passed are plain `{ fontSize, lineHeight, color }` — nothing `TextInput`-specific is lost. Accessibility improves, since a read-only `<textarea>` was tabbable and exposed as a textbox.

---

tests/e2e/run-sheet-expand.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    // And the row genuinely grew rather than swapping one clipped box for
+    // another (kept relative — how many lines the note wraps to is viewport
+    // dependent).
+    const expandedHeight = await row.evaluate((el) => el.clientHeight);
+    expect(expandedHeight).toBeGreaterThan(collapsedHeight);
+
+    // And it collapses again.
+    await row.getByRole("button", { name: /Long note/ }).click();
+    await expect(row).toHaveJSProperty("clientHeight", collapsedHeight);
```
lilseyi: **Test flakiness (should fix): exact pixel equality against a height captured before layout settles.**

`collapsedHeight` is captured on line 27, immediately after the harness becomes visible — no font wait, no stable-box check. This line then asserts the post-collapse `clientHeight` equals that number exactly.

The row's height is `max(title line box @15px, chevron line box @16px)`. Ionicons is injected as a web `@font-face` by `@expo/vector-icons`; until it resolves the chevron's line box uses fallback metrics (~18–19px), and after it resolves it uses Ionicons' own (~16px), at which point the 15px title (~17px) becomes the max. That's a ~1px shift in row height. If line 27's measurement lands before the font resolves — more likely on a cold CI runner than locally — this assertion can never match and fails after the expect timeout. `retries: 2` in CI would mask it as a flake rather than a hard failure.

Also worth noting for the record, since the comment above it implies otherwise: this assertion and the `expandedHeight > collapsedHeight` one both stay **green on the buggy renderer** (removing the textarea restores the original preview layout either way, and the pre-fix expanded state was ~19px taller thanks to the category line). The real guards in this spec are the `textarea` count, the `getByText(/Line 8:/)` visibility, and the `toContainText` — each of which does fail pre-fix.

Fix: re-measure and assert `toBeLessThan(expandedHeight)`. That still catches "collapse didn't restore layout" without pinning an exact pixel value.

---

lilseyi (commented):
## Test-quality & spec-fidelity review

Second pass, independent of the correctness review above. **It independently reached the same finding #1** (expand/collapse state moved below `PlanRunSheet`'s loading early-returns), by a different route — worth noting that `collapsedHeaders` partly self-heals there because it's persisted to AsyncStorage, while `expandedItemIds` is not, so that one is silently lost. One new finding inline.

### Does the spec actually fail pre-fix?

Verified against `react-native-web@0.21.2` source: `TextInput/index.js:323` picks `'textarea'` when `multiline`, and `:377` sets `rows = multiline ? (rows ?? numberOfLines) : 1` — with neither prop set, `rows` is `undefined`, so the browser default of **2 rows** applies. The PR's bug description is accurate.

Per assertion:

| Assertion | Pre-fix |
| --- | --- |
| `expect(row.locator("textarea")).toHaveCount(0)` | **FAILS** — real guard, directly encodes the regression |
| `getByText(/Line 8:/)` visible | **FAILS** — React puts the text in the `<textarea value>` property, not a DOM text node |
| `toContainText("Line 8: …")` | **FAILS** — same reason |
| `expandedHeight > collapsedHeight` | passes pre-fix — not a guard |
| exact `clientHeight` equality after re-collapse | passes pre-fix — not a guard |

Three genuine guards; the two height assertions document behaviour rather than catching the bug.

### Harness fidelity

Faithful. Every prop matches the production `PlanRunSheet` call site in type and semantics, and it exercises the real path end to end (`ReadOnlyRow` → `renderTextWithLinks` → `SelectableText`), with URL-free fixture text so `SelectableText` is genuinely rendered rather than replaced by a link preview. Nothing papered over.

### Other checks

- **CI picks it up.** `detect-changes`'s `e2e` filter matches both `apps/mobile/**` and `tests/e2e/**`; `test-e2e` runs `pnpm test:e2e` → `testDir: ./tests/e2e`. Confirmed by running it.
- **Unit coverage.** `jest-expo` forces `caller.platform: 'ios'`, so `Platform.OS` is never `"web"` in jest and the branch can't be covered there naturally. It's *technically* pinnable by overriding `Platform.OS`, but jest renders RN primitives rather than react-native-web, so it could never catch the actual clipping. Not worth blocking on — the e2e is the only thing that can.
- **Harness storage key.** `native_runsheet_collapsed_…` scoped to a harness-only id; real keys interpolate a Convex `Id<"groups">`, so no collision, and the PCO viewer uses a different prefix entirely.

### Informational, not a finding

`/ui-test/run-sheet-expand` will be publicly reachable on production web — there's no `__DEV__` guard or build-time filter for the `ui-test` route group, and `apps/link-preview/cloudflare-worker.js` explicitly whitelists `ui-test` as a known app route. That's the existing precedent (`ui-test/attendance` and `ui-test/event-rsvp-refresh` ship today), not something this PR introduces. Flagging because this is now the third one and gating the whole group is probably worth doing once, separately.

---

apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
lilseyi: Fixed in 426c099. Both `expandedItemIds` and `collapsedHeaders` now live in `useRunSheetExpansion(groupId)`, called by `PlanRunSheet` above its early returns, and come into `RunSheetItemList` as an `expansion` prop — the list is presentational again. The hook's docstring names this failure mode so the state doesn't drift back down.

Side benefit: the harness no longer needs the `groupId={… as never}` cast, since it just calls the same hook.

---

apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
lilseyi: Fixed in 426c099, same change as the thread above. The AsyncStorage load now runs from `useRunSheetExpansion`, which `PlanRunSheet` calls before its early returns — so the read is back in parallel with the Convex query, behind the spinner, and the window where a header tap could be overwritten by the saved set is unreachable again.

---

tests/e2e/run-sheet-expand.spec.ts
lilseyi: Fixed in 91037cc — re-measures and asserts `toBeLessThan(expandedHeight)` via `expect.poll`, so nothing is pinned to a pixel value. Also rewrote the comment above it, which was overclaiming: it now says the height check guards "collapse restored the layout", not the bug itself.

The point about which assertions are the real guards is right and worth keeping on the record: the `textarea` count, `getByText(/Line 8:/)`, and `toContainText` are the three that fail pre-fix. I re-confirmed that empirically before and after this change by reverting the `Platform.OS === "web"` branch — the spec fails on the `textarea` assertion and passes once restored.

---

PR #773, opened 2026-08-09
feat(rostering): song title on run sheet rows + leader manual confirm

lilseyi:
Two roster/run-sheet fixes requested by David Walker Jr.:

## 1. Song imports are now visible on run sheet segments

Importing a song into a run sheet segment was invisible in the native run sheet view: the row kept its segment title (often the default "New item"), and the subtext only rendered a per-service key/BPM *override* — ignoring both the song's title and its library defaults, so a linked song with no override showed nothing at all.

- New `songMetaLine()` helper next to the existing `effectiveKey`/`effectiveBpm` adapters: leads with the linked song's title (skipped when the row title already matches) and resolves Key/BPM through the ADR-027 override layering. Rows now read e.g. **"Way Maker · Key G · 68 BPM"**.
- `NativeRunSheetView`'s local item type now declares the joined `song` the `listItems` query has been sending all along (optional — pre-existing offline caches rehydrate without it).
- Covers both the leader-tools run sheet and the serving run sheet (shared component).

## 2. Leaders/managers can manually confirm a roster assignment

Volunteers often say yes verbally or over text without tapping their confirm request, leaving the slot amber and reminder nudges firing. The only status-writing path (`respondToAssignment`) is strictly self-service.

- New `confirmAssignment` mutation, gated by `requirePlanTeamScheduler` (same team-scoped gate as `unassign`): group leaders, community admins, and that team's managers can mark someone confirmed. Works on a declined row too (clears the decline note); schedules the same team-channel reconciles as a self-response; deliberately skips `notifyLeadersOfResponse`.
- **Confirm** action on the roster grid's role and member cell popovers for any occupant not yet confirmed.
- Event Plans guide updated (roster grid section) per the guides-sync rule.

## Tests

- 7 new unit tests for `songMetaLine` (17 passing in the suite).
- 4 new backend tests for `confirmAssignment` (leader can confirm, non-scheduler rejected, already-confirmed no-op, declined → confirmed clears note); 39 passing across the assignments/permissions suites.
- Typecheck clean for all touched files (remaining mobile errors are pre-existing on main, verified against the clean tree).

https://claude.ai/code/session_01J75sjErevGEvdzBAKCFLx8

---
_Generated by [Claude Code](https://claude.ai/code/session_01J75sjErevGEvdzBAKCFLx8)_

### DIFF
--- a/apps/convex/functions/scheduling/assignments.ts
+++ b/apps/convex/functions/scheduling/assignments.ts
@@ -875,6 +875,71 @@ export const respondToAssignment = mutation({
 });
 
 /**
+ * Leader/manager manual confirm — mark someone ELSE's assignment `confirmed`.
+ *
+ * Why: volunteers often confirm verbally or in person ("yes, I'll be there
+ * Sunday") without ever tapping the request. This lets the scheduler record
+ * that answer on the roster, including overriding a `declined` row (the
+ * decline note is cleared) when the volunteer changed their mind offline.
+ * Deliberately does NOT notify leaders (`notifyLeadersOfResponse`) — the
+ * leader is the one acting.
+ *
+ * Auth: same gate as `unassign` — `requirePlanTeamScheduler`, scoped to the
+ * assignment's own team.
+ */
+export const confirmAssignment = mutation({
+  args: {
+    token: v.string(),
+    assignmentId: v.id("roleAssignments"),
+  },
+  handler: async (ctx, args) => {
+    const callerId = await requireAuth(ctx, args.token);
+
+    const assignment = await ctx.db.get(args.assignmentId);
+    if (!assignment) {
+      throw new ConvexError("Assignment not found");
+    }
+    // Scoped to the assignment's own team, so a manager can confirm their
+    // team's cells but not another team's.
+    await requirePlanTeamScheduler(
+      ctx,
+      assignment.planId,
+      assignment.teamId,
+      callerId,
+    );
+
+    // Already confirmed — nothing to record.
+    if (assignment.status === "confirmed") {
+      return { assignmentId: args.assignmentId, status: "confirmed" as const };
+    }
+
+    await ctx.db.patch(args.assignmentId, {
+      status: "confirmed",
+      declineNote: undefined,
+      respondedAt: Date.now(),
+    });
+
+    // Same uniform channel-sync trigger as `respondToAssignment`: a confirm
+    // has no membership effect on the team channel itself, but a declined →
+    // confirmed override can re-add the user to derived membership, and
+    // reconciling on every status change keeps the trigger cheap and uniform.
+    await ctx.scheduler.runAfter(
+      0,
+      internal.functions.scheduling.teamChannelSync.reconcileTeamChannel,
+      { teamId: assignment.teamId },
+    );
+    await ctx.scheduler.runAfter(
+      0,
+      internal.functions.scheduling.teamChannelSync
+        .reconcileCrossTeamChannelsForSource,
+      { sourceTeamId: assignment.teamId },
+    );
+
+    return { assignmentId: args.assignmentId, status: "confirmed" as const };
+  },
+});
+
+/**
  * Distinct users who have previously *confirmed* the given role, most
  * recent first. Powers the assign-UI "previously filled by" quicklink
  * (ADR-023 — a derived query in place of a qualification table).
--- a/apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
+++ b/apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
@@ -43,6 +43,7 @@ import {
   runSheetItemMatchesViewer,
   segmentHasContentForViewer,
 } from "@features/scheduling/utils/runSheetViewerFilter";
+import { songMetaLine } from "@features/scheduling/utils/songRehearsal";
 import { ServiceTimeSelector } from "@features/scheduling/components/ServiceTimeSelector";
 import { renderTextWithLinks } from "../utils/runSheetLinks";
 import { assigneeLabels } from "../utils/runSheetUtils";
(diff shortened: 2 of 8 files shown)


apps/convex/functions/scheduling/assignments.ts
```diff
@@ -874,6 +874,71 @@ export const respondToAssignment = mutation({
...
+  handler: async (ctx, args) => {
+    const callerId = await requireAuth(ctx, args.token);
+
+    const assignment = await ctx.db.get(args.assignmentId);
+    if (!assignment) {
+      throw new ConvexError("Assignment not found");
+    }
+    // Scoped to the assignment's own team, so a manager can confirm their
+    // team's cells but not another team's.
+    await requirePlanTeamScheduler(
```
lilseyi, line 904: **[test-coverage · medium — fixed in bd62d51]** The team scoping of this gate was originally unexercised: all four tests authorized as the group leader, who passes `requirePlanTeamScheduler` for *any* team in the group, so a bug that wired the wrong `teamId` (or dropped the team check) would have left every test green. Now pinned by three new tests: a community-admin confirm, an own-team-manager confirm, and a cross-team manager rejection (a manager of a second team in the same group is refused and the row stays unconfirmed).

---

apps/convex/functions/scheduling/assignments.ts
```diff
@@ -874,6 +874,71 @@ export const respondToAssignment = mutation({
...
+      status: "confirmed",
+      declineNote: undefined,
+      respondedAt: Date.now(),
+    });
+
+    // Same uniform channel-sync trigger as `respondToAssignment`: a confirm
+    // has no membership effect on the team channel itself, but a declined →
+    // confirmed override can re-add the user to derived membership, and
+    // reconciling on every status change keeps the trigger cheap and uniform.
+    await ctx.scheduler.runAfter(
```
lilseyi, line 926: **[test-coverage · medium — fixed in bd62d51]** These two scheduled reconciles are load-bearing for the declined → confirmed override (`teamChannelSync` derives channel membership *excluding* declined assignments, so the override must re-add the volunteer), but no test asserted they were enqueued — deleting both `runAfter` calls would have broken the re-add silently. Now pinned: after `confirmAssignment`, the scheduler queue holds exactly one `reconcileTeamChannel` and one `reconcileCrossTeamChannelsForSource` job, and zero `notifyLeadersOfResponse` jobs (the deliberate no-notify behavior documented above is asserted too).

---

apps/mobile/features/scheduling/components/RosterGridScreen.tsx
```diff
@@ -3179,6 +3201,18 @@ function RoleCellPopover({
                 </Text>
               </TouchableOpacity>
             ) : null}
+            {o.status !== "confirmed" ? (
```
lilseyi, line 3204: **[test-coverage · low — accepted, not fixing]** The Confirm-button visibility (`status !== "confirmed"`, here and in `MemberCellPopover`) has no component test — `RosterGridScreen` has none at all, and the popovers are private components in a ~3,400-line file. The condition is a single comparison with essentially no branching to get wrong; the one non-obvious case (a *declined* occupant intentionally still shows Confirm) is the documented override behavior and its backend half is test-covered. Extracting the popovers just to test a one-line condition isn't worth it. Also noting for the record: the button isn't permission-gated client-side — a non-scheduler tapping it surfaces the backend `ConvexError` via `surfaceError`, and the backend gate is what's load-bearing (now fully test-pinned).

---

apps/mobile/features/leader-tools/components/NativeRunSheetView.tsx
```diff
@@ -817,6 +824,7 @@ function ReadOnlyRow({
 }) {
   const isHeader = item.type === "header";
   const duration = formatDuration(item.durationSec);
+  const songMeta = item.type === "song" ? songMetaLine(item) : null;
```
lilseyi, line 827: **[test-coverage · low — accepted, not fixing]** `ReadOnlyRow` itself is untested, so the `item.type === "song"` gate and the render-in-situ path aren't exercised — the meta line now renders in strictly more cases than before (previously required a `songDetails.key` override; now any linked song title, default key, or BPM triggers it). Accepting because the string logic is exhaustively unit-tested (17 cases in `songRehearsal.test.ts`, including the offline-rehydration shape where cached items carry no `song` field), the component change is a two-line conditional render, and the backend `listItems` join was verified live. A component test would require standing up the full run-sheet render harness for marginal coverage.

---

lilseyi (commented):
Adversarially-verified review by four parallel subagents (correctness, security, spec-fidelity, test-coverage) on this diff.

**Correctness — clean.** All three assignment statuses handled; already-confirmed is a true early-return no-op; declined→confirmed clears `declineNote` and re-schedules the channel reconciles that a decline had dropped; `songMetaLine` edge cases (no key/bpm, title-match suppression, no linked song, offline-cached items without the `song` join) verified against the live `listItems` hydration; all four popover call sites pass `onConfirm`; reminder nudges re-query status at fire time so confirmed volunteers stop being nudged.

**Security — clean.** `confirmAssignment` reads plan/team from the assignment row itself (no client-supplied scope → no IDOR); `requirePlanTeamScheduler` re-verifies team-belongs-to-group and manager rows against active membership; privilege scope is identical to the pre-existing `unassign` (anyone who can confirm could already delete). No user-supplied strings written; RN `<Text>` rendering leaves no injection surface for song titles.

**Spec-fidelity — clean.** ADR-027 override layering confirmed (`songDetails.key ?? song.defaultKey`, same for BPM); both run sheets covered via the shared `PlanRunSheet`; Event Plans guide claims verified accurate.

**Test-coverage — 2 medium + 3 low findings**, detailed inline: both mediums and one low fixed in bd62d51 (4 new backend tests: admin/manager positive paths, cross-team manager rejection, reconcile + no-notify queue assertions — suite now 35/35); the two remaining lows are accepted with rationale on their threads.

---

PR #819, opened 2026-08-23
fix(chat): composer input is untappable on Android

lilseyi:
## The bug

Tapping the message field on Android never raised the keyboard. The `+`, camera and mic beside it all worked — only the field itself was inert. Since the field could never be typed into, the state couldn't be recovered from.

Reported via device video (2026-08-23): repeated taps on the composer do nothing, then a tap lands on `+` (attachment tray opens) and later on the mic (record-audio permission prompt) — confirming everything *around* the field is live.

## Root cause

The WhatsApp-shell composer draws its empty-field hint as an absolutely-positioned `Text` with `pointerEvents="none"`, layered over the `TextInput` and covering its whole tappable area. That opt-out works only on iOS.

On Android, in RN 0.81:

- `TouchTargetHelper.kt:317` reads `pointerEvents` **only** from views implementing `ReactPointerEventsView`. `ReactViewGroup` (the backing view for `View`) is the **only** implementer in the tree — `ReactTextView` is not one.
- The sibling hit test (`TouchTargetHelper.kt:217`) walks children topmost-first and returns the **first** hit, with no fall-through to lower siblings.
- `ReactEditText.kt:99`: *"ReactEditTexts have setFocusableInTouchMode set to false automatically because touches on the EditText are managed on the JS side."* Android focus therefore comes **only** from `TextInput.js:585-593` (`usePressability` → `inputRef.current.focus()`). Losing the JS hit test loses focus outright — native dispatch alone would never have focused the field either.

So the hint `Text` won the taps and the `ReactEditText` never focused. It only unmounts once `text.length > 0`, which requires typing, so the field couldn't be typed out of the broken state. (Not quite 100% dead: the hint stops at `right: 44`, so taps in that strip still landed when no KLIPY glyph was there to take them.)

**Why this shipped green:** `Text.d.ts:211-213` *does* declare `pointerEvents`, documented as *"Similar to `View`'s `pointerEvents`"*. Only the Flow `TextProps.js` omits it. So the broken version typechecks, passes tests, and renders correctly — nothing but a real device catches it.

The flag-off composer was never affected: it uses the native `placeholder` prop and has no overlay.

## The fix

Move the tap-through opt-out onto a wrapping `View` with `pointerEvents="none"`, which Android does honour. The style is split so the box keeps the absolute insets and the text keeps its italic line box.

Geometry is unchanged, verified against Yoga: `AbsoluteLayout.cpp:116` uses `resolveChildAlignment` for absolute children when the inset is undefined, so `waFieldWrap`'s `alignItems: 'flex-end'` still seats the box — and the box takes the same height from the child's `lineHeight`. The wrapper also can't be flattened away by Fabric: `ViewShadowNode.cpp:49` lists `pointerEvents == PointerEventsMode::None` as forcing `formsView`.

## Tests

The guard asserts the **invariant**, not just this node: every absolutely-positioned child of the pill must be tap-through, so the naive next overlay (a character counter, an edit badge) can't silently kill the composer the same way. It runs under `Platform.OS = 'android'`, accepts the opt-out in style or prop form, and is scoped to the composer field. A second test pins the hint's copy and that it unmounts once the field is non-empty.

Mutation-verified — each of these slips past a node-bound guard and fails against this one:

| Mutation | Result |
| --- | --- |
| `pointerEvents` moved back onto the `Text` | caught |
| wrapper kept, `pointerEvents` dropped | caught |
| a rogue absolute overlay `Text` added to the pill | **caught (was missed)** |
| `Platform.OS === 'android'` reintroduction of the original bug | **caught (was missed)** |
| unmodified code (control) | passes |

- `pnpm test` in `apps/mobile` → 252 suites, 2677 tests, all pass
- eslint clean via the pre-commit hook
- Pre-existing and untouched: `MessageInput.tsx` already fails typecheck on `main` at `:292`, `:306`, `:898` (TS2322) and `:1386` (TS2339)

## Review

Four parallel lenses (correctness, test adequacy, spec-fidelity, completeness), each instructed to refute the diagnosis, with findings posted in the review thread. The diagnosis survived. Six findings were addressed in ca432db — the most significant being that the original code comment stated the *opposite* of the load-bearing fact about `Text.d.ts`, and that the first version of the test couldn't catch two realistic regressions.

One claim from the first draft of this PR was **disproved** and is withdrawn: `behavior="padding"` on `KeyboardAvoidingView` is redundant on Android, not harmful — `KeyboardAvoidingView.js:107-109` computes from the measured frame, so under `adjustResize` the term collapses to ~0.

No `apps/web` guide update is owed (no chat/messaging row in the feature→guide map, and this restores intended behavior rather than changing it). No dependency surface touched — `native-deps.json`, `app.json`, both `package.json`s and the lockfile are all unchanged.

## Merge gate

CI structurally cannot validate this — the original defect shipped with a green suite, and RNTL runs no hit test, so these tests prove tree shape, never that a real tap focuses a real `EditText`. Per CLAUDE.md's native-safety rule, **this needs a device or staging-OTA pass on Android before merge**: open a chat, tap the empty composer, confirm the keyboard comes up.

## Follow-up (separate PR)

The same defect is live in `ChatInboxScreen`, and one instance is over a real button. `styles.smallTitle` (`:2814-2821`) is `position: absolute; left: 0; right: 0`, always mounted and only opacity-animated. Because the hit walk is reverse-index, the trailing `+` survives but the community-avatar button at index 0 is tested *after* the Text — so on Android roughly the middle half of "Open community page" is dead. (The file's own comment at `:946-950` already notes that an opacity-0 view still hit-tests on Android.) The second instance at `:1154-1160` is benign — the only interactive element in `navRow` is rendered after the Text.

ca432db adds `// TODO: Investigate` breadcrumbs at both sites. The fix needs an `Animated.View` wrapper with the animated style moved onto it, plus device verification of the header animation — hence a separate PR.

That PR should also carry an ESLint `no-restricted-syntax` rule banning `pointerEvents` as a JSX attribute on `Text`/`Animated.Text`. Since `Text.d.ts` declares the prop, **typecheck will never catch a recurrence**; a lint rule will. A repo-wide scan finds exactly 3 non-`.web` instances today — the one fixed here plus those two — so the rule lands clean once they're fixed.

### DIFF
--- a/apps/mobile/features/chat/components/ChatInboxScreen.tsx
+++ b/apps/mobile/features/chat/components/ChatInboxScreen.tsx
@@ -1070,6 +1070,13 @@ export function ChatInboxScreen({
             </WaFloatingButton>
           )}
         </View>
+        {/* TODO: Investigate - `pointerEvents` is a no-op on <Text> on Android
+            (only `ReactViewGroup` implements `ReactPointerEventsView`), so this
+            title swallows taps in its rect there. `Text.d.ts` declares the prop,
+            which is why it typechecks. Fix needs an <Animated.View> wrapper with
+            the animated style moved onto it — device-verify the header
+            animation when doing it. Same defect as the composer hint fixed in
+            MessageInput.tsx; see the note there. */}
         <Animated.Text
           style={[styles.smallTitle, { color: colors.text }, smallTitleStyle]}
           numberOfLines={1}
@@ -1151,6 +1158,13 @@ export function ChatInboxScreen({
             {collapsingHeaderTitle}
           </Text>
         </Animated.View>
+        {/* TODO: Investigate - `pointerEvents` is a no-op on <Text> on Android
+            (only `ReactViewGroup` implements `ReactPointerEventsView`), so this
+            title swallows taps in its rect there. `Text.d.ts` declares the prop,
+            which is why it typechecks. Fix needs an <Animated.View> wrapper with
+            the animated style moved onto it — device-verify the header
+            animation when doing it. Same defect as the composer hint fixed in
+            MessageInput.tsx; see the note there. */}
         <Animated.Text
           style={[styles.smallTitle, { color: colors.text }, smallTitleStyle]}
           numberOfLines={1}
--- a/apps/mobile/features/chat/components/MessageInput.tsx
+++ b/apps/mobile/features/chat/components/MessageInput.tsx
@@ -1293,14 +1293,26 @@ export function MessageInput({ channelId, replyToMessage, onCancelReply, onReply
             editable={!uploading}
           />
           {text.length === 0 && (
-            <Text
-              pointerEvents="none"
-              numberOfLines={1}
-              ellipsizeMode="tail"
-              style={[styles.waHintOverlay, { color: themeColors.textTertiary }]}
-            >
-              {(placeholder ?? "").trim() || "Message..."}
-            </Text>
+            /* The tap-through opt-out lives on this <View>, NOT on the <Text>
+               inside it. The hint covers the whole tappable area of the
+               TextInput above, and on Android `pointerEvents` is only read off
+               views implementing `ReactPointerEventsView` — `ReactViewGroup`
+               (<View>) is the only one; `ReactTextView` isn't, and `TextProps`
+               doesn't even declare the prop. A bare <Text pointerEvents="none">
+               therefore stays the hit target (the sibling walk is topmost-first
+               with no fall-through), the EditText never focuses, and since the
+               hint only unmounts once the field is non-empty the composer is
+               permanently dead. See the regression test next door. */
+            <View pointerEvents="none" style={styles.waHintOverlay}>
+              <Text
+                testID="wa-composer-hint"
+                numberOfLines={1}
+                ellipsizeMode="tail"
+                style={[styles.waHintOverlayText, { color: themeColors.textTertiary }]}
+              >
+                {(placeholder ?? "").trim() || "Message..."}
+              </Text>
+            </View>
           )}
           {/* Hidden without a KLIPY key — the documented degradation is "GIF
               picker hidden", matching the attachment-sheet option's gate. */}
@@ -1637,9 +1649,12 @@ const styles = StyleSheet.create({
     // paints the browser's focus outline around it. Ignored on native.
     outlineStyle: 'none',
   } as any,
-  /** Empty-field hint overlay: one quiet italic line, vertically centered
-   *  in the pill; sits behind taps (pointerEvents none) and clear of the
-   *  in-field sticker glyph. */
+  /** Empty-field hint overlay: the positioning box. Takes its height from the
+   *  single line of text inside it (`lineHeight` === the field height), and
+   *  with no `top`/`bottom` the wrap's `alignItems: 'flex-end'` seats it on the
+   *  field's baseline row — the geometry the hint Text used to own directly.
+   *  Carries `pointerEvents="none"` in the JSX; see the note there for why the
+   *  <Text> can't. */
   waHintOverlay: {
     position: 'absolute',
     left: 16,
@@ -1647,6 +1662,9 @@ const styles = StyleSheet.create({
     // padding + the wrap's 4pt right padding = 38, rounded up) — NOT the field
     // height, which this deliberately does not track.
     right: 44,
+  },
+  /** One quiet italic line, vertically centered in the pill by its line box. */
+  waHintOverlayText: {
     fontSize: 14,
     fontStyle: 'italic',
     lineHeight: WA_COMPOSER_FIELD_HEIGHT,
--- a/apps/mobile/features/chat/components/MessageInput.tsx
+++ b/apps/mobile/features/chat/components/MessageInput.tsx
@@ -1297,12 +1297,21 @@ export function MessageInput({ channelId, replyToMessage, onCancelReply, onReply
                inside it. The hint covers the whole tappable area of the
                TextInput above, and on Android `pointerEvents` is only read off
                views implementing `ReactPointerEventsView` — `ReactViewGroup`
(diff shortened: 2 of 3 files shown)


lilseyi (commented):
## Review cycle — 4 parallel lenses, adversarially verified

Per `docs/dev-assistant/ROUTINE-PROMPT.md` Routine 3: correctness, test adequacy, spec-fidelity, and completeness-of-fix, each instructed to try to refute the diagnosis. Findings that survived are addressed in ca432db.

### Verdict: the diagnosis holds

Three lenses independently verified the mechanism against the vendored RN 0.81.5 source, and one surfaced stronger evidence than the PR originally had:

- `ReactEditText.kt:99` — *"ReactEditTexts have setFocusableInTouchMode set to false automatically because touches on the EditText are managed on the JS side."* Android focus therefore comes **only** from `TextInput.js:585-593` (`usePressability` → `inputRef.current.focus()`), i.e. entirely from the JS target `TouchTargetHelper` picks. Losing the hit test means losing focus outright — native dispatch alone would never have focused the field.
- Post-fix both layers fall through: JS-side `TouchTargetHelper.kt:229-234` restores coords and continues the sibling loop past a `NONE` child; native-side `ReactViewGroup.kt:277-280` returns `false` from `onTouchEvent` when `canBeTouchTarget(NONE)` is false.
- The wrapper cannot be flattened away by Fabric — `ViewShadowNode.cpp:49` lists `pointerEvents == PointerEventsMode::None` as forcing `formsView`.
- Layout is genuinely unchanged: `yoga/algorithm/AbsoluteLayout.cpp:116` uses `resolveChildAlignment` for absolute children when the inset is undefined, so `alignItems: 'flex-end'` still applies, to a box of the same height.

### Fixed in ca432db

**1. (Medium) The comment stated a falsehood, and it was the load-bearing fact.** It said `TextProps` doesn't declare `pointerEvents`. Only the *Flow* `TextProps.js` omits it — `Text.d.ts:211-213` **does** declare it, as *"Similar to `View`'s `pointerEvents`"*. That is the entire trap: the broken version typechecks and ships green. The comment told a future reader the opposite of what will bite them.

**2. (Medium) The guard was bound to one node, not the invariant.** Mutation testing showed a rogue absolute overlay `Text` added to the pill — the naive way anyone would add a character counter or edit badge — killed the composer on both platforms and the test still passed. It now asserts every absolutely-positioned child of the pill is tap-through.

**3. (Medium) The test ran under `Platform.OS = 'ios'` for an Android-only bug.** A `Platform.OS === 'android' ? <Text pointerEvents> : <View>` reintroduction passed. Now runs under Android.

Both regressions were mutation-verified as newly caught; the control still passes.

**4. (Low) Sweep scoped to the composer field.** `UNSAFE_root.findAllByType('Text')` also covered Ionicons glyphs, so a vector-icons bump would have failed this test pointing at the composer hint.

**5. (Low) Assertions cover what the style split could actually lose** — `fontSize` (inert if left on the `View`) and the absent `top`/`bottom` (which is what makes `alignItems: 'flex-end'` seat the hint). Plus a second test pinning the hint's copy and that it unmounts once the field is non-empty.

**6. (Low) "Permanently dead" was overstated.** The hint stops at `right: 44`, so taps in that strip did still reach the input when no KLIPY glyph was there to take them. Corrected.

### Accepted as-is

- **No `apps/web` guide update owed.** The feature→guide map in `guides-and-link-previews` has no chat/messaging row, and this restores intended behavior rather than changing it. `WA-VISUAL-DELTAS.md` §S4.6 never describes the hint overlay, so the spec doc isn't stale either.
- **No dependency surface touched** — `native-deps.json`, `app.json`, both `package.json`s and the lockfile are all untouched, so the `native-deps-safety` gates aren't implicated beyond the device-verification clause.
- **`ChatRoomSurface`'s full-screen `Pressable` cannot steal the tap** — `Pressability.js:453-480` supplies only bubble-phase `onStartShouldSetResponder` with no capture variant, so the deeper `TextInput` claims the responder first; and `Pressable` never sets `blockNativeResponder`.
- **`behavior="padding"` on Android is redundant, not harmful** — `KeyboardAvoidingView.js:107-109` computes from the measured frame, so under `adjustResize` the term collapses to ~0. I'd flagged this as a possible second bug in the original PR body; it isn't one.

### Outstanding — not fixed here

**The `ChatInboxScreen` sibling is not benign, and one instance is over a live button.** `styles.smallTitle` (`:2814-2821`) is `position: absolute; left: 0; right: 0`, mounted at all times and only opacity-animated. Because the hit walk is reverse-index, the trailing `+` (index 2) survives, but the community-avatar button at index 0 is tested *after* the Text — so on Android roughly the middle half of "Open community page" is dead. The file's own comment at `:946-950` already documents that an opacity-0 view still hit-tests on Android. The second instance (`:1154-1160`, flag-off header) *is* benign: the only interactive element in `navRow` is rendered after the Text.

ca432db adds `// TODO: Investigate` breadcrumbs at both sites. Fixing needs an `Animated.View` wrapper with the animated style moved onto it, and device verification of the header animation — deliberately out of scope for a composer fix, but it should not sit unfixed for long.

**A durable guard is worth considering**: since `Text.d.ts` declares the prop, typecheck will *never* catch a recurrence. An ESLint `no-restricted-syntax` rule on `pointerEvents` as a JSX attribute of `Text`/`Animated.Text` would. A repo-wide scan finds exactly 3 non-`.web` instances today — the fixed one plus those two — so the rule would land clean once they're fixed.

**Pre-existing, untouched:** `MessageInput.tsx` already fails typecheck on `main` at `:292`, `:306`, `:898` (TS2322, `number` vs `Timeout`) and `:1386` (TS2339, `textDisabled`). Not introduced here and out of scope.

### The real merge gate

CI cannot validate this change — the original defect shipped with a green suite, and RNTL runs no hit test, so these tests prove tree shape, never that a real tap focuses a real `EditText`. Per CLAUDE.md's native-safety rule this needs a device or staging-OTA pass on Android before merge.

---

